PR #99, opened 2014-01-22
Update RibbonModule.java

amajumdar:
I noticed several severe issues with the use of feign client and feign.RibbonModule. All this while I was under the impression that it leverages the Ribbon implementation of http client(RestClient) but it does not. 

I saw the following issues:
1. Connection leaks. The server runs out of file descriptors after a day of hundreds of client calls.
2. Every time a request is made, a new connection is opened.
3. Also some of the Ribbon properties like shiro-client.ribbon.OkToRetryOnAllOperations=false are ignored.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -113,7 +113,8 @@ project(':feign-ribbon') {
 
     dependencies {
         compile     project(':feign-core')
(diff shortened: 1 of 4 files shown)


adriancole:
this probably needs to be revisited using latest ribbon and latest feign. Feel free to reopen!

PR #107, opened 2014-05-20
Added ability to expand and encode variables in body annotation and expand request line variables that are not already key pairs 

sjhorn:
Example below serialises complex type "Operation" into the inputData placeholder in @Body.
 
```java

@RequestLine("POST /sdpapi/request")
@Body("OPERATION_NAME=GET_REQUESTS&TECHNICIAN_KEY={technicalKey}&INPUT_DATA={inputData}")
    public API getRequests(
        @Named("technicalKey") String technicalKey,
        @Named("inputData") Operation operation
    )

```

The example below automatically expands the key-value pairs for the fields variable.

```java
@RequestLine("PUT /1/cards/{cardIdOrShortLink}?key={key}&token={token}&{fields}")
    public Card updateCardWithFields(
        @Named("key") String key,
        @Named("token") String token,
        @Named("cardIdOrShortLink") String cardIdOrShortLink,
        @Named("fields") Object fields
    )
```

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -169,6 +169,7 @@ public interface Contract {
           String varName = '{' + name + '}';
           if (data.template().url().indexOf(varName) == -1 &&
               !searchMapValues(data.template().queries(), varName) &&
+              !data.template().queries().containsKey(varName) &&
               !searchMapValues(data.template().headers(), varName)) {
             data.formParams().add(name);
           }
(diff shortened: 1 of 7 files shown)


adriancole:
mind having a try with the latest `@Param(value="name", expander = CustomExpander.class)`? Not sure it will work, but worth a shot.

#152

adriancole:
hopefully expander addresses this. If not, it will need rework: feel free to open in such case.

PR #143, opened 2015-01-25
Replace TestNG with JUnit + Rules

adriancole:
JUnit Rules, such as MockWebServerRule, reduce boilerplate setup present
in our tests. By migrating off TestNG, and onto rules, our tests become
more maintainable as JUnit is well understood.

Note:

Once this is in, I plan to eliminate a large amount of clutter by using assertj as opposed to assertX methods. I will also make a couple custom ones (like RequestTemplate, which is often the source of hard-to-read assertions). After this, I'll rebase my de-dagger pull request.


adriancole:
note. this didn't affect any main source files, so I plan to merge it after dinner. review-after-commit is welcome if anyone feels the urge.

PR #147, opened 2015-01-26
Adds OkHttp integration

### LINKED ISSUE #134: using patch or other methods with HttpURLConnection
What's the thought on applying a workaround similar to jersey to use PATCH or other methods to HttpURLConnection?   

see:
https://github.com/jersey/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/HttpUrlConnector.java#L388

from: http://java.net/jira/browse/JERSEY-639

adriancole:
Adds OkHttp integration. Allows multiple headers with the same name. Backfills default client tests.

closes #134


jdamick:
Thanks! this helps a lot!

PR #153, opened 2015-01-27
Makes examples standalone and built from standard Gradle or Maven

### LINKED ISSUE #111: Change feign-example-* build to compile against the current sources
This will ensure that the examples are always up-to-date. Also, this fixes the problem that snapshot/release build jobs do not always have access to old releases in its internal artifacts repository.

adriancole:
Examples are intended to not rely on netflix build, rather be something for users to make practice projects, using latest stable version. This disconnects the projects and also adds a maven option.

Closes #111

### DIFF
--- a/example-wikipedia/pom.xml
+++ b/example-wikipedia/pom.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.sonatype.oss</groupId>
+    <artifactId>oss-parent</artifactId>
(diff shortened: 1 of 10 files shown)


adriancole:
cc @allenxwang @rspieldenner @brianm

example-wikipedia/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+              </transformers>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.skife.maven</groupId>
+        <artifactId>really-executable-jar-maven-plugin</artifactId>
+        <version>1.3.0</version>
```
adriancole: I'd have used 1.4.0, but.. not in central yet :) @brianm

example-wikipedia/src/main/java/feign/example/wikipedia/ResponseAdapter.java
```diff
@@ -58,17 +58,11 @@
           }
         }
         reader.endObject();
-      } else if ("query-continue".equals(nextName)) {
+      } else if ("continue".equals(nextName)) {
```
adriancole: wikipedia api drift

PR #155, opened 2015-01-31
Makes body parameter type explicit

adriancole:
Feign has `MethodMetadata.bodyType()`, but never passed it to encoders.
Encoders that register type adapters need to do so based on the
interface desired as opposed to the implementation class. This change
breaks api compatibility for < 8.x, by requiring an additional arg
on `Encoder.encode`.

see https://github.com/square/retrofit/issues/713

### DIFF
--- a/jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
+++ b/jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
@@ -29,100 +29,100 @@ import java.util.concurrent.ConcurrentHashMap;
  * Creates and caches JAXB contexts as well as creates Marshallers and Unmarshallers for each context.
  */
 public final class JAXBContextFactory {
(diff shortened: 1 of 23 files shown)


jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
```diff
@@ -29,100 +29,100 @@
  * Creates and caches JAXB contexts as well as creates Marshallers and Unmarshallers for each context.
  */
 public final class JAXBContextFactory {
-    private final ConcurrentHashMap<Class, JAXBContext> jaxbContexts = new ConcurrentHashMap<Class, JAXBContext>(64);
```
adriancole: lots of reformatting on jaxb as the inconsistent indent was driving me bonkers.

PR #159, opened 2015-02-02
Reformats code according to Google Java Style

adriancole:
Files had various formatting differences, as did pull requests. Rather than
create our own style, this inherits and requires the well documented Google
Java Style.


adriancole:
Specifically, I reformatted using Intellij and [this file](https://google-styleguide.googlecode.com/svn/trunk/intellij-java-google-style.xml)

PR #161, opened 2015-02-04
Headers substitutions were not being expanded by the value name

jdamick:
Header params were using the header name instead of the value.

So if you had:

```
@Header({'Auth-Token', 'authToken'})
public String get(@Param('authToken') token);
```

It was looking up by 'Auth-Token' instead of 'authToken'  so it wouldn't substitute the value in 'token' above.


adriancole:
whoops. thanks for the fix!

adriancole:
when I get home, I'll cherry-pick into 7.x

we should collect any other bugs you, others, or I pickup prior to cutting 7.2.1 maybe timeout of next mondayish.

jdamick:
this is actually blocking us right now.. is it possible to release it sooner?

PR #178, opened 2015-02-06
Changed RibbonClient to pass Feign Request.Options to LBClient

bstick12:
With suggested changes and from a different branch.

Could this be merged back to the next 7.x release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 * Removes support for parameters annotated with `javax.inject.@Named`. Use `feign.@Param` instead.
 * Makes body parameter type explicit.
 
+### Version 7.3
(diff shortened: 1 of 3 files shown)


adriancole:
Looks good. Thanks!

An editorial request below:

Add an entry in CHANGES.md for 7.3 that says the following, and also reword your commit to say the same.

"Adds Request.Options support to RibbonClient"

When worded like this, the git log basically reads:
this commit "Adds Request.Options support to RibbonClient"

Also, those wondering why they should update to Feign 7.3 will know what's new via CHANGES.md

Looks nice, right?

adriancole:
fantastic. will cherry-pick into 7.x

bstick12:
@adriancole Thanks very much.

PR #185, opened 2015-02-12
Retain scheme in LBClient.RibbonRequest URI - fixes #183

### LINKED ISSUE #183: RibbonClient doesn't support HTTPS
Using the following test - https://gist.github.com/bstick12/2a21ffd30b14ec68716f#file-ribbonclienttest

You will see the following exception raised in the logs

```
21:04:25.583 [pool-1-thread-2] DEBUG c.n.l.LoadBalancerExecutor - Got error java.net.SocketException: Unexpected end of file from server when executed on server api.github.com:443
```
This is due to the request being made to http://api.github.com:443/. 

The URI created in RibbonClient does include a scheme and eventually com.netflix.loadbalancer.LoadBalancerContext.deriveSchemeAndPortFromPartialUri(URI) will default the scheme to http.

bstick12:


adriancole:
Thanks for the help. I think we need a test for this, right? Can you port something like this into the ribbon test? That way future work doesn't undo yours. https://github.com/Netflix/feign/blob/master/core/src/test/java/feign/client/DefaultClientTest.java#L119

bstick12:
Added a test to check the https connection works

ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
```diff
@@ -123,6 +125,26 @@ public void urlEncodeQueryStringParameters() throws IOException, InterruptedExce
     assertEquals(recordedRequestLine, expectedRequestLine);
   }
 
+
+  @Test
+  // Issue-183
```
adriancole: nit I don't think this comment is required. the commit log already says this, and the need for this feature is obvious.

adriancole:
hypernit on comment, also I'm trying to help us get more readable commit log. can you switch to this form?
Note using retains vs retain makes the commit log look like.

"this commit retains scheme in LBClient.RibbonRequest"

```
Retains scheme in LBClient.RibbonRequest URI

Before this change, we were dropping scheme, which prevented use of https.

closes #183
``

Issue #182, opened 2015-02-12
JmxMonitorRegistry race condition with RibbonClient


bstick12:
If you run the following test - https://gist.github.com/bstick12/2a21ffd30b14ec68716f#file-ribbonclienttest

You will occasionally get the the following warning issued from servo.

The com.netflix.loadbalancer.LoadBalancerContext.initWithNiwsConfig(IClientConfig) registers the load balancer and the RibbonClient is recreating the LBClient on each request.

```
0:44:38.155 [pool-1-thread-2] WARN  c.n.servo.jmx.JmxMonitorRegistry - Unable to register Monitor:MonitorConfig{name=Client_github, tags=class=LBClient, policy=DefaultPublishingPolicy}
javax.management.InstanceNotFoundException: com.netflix.servo:name=github_LoadBalancerExecutionTimer,statistic=max,class=LBClient,type=GAUGE,id=Client_github,unit=MILLISECONDS
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getMBean(DefaultMBeanServerInterceptor.java:1095) ~[na:1.7.0_72]
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.exclusiveUnregisterMBean(DefaultMBeanServerInterceptor.java:427) ~[na:1.7.0_72]
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.unregisterMBean(DefaultMBeanServerInterceptor.java:415) ~[na:1.7.0_72]
	at com.sun.jmx.mbeanserver.JmxMBeanServer.unregisterMBean(JmxMBeanServer.java:546) ~[na:1.7.0_72]
	at com.netflix.servo.jmx.JmxMonitorRegistry.register(JmxMonitorRegistry.java:82) ~[servo-core-0.7.4.jar:na]
	at com.netflix.servo.jmx.JmxMonitorRegistry.register(JmxMonitorRegistry.java:106) ~[servo-core-0.7.4.jar:na]
	at com.netflix.servo.DefaultMonitorRegistry.register(DefaultMonitorRegistry.java:135) [servo-core-0.7.4.jar:na]
	at com.netflix.servo.monitor.Monitors.registerObject(Monitors.java:207) [servo-core-0.7.4.jar:na]
	at com.netflix.loadbalancer.LoadBalancerContext.initWithNiwsConfig(LoadBalancerContext.java:111) [ribbon-loadbalancer-2.0-RC5.jar:na]
[...]

PR #187, opened 2015-02-13
Reuse LBClient to avoid duplicate servo registrations

bstick12:
issue #182 found that the default configuration of servo does static registration for each LBClient instance. This creates problems at runtime. The solution proposed here is to avoid duplicating LBClient instances by caching them.

### DIFF
--- a/ribbon/src/main/java/feign/ribbon/RibbonClient.java
+++ b/ribbon/src/main/java/feign/ribbon/RibbonClient.java
@@ -1,30 +1,39 @@
 package feign.ribbon;
 
-import com.netflix.client.ClientException;
+import java.io.IOException;
+import java.net.URI;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
 import com.netflix.client.ClientFactory;
 import com.netflix.client.config.CommonClientConfigKey;
 import com.netflix.client.config.DefaultClientConfigImpl;
 import com.netflix.client.config.IClientConfig;
 import com.netflix.loadbalancer.ILoadBalancer;
 
-import java.io.IOException;
-import java.net.URI;
-
 import feign.Client;
 import feign.Request;
 import feign.Response;
 
 /**
- * RibbonClient can be used in Fiegn builder to activate smart routing and resiliency capabilities
+ * RibbonClient can be used in Feign builder to activate smart routing and resiliency capabilities
  * provided by Ribbon. Ex.
+ * <p>
+ * 
  * <pre>
- * MyService api = Feign.builder.client(new RibbonClient()).target(MyService.class,
- * "http://myAppProd");
+ * MyService api = Feign.builder.client(RibbonClient.DEFAULT).target(MyService.class,
+ *     &quot;http://myAppProd&quot;);
  * </pre>
+ * 
+ * <p>
  * Where {@code myAppProd} is the ribbon client name and {@code myAppProd.ribbon.listOfServers}
  * configuration is set.
  */
 public class RibbonClient implements Client {
+
+  public static RibbonClient DEFAULT = new RibbonClient();
+  
+  private volatile Map<String, LBClient> lbClientCache = new LinkedHashMap<String, LBClient>();
 
   private final Client delegate;
 
@@ -41,15 +50,13 @@ public class RibbonClient implements Client {
     try {
       URI asUri = URI.create(request.url());
       String clientName = asUri.getHost();
-      URI
-          uriWithoutSchemeAndPort =
-          URI.create(request.url().replace(asUri.getScheme() + "://" + asUri.getHost(), ""));
-      LBClient.RibbonRequest
-          ribbonRequest =
-          new LBClient.RibbonRequest(request, uriWithoutSchemeAndPort);
-      return lbClient(clientName).executeWithLoadBalancer(ribbonRequest,
-          new FeignOptionsClientConfig(options)).toResponse();
-    } catch (ClientException e) {
+      URI uriWithoutHost = URI.create(request.url().replace(asUri.getHost(), ""));
+      LBClient.RibbonRequest ribbonRequest =
+          new LBClient.RibbonRequest(request, uriWithoutHost, delegate);
+      return getOrCreateLBClient(clientName).executeWithLoadBalancer(ribbonRequest,
+          new FeignOptionsClientConfig(options))
+          .toResponse();
+    } catch (Exception e) {
       if (e.getCause() instanceof IOException) {
         throw IOException.class.cast(e.getCause());
       }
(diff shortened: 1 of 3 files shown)


ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -1,31 +1,40 @@
...
+ * 
+ * <p>
  * Where {@code myAppProd} is the ribbon client name and {@code myAppProd.ribbon.listOfServers}
  * configuration is set.
  */
 public class RibbonClient implements Client {
 
+  public static RibbonClient DEFAULT = new RibbonClient();
+  
+  private volatile Map<String, LBClient> lbClientCache = new HashMap<String, LBClient>();
```
adriancole: nit: always linkedhashmap
Discussed in the past, but this has consistent ordering which costs almost nothing. when tests are written and/or JREs switch, this pays dividends.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -41,28 +50,37 @@ public Response execute(Request request, Request.Options options) throws IOExcep
...
     }
   }
-
-  private LBClient lbClient(String clientName) {
-    IClientConfig config = ClientFactory.getNamedConfig(clientName);
-    ILoadBalancer lb = ClientFactory.getNamedLoadBalancer(clientName);
-    return new LBClient(delegate, lb, config);
-  }
   
+  private LBClient getOrCreateLBClient(String clientName) {
```
adriancole: add a quick javadoc saying why we do this.

adriancole:
Nice, brendan. I think this is ready for tests.

adriancole:
once this is all tidy, squash and rename the commit to something like..

```
Reuses LBClient

Ribbon does not expect multiple LBClient instances for the same client. This
introduces a cache to avoid redundantly creating them.

closes #182
```

bstick12:
@adriancole Changes for #182 and #183 should done now. Could these be merged back to the 7.x?

adriancole:
Yeah. I will cherry-pick and resolve anything that comes up. Thanks!

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -57,12 +64,33 @@ public Response execute(Request request, Request.Options options) throws IOExcep
...
+   * to avoid redundantly creating them.
+   * 
+   * @param clientName - The clientName that the {@link LBClient} will be cache against
+   * @return
+   */
+  private LBClient getOrCreateLBClient(String clientName) {
+    if (lbClientCache.containsKey(clientName)) {
+      return lbClientCache.get(clientName);
+    } else {
+      synchronized (this) {
```
adriancole: nit: switch this to lock on lbClientCache.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -1,31 +1,40 @@
...
+ * 
+ * <p>
  * Where {@code myAppProd} is the ribbon client name and {@code myAppProd.ribbon.listOfServers}
  * configuration is set.
  */
 public class RibbonClient implements Client {
 
+  public static RibbonClient DEFAULT = new RibbonClient();
+  
+  private volatile Map<String, LBClient> lbClientCache = new LinkedHashMap<String, LBClient>();
```
adriancole: change from volatile to final. final ensures threads can see this. volatile isn't needed as we aren't changing the reference, rather the contents.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -57,12 +64,33 @@ public Response execute(Request request, Request.Options options) throws IOExcep
...
+   * is created and added to the cache.
+   * <p>
+   * Ribbon does not expect multiple LBClient instances for the same client. This introduces a cache
+   * to avoid redundantly creating them.
+   * 
+   * @param clientName - The clientName that the {@link LBClient} will be cache against
+   * @return
+   */
+  private LBClient getOrCreateLBClient(String clientName) {
+    if (lbClientCache.containsKey(clientName)) {
```
adriancole: switch to this form, as containsKey and get invoke the same logic internally.

```java
LBClient result = cache.get(clientName);
if (result != null) return result;
synchronized(cache) {
  // your code here.
}
```

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -57,12 +64,33 @@ public Response execute(Request request, Request.Options options) throws IOExcep
...
-    ILoadBalancer lb = ClientFactory.getNamedLoadBalancer(clientName);
-    return new LBClient(delegate, lb, config);
+  /**
+   * Returns a cached version of the {@link LBClient} if it exists otherwise a new {@link LBClient}
+   * is created and added to the cache.
+   * <p>
+   * Ribbon does not expect multiple LBClient instances for the same client. This introduces a cache
+   * to avoid redundantly creating them.
+   * 
+   * @param clientName - The clientName that the {@link LBClient} will be cache against
```
adriancole: nit remove this as it is obvious. also remove the empty return

ribbon/src/main/java/feign/ribbon/RibbonClient.java
adriancole: actually, double-check locking won't save us. ex. cache.get could read a corrupt table if another client is being inserted (ex. during the lookup of client1, client2 is being inserted). put the whole shebang in a lock.

```java
synchronized(cache) {
  LBClient result = cache.get(clientName);
  if (result != null) return result;
  
  IClientConfig config = ClientFactory.getNamedConfig(clientName);
  ILoadBalancer lb = ClientFactory.getNamedLoadBalancer(clientName);
  result = new LBClient(lb, config);

  cache.put(clientName, result);
  return result;
}
```

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -1,31 +1,40 @@
...
+ *     &quot;http://myAppProd&quot;);
  * </pre>
+ * 
+ * <p>
  * Where {@code myAppProd} is the ribbon client name and {@code myAppProd.ribbon.listOfServers}
  * configuration is set.
  */
 public class RibbonClient implements Client {
 
+  public static RibbonClient DEFAULT = new RibbonClient();
```
adriancole: kill this and instead make the cache static. I hate both, but the latter less.

adriancole:
Pinging spring cloud folks for a second opinion. @philwebb @spencergibb @dsyer

So, here's the deal. Brendan hunted through ribbon stack traces and found that each time a LBClient is created, a call to this static beauty ends up happening.

`Monitors.registerObject("Client_" + clientName, this);`

With default config, there's a race possible leading to an exception.

The attached code caches the LBClient. Since `Monitors.registerObject` ends up doing static singleton things, the right cache scope seems also to be singleton. That implies locking, etc. but then again a lot of ribbon is as well.

So, point is this.. do you see any issue merging this code w/notes? Or do you think it is better to punt and say "don't use the default servo registration"? Or do you think there's another solution?

ribbon/src/main/java/feign/ribbon/RibbonClient.java
bstick12: We could we then decide to use the LoadingCache instead as well. As we are again back at a lock for the each request.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
bstick12: If we do got with this rather than the LoadingCache could we make this method protected?Locally I would prefer an override to use a LoadingCache?

dsyer:
I don't think we use `RibbonClient` in Spring Cloud (we duplicate some of it to provide a load balancer to our own Feign client, but we don't register anything in Servo), so I doubt this will affect us much directly. Of course there's nothing to stop users grabbing the native Feign configuration/factories instead of the ones we provide, so anything that improves things is good.

The only alternative approach I can think of would put the cache and locking in Servo, and give monitor clients the option to obtain the current object by name if it already exists.

I can see calls to `Monitors.registerObject()` elsewhere in Spring Cloud, that we should maybe look at, but that's off topic.

adriancole:
closing this and will comment in 182 why

PR #191, opened 2015-02-19
Delegate the creation of the LBClient in the RibbonClient to a LBClientF...

bstick12:
Delegate the creation of the LBClient in the RibbonClient to a LBClientFactory

### DIFF
--- a/ribbon/src/main/java/feign/ribbon/LBClient.java
+++ b/ribbon/src/main/java/feign/ribbon/LBClient.java
@@ -35,19 +35,21 @@ import feign.Request;
 import feign.RequestTemplate;
 import feign.Response;
 
-class LBClient
-    extends AbstractLoadBalancerAwareClient<LBClient.RibbonRequest, LBClient.RibbonResponse> {
+public final class LBClient extends
+    AbstractLoadBalancerAwareClient<LBClient.RibbonRequest, LBClient.RibbonResponse> {
 
-  private final Client delegate;
   private final int connectTimeout;
   private final int readTimeout;
   private final IClientConfig clientConfig;
 
-  LBClient(Client delegate, ILoadBalancer lb, IClientConfig clientConfig) {
+  public static LBClient create(ILoadBalancer lb, IClientConfig clientConfig) {
+    return new LBClient(lb, clientConfig);
+  }
+
+  LBClient(ILoadBalancer lb, IClientConfig clientConfig) {
     super(lb, clientConfig);
     this.setRetryHandler(RetryHandler.DEFAULT);
     this.clientConfig = clientConfig;
-    this.delegate = delegate;
     connectTimeout = clientConfig.get(CommonClientConfigKey.ConnectTimeout);
     readTimeout = clientConfig.get(CommonClientConfigKey.ReadTimeout);
   }
@@ -64,7 +66,7 @@ class LBClient
     } else {
       options = new Request.Options(connectTimeout, readTimeout);
     }
-    Response response = delegate.execute(request.toRequest(), options);
+    Response response = request.client().execute(request.toRequest(), options);
     return new RibbonResponse(request.getUri(), response);
   }
 
@@ -84,8 +86,10 @@ class LBClient
   static class RibbonRequest extends ClientRequest implements Cloneable {
 
     private final Request request;
+    private final Client client;
 
-    RibbonRequest(Request request, URI uri) {
+    RibbonRequest(Client client, Request request, URI uri) {
+      this.client = client;
       this.request = request;
       setUri(uri);
     }
@@ -99,8 +103,12 @@ class LBClient
           .request();
     }
 
+    Client client() {
+      return client;
+    }
+
     public Object clone() {
-      return new RibbonRequest(request, getUri());
+      return new RibbonRequest(client, request, getUri());
     }
   }
 
--- a/ribbon/src/main/java/feign/ribbon/LBClientFactory.java
+++ b/ribbon/src/main/java/feign/ribbon/LBClientFactory.java
@@ -1,0 +1,22 @@
+package feign.ribbon;
+
+import com.netflix.client.ClientFactory;
+import com.netflix.client.config.IClientConfig;
+import com.netflix.loadbalancer.ILoadBalancer;
+
+public interface LBClientFactory {
+
+  LBClient create(String clientName);
+
+  /**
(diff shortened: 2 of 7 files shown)


adriancole:
Few notes (from mobile)

 * I forgot that LBClient was a type we invented. Perhaps it is best to have the factory return what we actually need (abstractloadbalancer..) vs exposing the LBClient type. For example, you can make a caching decorator on top of the default factory without needing to know the concrete result type. In doing so, we expose the least possible interface. If there are other reasons you need LB client exposed then that's another story ;)

* RibbonClient is starting to look like it needs a builder. Instead of adding new public ctors, we probably want to make a nested Builder class at this point. There are similar classes you could look at for style reasons.

* nit.. it seems your changes reorder imports. Are you using the google style formatter?

adriancole:
Oh ic why you need to expose LBClient. Scratch bullet 1 above and replace with this. Make LBClient final and expose a static factory method instead of exposing the ctor. In general, ctors limit us. Ex we cant easily migrate to AutoValue when folks use ctors directly or can subtype our classes.

Also it seems clear enough to nest LBClientFactory as LBClient.Factory

Finally, I noticed exposing a getClient() method. Prefer client() as that's more conventional to feign.

bstick12:
I've added the Builder for the RibbonClient and marked the old ctors as deprecated to maintain backwards compatibility.

I generally have Organize Imports turned on in Eclipse. I've reverted the imports that were re-organised but I don't think there is a consistent ordering across the source files.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -78,4 +94,25 @@ public void loadDefaultValues() {
 
   }
   
+  public static class Builder {
```
adriancole: nit final

ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -35,19 +35,17 @@
...
+public final class LBClient extends
+    AbstractLoadBalancerAwareClient<LBClient.RibbonRequest, LBClient.RibbonResponse> {
 
-  private final Client delegate;
   private final int connectTimeout;
   private final int readTimeout;
   private final IClientConfig clientConfig;
 
-  LBClient(Client delegate, ILoadBalancer lb, IClientConfig clientConfig) {
+  public LBClient(ILoadBalancer lb, IClientConfig clientConfig) {
```
adriancole: hide this and expose factory method

ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -99,8 +99,12 @@ Request toRequest() {
           .request();
     }
 
+    Client client() {
```
adriancole: ps why was this change made? (from field on LBClient to field RibbonRequest) I may have lost it during the review.

ribbon/src/main/java/feign/ribbon/LBClientFactory.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Factory to be used for the creation of {@link LBClient}'s for usage in {@link RibbonClient}
+ */
+public interface LBClientFactory {
+
+  LBClient create(String clientName);
+
+  /**
+   * Uses {@link ClientFactory} static factories from ribbon to create an LBClient.
+   * */
```
adriancole: nit bad format

ribbon/src/main/java/feign/ribbon/LBClientFactory.java
```diff
@@ -0,0 +1,25 @@
+package feign.ribbon;
+
+import com.netflix.client.ClientFactory;
+import com.netflix.client.config.IClientConfig;
+import com.netflix.loadbalancer.ILoadBalancer;
+
+/**
+ * Factory to be used for the creation of {@link LBClient}'s for usage in {@link RibbonClient}
```
adriancole: I appreciate the polish, but I'm not sure this clarifies much. Let's remove javadoc that don't pull their weight.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -4,36 +4,53 @@
...
 
   private final Client delegate;
+  private final LBClientFactory lbClientFactory;
 
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * @deprecated Use the {@link RibbonClient#builder()}
```
adriancole: `RibbonClient.builder().build()` is onerous for the simple case. how about just `RibbonClient.create()`?

```java
 public static RibbonClient create() {
   return new RibbonClient.builder().build();
 }
```

adriancole:
couple last notes. nearly there! thanks for the follow-up.

wrt import order, I've used the google style format for idea and reformatted everything. I also do so on each change, and order tends to be deterministic. That said, I probably missed some files..

https://code.google.com/p/google-styleguide/source/browse/trunk/intellij-java-google-style.xml

ribbon/src/main/java/feign/ribbon/LBClient.java
bstick12: It was moved here as otherwise the LBClientFactory would have the take the delegate and the clientname and I wanted to be able to key the LBClient just off the clientName.

adriancole: ahh gotcha. thanks for the explanation!

ribbon/src/main/java/feign/ribbon/RibbonClient.java
bstick12: agreed

bstick12:
Made the suggested changes. 

The Eclipse formatter configuration doesn't control the organization of imports its done as part of a different config section.

adriancole:
eek.. just looked at the idea import settings and noticed they are very
specific, so probably easy to get wrong manually. If you find a solution to
this problem for eclipse, please share. In the mean time, I'll try to keep
this difficulty in mind when reviewing contribs from eclipse IDE.

adriancole:
good work. so I am going to reword the commit while cherry-picking in order to save you from another nit change. will update in a few.

adriancole:
closed via 222793057764d138399fff8fc857a1649ef8705a (in master) and 4f6f8d4ecd9fe6ecb9ef0abe9eb96bb02aacaf6e in (7.x)

PR #192, opened 2015-02-20
Fix NPE when building a client with a query param with no values

jebeaudet:
Using feign-core 7.1.0, when creating a feign client with a query param with no values, the builder throws a NPE. I didn't want to include Apache CollectionsUtils so I just added a `null` check. Thanks for the great library btw, hope this PR is OK!

TestApi.java: 
```java
public interface TestApi
{
    @RequestLine("POST /api/{parameter}?in_error")
    void updateInError(@Named("parameter")String parameter);
}
```

Builder : 
```java
Feign.builder()
         .decoder(new GsonDecoder())
         .encoder(new GsonEncoder())
         .logLevel(Level.FULL)
         .target(TestApi.class, "http://localhost:8888");
```

throws this stackstrace : 
```
java.lang.NullPointerException: null
	at feign.RequestTemplate.allValuesAreNull(RequestTemplate.java:494) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.RequestTemplate.pullAnyQueriesOutOfUrl(RequestTemplate.java:476) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.RequestTemplate.append(RequestTemplate.java:207) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.Contract$Default.processAnnotationOnMethod(Contract.java:137) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.Contract$BaseContract.parseAndValidatateMetadata(Contract.java:62) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.Contract$BaseContract.parseAndValidatateMetadata(Contract.java:48) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:143) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:58) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.Feign$Builder.target(Feign.java:265) ~[feign-core-7.1.0.jar:7.1.0]
	at feign.Feign$Builder.target(Feign.java:261) ~[feign-core-7.1.0.jar:7.1.0]
```

PR #193, opened 2015-02-21
Supports query params without values

jebeaudet:
Update of the PR #192 that was incomplete, thanks for the tips!

I added tests to validate the changes and discovered some more changes had to be made. Sorry about yesterday's incomplete PR, I should have added tests from the beginning.

Hope this one is OK, let me know if there is anything I can do to improve it.


adriancole:
Looks beautiful

PR #195, opened 2015-02-23
added ErrorDecoder example with nested Decoder

sfussenegger:
as discussed in https://github.com/Netflix/feign/issues/194

### DIFF
--- a/example-github/src/main/java/feign/example/github/GitHubExample.java
+++ b/example-github/src/main/java/feign/example/github/GitHubExample.java
@@ -21,6 +21,9 @@ import feign.Feign;
 import feign.Logger;
 import feign.Param;
 import feign.RequestLine;
+import feign.Response;
+import feign.codec.Decoder;
+import feign.codec.ErrorDecoder;
 import feign.gson.GsonDecoder;
 
 /**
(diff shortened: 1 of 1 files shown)


example-github/src/main/java/feign/example/github/GitHubExample.java
```diff
@@ -53,4 +64,51 @@ public static void main(String... args) throws InterruptedException {
     String login;
     int contributions;
   }
+
+  static class ClientError {
```
adriancole: would be neat if this could extend RuntimeException, that would save some boilerplate in the example.. is that possible to make without gson getting complicated?

sfussenegger: noto sure about gson. with jackson it would be just about annotating the constructor. but the JacksonDecoder needs some configuration as it won't use non-public fields by default.

adriancole: Gson should probably work as it uses sun.misc.unsafe

Give it a shot! Main goal is to show the fast path, and I think declaring
one type helps minimize distraction.

If doesn't work, I apologize in advance ;)

sfussenegger: I already had a quick shot at it without success. I haven't had the time to go into details though.

adriancole:
Looks good. can you squash into a single commit with the following message

`adds ErrorDecoder example with nested Decoder`

I know.. added vs adds :)

Just that "this commit " + works better with that tense.

sfussenegger:
done

adriancole:
Thanks!

PR #196, opened 2015-02-25
Polishes GitHub example

adriancole:
The GitHub example could be better organized as top-down. Also, it is
easier to show basic error decoding when there is less structure.

### DIFF
--- a/example-github/src/main/java/feign/example/github/GitHubExample.java
+++ b/example-github/src/main/java/feign/example/github/GitHubExample.java
@@ -15,6 +15,7 @@
  */
 package feign.example.github;
 
+import java.io.IOException;
 import java.util.List;
 
 import feign.Feign;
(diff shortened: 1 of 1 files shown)


adriancole:
cc @sfussenegger this just builds on what you did. I noticed the original example you added to could be structured a little better.

example-github/src/main/java/feign/example/github/GitHubExample.java
```diff
@@ -46,69 +66,30 @@ public static void main(String... args) throws InterruptedException {
...
     final ErrorDecoder defaultDecoder = new ErrorDecoder.Default();
 
-    public GitHubErrorDecoder(Decoder decoder) {
+    GitHubErrorDecoder(Decoder decoder) {
       this.decoder = decoder;
     }
 
+    @Override
     public Exception decode(String methodKey, Response response) {
-      if (response.status() >= 400 && response.status() < 500) {
```
adriancole: I chose to take out this detail, as the example oversimplifies anyway. key story to tell is that we can use a decoder in the error decoder.

example-github/src/main/java/feign/example/github/GitHubExample.java
```diff
@@ -27,11 +28,30 @@
...
+    List<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
+  }
+
+  static class Contributor {
+    String login;
+    int contributions;
+  }
+
+  static class GitHubClientError extends RuntimeException {
+    private String message; // parsed from json
```
adriancole: chose to only show a field we are displaying, which keeps the file short

sfussenegger:
ok, thanks for info

adriancole:
ok both are in 7.x too

PR #204, opened 2015-03-10
Allows `@Headers` to be applied to a type

### LINKED ISSUE #184: interface level header defaults
I implemented an enhanced contract that extends the default contract but adds the ability to set interface level Headers as defaults unless overridden at the method level..

I also added convenience  annotations for ContentType & Accept since those (at least in our experience) are most typically set.
The annotations follow more of the jaxrs style, but another option I could see would be adding these as settables in the Feign.Builder when creating the client..   I'm on fence as which one i like better.

Are either of these of interest as a pull?  @adriancole

adriancole:
This supports interface-level defaults, such as Content-Type.

closes #184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
+### Version 8.1
+* Allows `@Headers` to be applied to a type
+
 ### Version 8.0
 * Removes Dagger 1.x Dependency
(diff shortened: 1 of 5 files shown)


adriancole:
cc @jdamick

adriancole:
weird.. no pull request feedback, yet..

adriancole:
kicking to see if somehow the webhook missed

jdamick:
oh that's easy

PR #206, opened 2015-03-10
Travis setup

rspieldenner:


rspieldenner:
Figuring out the key encryption still

rspieldenner:
TravisCI build is going to fail until this is committed.

rspieldenner:
@adriancole I think this is ready to play with, if it works we should copy these 2 files to the other branches you release off of and then I can do the same for the other projects you're in charge of releasing.

adriancole:
@rspieldenner cool. can I cherry-pick this onto the other branches, or should I actually copy the files?

adriancole:
also.. is this the same thing for denominator or would the contents differ?

rspieldenner:
cherry-pick would be fine. I'd like to see a release (or a candidate release) on a branch before we copy everywhere to make sure I got it right. If de

Denominator the contents would be almost identical. The secure bits need generated per repo I think.

PR #208, opened 2015-03-10
Change syntax to make it clear calling a script in directory instead of ...

rspieldenner:
...an OS command


adriancole:
picked into 8.x, 7.x

adriancole:
bumped tag v7.4.0

PR #209, opened 2015-03-12
Updates to nebula 2.2.7

adriancole:


adriancole:
oops shouldn't have named the branch like this :)

PR #215, opened 2015-04-01
Enable caching for gradle dependencies

rspieldenner:
If I read their docs correctly this should cache the dependencies.


adriancole:
merging eventhough I don't know if this works yet :)

Issue #214, opened 2015-03-30
Better support for request-specific headers


chadjaros:
This is more of a question than an issue, but there is no mailing list for Feign that I could find.

From what I can tell, feign does not support request-specific headers unless they are declared as a method parameter. This is problematic from the standpoint of incorporating Feign into a set of stateless web services as a singleton bean, and wanting to optionally forward a set of headers from service to service without having the client explicitly have to deal with them.

It seems that creating feign clients per-request could be a performance concern since reflection is used to generate the proxy. I couldn't find anywhere in the Feign builder infrastructure which caches the reflected object state to improve the performance of repeated builds of the same interface with minor adjustments to the builder object.

Has any thought been put into either adding some mechanism for allowing modifications to a specific request before sending it, or reducing the cost of creating Feign instances? Either of these approaches would help me achieve my goal. The first suggestion could potentially dovetail with the ticket for asynchronous requests that I saw on the board.

Thanks,
Chad

adriancole:
Hi, Chad.

Good point on no mailing list :) I suppose if we get enough issues etc
we'll start looking at SO or maybe a mailing list.

So, it is true that one should cache instances created by feign. It
isn't impossible to make a feign compiler, and that sort of thing
would be welcome. Since feign has only one method, it shouldn't be too
hard to make a forwarding caching function.

Also, I might look at RequestInterceptor, as this is what I use in
denominator for example to decorate requests. (Also look at
denominator, as a lot of patterns are there).

Does any of this help? If not, can you give some concrete examples
you'd like to enable?

chadjaros:
Hi Adrian,

I was looking at using RequestInterceptor, but it seems like it would be less than straightforward to leverage that functionality to achieve per-request header forwarding if one were to share a single Feign instance across multiple requests. 

My plan right now is to test out the possibility of using ThreadLocal to write an interceptor capable of being hooked into Spring MVC. Since SpringMVC uses the thread-per-request model, this should theoretically allow injection of request-specific headers in my use case. This is obviously a very tailored solution for my platform, and wouldn't work for a platform based on Netty's non-blocking IO (like the Play framework). Assuming that works, I might not have an immediate need for improved performance when creating Feign instances. 

[...]

PR #217, opened 2015-04-07
Adds JMH benchmark module

adriancole:
Starts with Contract tests as this was suggested a place we may need
caching.

See #214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### Version 8.2
+* Adds JMH benchmark module
 * Enforces source compatibility with animal-sniffer
 
 ### Version 8.1
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,10 @@
+Feign Benchmarks
+===================
+
+This module includes [JMH](http://openjdk.java.net/projects/code-tools/jmh/) benchmarks for Feign.
+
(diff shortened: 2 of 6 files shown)


adriancole:
@chadjaros @spencergibb If you think this will be helpful, plus one it. We can iterate on this to tune the benchmarks (warmups, what we are measuring, etc) and isolate hot spots that may need work. FWIW, here's the results on my laptop.

```
$ mvn -Dfeign.version=8.1.0
--snip--
$ target/benchmark 
# JMH 1.8 (released 1 days ago)
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home/jre/bin/java
--snip--
Benchmark                       Mode  Cnt      Score      Error  Units
ContractBenchmarks.parseFeign  thrpt   15  12485.953 ± 681.300  ops/s
ContractBenchmarks.parseJAXRS  thrpt   15  11913.406 ± 679.382  ops/s
```

adriancole:
@jdamick I recall you had an alternate Contract idea.. this may be relevant to you, too.

spencergibb:
:+1: I'd be interested to run this against my Spring MVC Contract.

chadjaros:
:+1:  This is exactly what you'd need to compare performance on various caching strategies within a contract

PR #224, opened 2015-04-24
Correct time decoding in RetryAfterDecoder

jnaous:
currentTimeNanos was not returning nanos.


adriancole:
holy moly.. that was quite a bug! thanks for the fix!

adriancole:
lemme know if this needs to be rushed out as a subminor

jnaous:
Would be great if you can, we were looking into using that for retries, but stopped when we noticed the bug. Not yet urgent. When were you planning the next release?

adriancole:
no plan for a subminor.. probably easier to cut it than to plan it :) will
do tomorrow.

adriancole:
8.1.2!

PR #230, opened 2015-05-10
Corrects encoding of space in URL parameter

### LINKED ISSUE #225: Space in URL parameter (encoding)
The RequestTemplate encodes all spaces in URL parameters as "+" instead of "%20". Parameters in the URL should not be encoded with the URLEncoder (https://www.ietf.org/rfc/rfc2396.txt).

Example:
http://example.com/api/{value1}?key={value2}
value1=ABC 123
value2=XYZ 123
results in
http://example.com/api/ABC+123?key=XYZ+123
According to rfc2396 the value1 should be encoded as ABC%20123
http://example.com/api/ABC%20123?key=XYZ+123

A cheap fix would be the following change on Line 193 in class RequestTemplate:
String resolvedUrl = expand(url.toString(), encoded).replace("%2F", "/"); 
to 
String resolvedUrl = expand(url.toString(), encoded).replace("%2F", "/")**.replace("+", "%20");**

santhosh-tekuri:
According to rfc2396, space in url should be encoded as "%20" instead
of "+".

Closes #225

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -188,7 +188,7 @@ public final class RequestTemplate implements Serializable {
     for (Entry<String, ?> entry : unencoded.entrySet()) {
       encoded.put(entry.getKey(), urlEncode(String.valueOf(entry.getValue())));
     }
-    String resolvedUrl = expand(url.toString(), encoded).replace("%2F", "/");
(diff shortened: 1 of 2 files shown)


adriancole:
LGTM. thanks, santhosh!

dems54:
Problem always exist in 8.18.0, it's normal ?

adriancole:
It is possible that this change didn't get cherry-picked

PR #232, opened 2015-05-10
Fixes deprecated method usage of Jackson library

santhosh-tekuri:
ObjectMapper#writerWithType is depreated. Code changed to use
ObjectMapper#writerFor method.


adriancole:
:trophy:

PR #233, opened 2015-05-15
Removes trailing slash from the URL just when present

thiagomoretto:
When you have a client built like this:
```
  Feign
      .builder()
      ...
      .target(TheAPI.class, "http://the-api.com/"); // Note the trailing slash.
```
And the API defined as:
```
public interface TheAPI {

  @RequestLine("GET /api/things")
  List<Thing> things();
}
```
The URL generated when calling the API is something like that `http://the-api.com//api/things` (note the double slash). This PR removes the trailing slash from the URL passed as paramater to `target`, but only when the slash is present, so the calls are done correctly to `http://the-api.com/api/things`.

### DIFF
--- a/core/src/main/java/feign/Target.java
+++ b/core/src/main/java/feign/Target.java
@@ -66,7 +66,7 @@ public interface Target<T> {
     public HardCodedTarget(Class<T> type, String name, String url) {
       this.type = checkNotNull(type, "type");
       this.name = checkNotNull(emptyToNull(name), "name");
-      this.url = removeTrailingSlash(checkNotNull(emptyToNull(url), "url"));
+      this.url = checkNotNull(emptyToNull(url), "url");
     }
 
     @Override
--- a/core/src/main/java/feign/Target.java
+++ b/core/src/main/java/feign/Target.java
@@ -17,6 +17,7 @@ package feign;
 
 import static feign.Util.checkNotNull;
 import static feign.Util.emptyToNull;
+import static feign.Util.removeTrailingSlash;
 
 /**
  * <br><br><b>relationship to JAXRS 2.0</b><br> <br> Similar to {@code
@@ -65,7 +66,7 @@ public interface Target<T> {
     public HardCodedTarget(Class<T> type, String name, String url) {
       this.type = checkNotNull(type, "type");
       this.name = checkNotNull(emptyToNull(name), "name");
-      this.url = checkNotNull(emptyToNull(url), "url");
+      this.url = removeTrailingSlash(checkNotNull(emptyToNull(url), "url"));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


adriancole:
Thanks, Thiago. I agree it would be nice to avoid double-slash.

One feature we support (even if not well tested), is only supplying the method. Ex. `@RequestLine("GET")` I think this implementation will break users who use that. Also, this only works for hard-coded templates, but the behavior you are looking to affect is really about `@RequestLine`.

I think a better place to address double-slash is in `RequestTemplate`. When you do, check the following to make sure things still work!

```java
@RequestLine("GET")  // expands to "/"
@RequestLine("GET /api/things") // expands to "/api/things"
```

thiagomoretto:
Nice, Adrian. I wrote unit tests to cover these cases you've mentioned and fortunately I didn't break them. But you're right, `RequestTemplate` seems the best place to do the proper URL building.

I gonna check the code in details to see how to do it.

adriancole:
hah! thanks for the update, Thiago. you are right to check my assumptions :)

thiagomoretto:
Adrian, please, check the code. Instead of just removing the trailing slash from the URL, I did the proper concatenation between it and the path provided by `@RequestLine`. So, these cases also work:

```java
@RequestLine("GET")
@RequestLine("GET api/thing")
```

... and, it's regardless if the URL provided has or not the trailing slash. Also, the job is done by `RequestTemplate`, as you suggested.

core/src/main/java/feign/Target.java
```diff
@@ -17,6 +17,7 @@
 
 import static feign.Util.checkNotNull;
 import static feign.Util.emptyToNull;
+import static feign.Util.removeTrailingSlash;
```
adriancole: remove

core/src/main/java/feign/Util.java
```diff
@@ -139,6 +139,18 @@ public static String emptyToNull(String string) {
     return collection.toArray(array);
   }
 
+  public static boolean isHttpUrl(CharSequence value) {
```
adriancole: as these methods are only used once, move them to RequestTemplate.java and mark private.

adriancole:
only a couple style nits, otherwise LGTM! Thanks for being thorough.

thiagomoretto:
Done! Thanks for the tips.

adriancole:
Hi, Thiago.

These commits need to be rebased and squashed into a single commit. You'll probably want to do that on a branch besides master, but I don't which branch you are using so long as this ends up as one commit.

Alternatively, I could rebase and squash these on my end into one commit (still owned by you).

Let me know which path you want to take, then lets get this in!

adriancole:
ping! I can squash these into a single commit for you, if you like.

thiagomoretto:
pong! It would be nice if you can. Thanks!

adriancole:
closed via #0bc994e7ca418df1c3138519d978197c00333d8a

Issue #189, opened 2015-02-18
feign with hystrix


jdamick:
I'd like to use hystrix with feign, but it appears i'd have to shoehorn it into the invocation handler factory, and if i wanted the ability to do things async it would be more changes to methodmetadata to look for Future<?> return types..    
I could do this on the outside of the feign, but that seems kludgy so i wanted to see if there were any better ideas floating around?

adriancole:
We could start (like beta) with invocation handler.

I suspect any async stuff (rx or hystrix) would imply contract extensions
of some sort and possibly also deeper looks or refactoring around
invocation handler.

I had considered a decorator feign when thinking about rxjava. Ex

new HystrixFeign.Builder () which intercepts or weaves things properly so
as to make tricky things easier and also availing the integration to the
feign package to access what it needs (either by package accessor or
literally being in feign package). With a builder like this, we could make
any details like invocationhandler implicit and also decide which contracts
to support, in case we haven't implemented extensions yet.

Make sense?

jdamick:
Yes, that's similar to what i was thinking.  That would make rxjava and/or hystrix or whatever else a lot easier to add as a bolt-on. I wonder if makes sense to add a builder "interceptor", something analogous to the requestInceptors, that would let you just pluign into the Feign.Builder but keep that same style of registering add-ons?

adriancole:
Which bolt-ons do you think you would use together with the same feign
instance? One concern about bolt-on builder interceptor is whether or not
bolt-ons are compatible. Ie it is easier to support hystrix or rxjava vs
both

spencergibb:
I had done some work on this a while ago [here: HystrixInvocationHandler](https://github.com/spencergibb/halfpipe/blob/master/halfpipe-client/src/main/java/halfpipe/client/HystrixInvocationHandler.java).  It required some funkyness in a [decoder](https://github.com/spencergibb/halfpipe/blob/master/halfpipe-client/src/main/java/halfpipe/client/HystrixJacksonDecoder.java).  I'm sure something cleaner can be worked out.

adriancole:
Thanks for the tips, Spencer!

rubesMN:
Are there any updates to the progress on this?   Would be awesome to combine Hystrix annotations with feign's integrations with ribbon into spring cloud.   How does 9.0.0 fit into Spring Cloud's labeling/releases?

spencergibb:
[...]

PR #238, opened 2015-06-03
Allow for customization of statusCode interpretation

nblair:
We just started looking at Feign for implementing client libraries for some REST APIs.

We wanted to setup an interface like so:

```java
public interface FooAPI {

  /**
   * Retrieve the {@link Foo} that has a matching value for {@link Foo#getId()}.
   * 
   * @param id the id of the {@link Foo}
   * @return the corresponding {@link Foo}, or null if not found
   */
 @RequestLine("GET /api/v1/foo/{id}")
 Foo retrieveById(@Param("id") String id);
```

The REST API's contract states that `/api/v1/foo/{id}` will return a 404 if no resource exists with that id.

In the current 8.3 release of Feign, a return code of 404 is treated as something that should raise an Exception. Our method signature would have to look like this instead:

```java
public interface FooAPI {

  /**
   * Retrieve the {@link Foo} that has a matching value for {@link Foo#getId()}.
   * 
   * @param id the id of the {@link Foo}
   * @return the corresponding {@link Foo}
   * @throws FooNotFoundException if no {@link Foo} exists with the id
   */
 @RequestLine("GET /api/v1/foo/{id}")
 Foo retrieveById(@Param("id") String id) throws FooNotFoundException;
```

We'd prefer not to have to create an exception for this scenario.
This pull request adds a new interface encapsulating logic currently at line 106 of:

https://github.com/Netflix/feign/blob/8.x/core/src/main/java/feign/SynchronousMethodHandler.java

Previously, inline
logic in SynchronousMethodHandler only treated 200 class status codes as
something to parse; everything else, including 404, as reason to throw exception (ErrorDecoder then takes over).

This change allows callers to inject StatusInterpreter.Include404 if they desire a contract that can return null for 404 like the one presented at the top rather than raising an Exception. The interface allows for further customization if desired.

The default behavior is unchanged.

### DIFF
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -84,6 +84,7 @@ public abstract class Feign {
     private InvocationHandlerFactory
         invocationHandlerFactory =
         new InvocationHandlerFactory.Default();
+    private StatusInterpreter statusInterpreter = new StatusInterpreter.Default();
 
     public Builder logLevel(Logger.Level logLevel) {
       this.logLevel = logLevel;
@@ -157,6 +158,18 @@ public abstract class Feign {
       this.invocationHandlerFactory = invocationHandlerFactory;
       return this;
     }
+    /**
+     * Allows you to override the {@link StatusInterpreter}.
+     * The default interpreter treats any non-200 class status code
+     * as a reason to raise exception. Use this to provide alternative
+     * logic (like allowing 404 to gracefully return null).
+     * 
+     * @see StatusInterpreter.Include404
+     */
+    public Builder statusInterpreter(StatusInterpreter statusInterpreter) {
+      this.statusInterpreter = statusInterpreter;
+      return this;
+    }
 
     public <T> T target(Class<T> apiType, String url) {
       return target(new HardCodedTarget<T>(apiType, url));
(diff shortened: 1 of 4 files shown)


adriancole:
gonna sleep on this. I admit that I've a lot of catch blocks in denominator
about this, which boils down to a fallback value, most usually fallback to
null, void, or false.

adriancole:
So, this is definitely in the area of static-simple fallbacks, somewhat related to the hystrix discussion.

When designing feign, originally, the error decoder had fallback handling as a part of it. This was inspired by prior art in [jclouds](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/Fallback.java), intended to be a bit simpler.

When doing an implementation with an Async pattern (Observer), we realized that this fallback handling was something invocation-specific, so in order to reuse the ErrorDecoder, we had to decouple it from fallbacks (46be6bd)

If we were to resurrect fallback handling, I'd prefer to not conflate it with the decoder. For example, if fallback processing occurs with context of the request, it could be statically mapped (ex. 404 -> default value or null), or others could choose to compose one out of a decoder.

For now, I'd recommend either working off your fork or abusing InvocationHandlerFactory to supply default values :) It is likely something related will come out of #189 so I'd likely watch there.

PR #240, opened 2015-06-12
Fix an issue with trailing slashes in @Path annotations on classes

jacob-meacham:
This fixes what appears to be a regression from 6.x to 7.x/8.x. If I have this:
```
@Path("/")
class MyClass {
    @Path("base")
    void get();
}
```
Then feign will try to send "//base", instead of just "/base". This also fixes the case where I have:
```
@Path("/api/")
class MyClass {
    @Path("base")
    void get();
}
```
which would try to hit /api//base instead of /api/base.


adriancole:
Jacob, we really appreciate your help!

jacob-meacham:
@adriancole thanks for the prompt response! Is there an ETA on the release of 8.4.0?

adriancole:
how does an hour sound? https://travis-ci.org/Netflix/feign/builds/66567332

PR #241, opened 2015-06-12
Strip regex allowed by the jaxrs spec.

jacob-meacham:
Jax-rs allows for a regex to appear at the end of an @Path annotation (https://jersey.java.net/documentation/latest/jaxrs-resources.html#d0e1961). This lets the JaxRsContract deal with that appropriately


adriancole:
thanks for the fix!

PR #242, opened 2015-06-12
Fix regex support when there are multiple @Path regexes

jacob-meacham:
My last PR had a bug in it when dealing with multiple path params that have regexes in them. This PR fixes that, and also finishes off implementing the jax-rs spec here https://jax-rs-spec.java.net/nonav/2.0/apidocs/javax/ws/rs/Path.html by stripping spaces before and after a param name.


adriancole:
thanks for the quick turnaround, pushing 8.4.1 shortly!

Issue #227, opened 2015-05-07
Why is the '/' character (%2F) decoded in the RequestTemplate ?


MaximeGaudin:
The RequestTemplate's method resolve does not allow to have encoded '/' : 
`String resolvedUrl = expand(url.toString(), encoded).replace("%2F", "/");`

Why is that ? How can I send encoded slash an other way ?

adriancole:
Good question, you can do this now, but only if you make a target in the feign package.

#228 makes it easier. Will merge and release 8.2 shortly.

MaximeGaudin:
Awesome ! Thanks for the quick response and the incredible work  :)

adriancole:
we aim to please! thanks for contributing feature ideas!

PR #243, opened 2015-06-16
Added possibility to leave slash encoded in path parameters

gdenhez:
We need that to communicate with rabbitMQ api using feign.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -168,6 +168,9 @@ public interface Contract {
           data.template().append(
               requestLine.substring(requestLine.indexOf(' ') + 1, requestLine.lastIndexOf(' ')));
         }
+        
+        data.template().decodeSlash(RequestLine.class.cast(methodAnnotation).decodeSlash());
+        
       } else if (annotationType == Body.class) {
         String body = Body.class.cast(methodAnnotation).value();
         checkState(emptyToNull(body) != null, "Body annotation was empty on method %s.",
--- a/core/src/main/java/feign/RequestLine.java
+++ b/core/src/main/java/feign/RequestLine.java
@@ -47,4 +47,5 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 public @interface RequestLine {
 
   String value();
+  boolean decodeSlash() default true;
 }
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -58,6 +58,7 @@ public final class RequestTemplate implements Serializable {
   private transient Charset charset;
   private byte[] body;
   private String bodyTemplate;
+  private boolean decodeSlash = true;
 
   public RequestTemplate() {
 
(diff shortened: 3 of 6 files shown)


core/src/main/java/feign/EncodeSlash.java
```diff
@@ -0,0 +1,13 @@
...
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface EncodeSlash {
```
spencergibb: Slightly misleading, since it isn't actually encoding the slash, it's just not decoding it.  Should it be a full annotation?  How about just an attribute on `RequestLine`?

gdenhez:
Thanks for the review! I made the changes you suggested.

adriancole:
hey there. Thanks for putting effort around this. We do want rabbitmq
or other things that don't follow normal url encoding to be
supportable.

So, this is perhaps a dupe of #227 (depending on POV)

Custom encoding can be accomplished with a custom target, though it
may be the case the needed commits aren't in the 7.x branch.

https://github.com/Netflix/feign/blob/master/core/src/test/java/feign/TargetTest.java#L51

I generally feel off about adding a one-off knob as the path can be
composed from places besides the requestline value. For example, it
may be that we skip the slash part of url encoding in general, similar
to apache's AllowEncodedSlashes directive.
http://httpd.apache.org/docs/2.2/mod/core.html#allowencodedslashes

What are your thoughts about this?

gdenhez:
Thanks a lot for your reply. The custom request is indeed not available in the 7.x branch but we may switch to version 8 then.

However, I'm not sure #227 fix our problem, but maybe I don't fully understand how custom request works. My problem is that, at the time my custom request is called, the request path is already resolved with the slash decoded. For example, let's say that my request is:

@RequestLine("GET /{path}/{path2}?query={query}")
    Response get2(@Param("path") String path, @Param("path2") String path2, @Param("query") String query);

And I call it like this:

Feign.builder().target(custom).get2("slash/foo", "slash/foofoo", "slash/bar");

then in method public Request apply(RequestTemplate input) {...
input.url() already has the value "/slash/foo/slash/foofoo"

I can't think of a clean way to get /slash%2Ffoo/slash%2Ffoofoo instead...

Thanks again!

adriancole:
Wow.. this is an old issue! There's no history of anyone except me touching
the %2F decoding thing..

https://github.com/Netflix/denominator/commit/488e4ccbca2160f513a9810510c072b384d7672f

So, I agree.. the scope of this is the request template, and as far as I
can tell, only the requestline component of it. Your current implementation
is correct and most appropriate. Good work.

Mind squashing into one commit and adding to CHANGES file?

adriancole:
picked into master and 8.x; cut 7.5.0 and 8.5.0

adriancole:
we are honored to merge your first commit to github :) thanks!

gdenhez:
Thanks a lot!

Issue #247, opened 2015-07-09
end-of-input exception in the JacksonDecoder


bcrowcs:
If an null/empty stream is returned in a body where a JacksonDecoder is used, you can receive an error when the ObjectMapper attempts to read the inputStream value on line 55. 
A quick check to inputSteam.available > 0 seems to stop this problem for me.

adriancole:
wanna give a go at a pull request? I can help merge/release it

bcrowcs:
Just put the wrong issue number on the pull request. It is attached to 246 instead of 247.

PR #248, opened 2015-07-09
Issue #247

bcrowcs:
No content to map due to end-of-input when decoding empty returns using JacksonDecoder.
Added UnitTests to show JacksonDecoder error. Added unit test to GsonDecoder to verify behavior is as expected.
Added a check for available inputStream data before trying to read into an Object.

### DIFF
--- a/jackson/src/test/java/feign/jackson/JacksonCodecTest.java
+++ b/jackson/src/test/java/feign/jackson/JacksonCodecTest.java
@@ -97,6 +97,15 @@ public class JacksonCodecTest {
   }
 
   @Test
+  public void emptyBodyDecodesToNull() throws Exception {
+    Response
+        response =
+        Response
+            .create(204, "OK", Collections.<String, Collection<String>>emptyMap(), "".getBytes());
+    assertNull(new JacksonDecoder().decode(response, String.class));
+  }
+
+  @Test
   public void customDecoder() throws Exception {
     JacksonDecoder decoder = new JacksonDecoder(
         Arrays.<Module>asList(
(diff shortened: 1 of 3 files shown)


jackson/src/test/java/feign/jackson/JacksonCodecTest.java
```diff
@@ -97,6 +97,15 @@ public void nullBodyDecodesToNull() throws Exception {
   }
 
   @Test
+  public void emptyBodyDecodesToNull() throws Exception {
+    Response
+        response =
+        Response
+            .create(204, "OK", Collections.<String, Collection<String>>emptyMap(), "".getBytes());
```
adriancole: nit. `new byte[0]` is quicker here and above

adriancole:
just a nit. otherwise, thanks tons for doing this!

also reword the commit title as `Fixes end-of-input exception when decoding an empty json body". That saves some navigation when scrolling through git history to figure out the intent of a change.

adriancole:
I noticed that this pull request was raised from master. I'll fix it up for you. After this merged, I'd recreate my fork (delete then re-fork) and create a branch for work in the future. (ex. `git checkout -b fix-npe`).

No worries in the mean time!

adriancole:
closed via 1f0464eba1b2dc1af3fe7f265247fc3f534fd9ca

bcrowcs:
Thanks @adriancole ! I am sorry for the incoherence in the ticket yesterday.

jackson/src/test/java/feign/jackson/JacksonCodecTest.java
bcrowcs: If I were reviewing the code, I probably would have had the same comment. Thanks.

adriancole:
Sorry.. we need to revert this. `InputStream.available()` is not a reliable guard. I should have thought about this earlier.

PR #253, opened 2015-08-02
Avoids InputStream.available when determining if a stream is empty

### LINKED ISSUE #250: Setting logLevel() to BASIC breaks functionality in 8.6.0
Following works in 8.5.0 but not in 8.6.0, but if I change to .logLevel(Level.FULL) it works with 8.6.0 too.

    client = Feign.builder()
        .encoder(new JacksonEncoder(mapper))
        .decoder(new JacksonDecoder(mapper))
        .client(new feign.okhttp.OkHttpClient(okHttpClient))
        .retryer(new Retryer.Default(0, 0, 1)) // do not retry calls
        .logger(new Slf4jLogger())
        .logLevel(Level.BASIC)
        .target(IReCaptchaClient.class, url);

In 8.6.0 call to client.someMethod() returns null with Level.BASIC but works as expected with Level.FULL. No other changes.

adriancole:
InputStream.available isn't a reliable api. This uses an alternative
approach, which is to read the first byte to see if it is present. This
allows us to continue to avoid "No content to map due to end-of-input"
errors, but in a more supportable way.

Fixes #250

### DIFF
--- a/jackson/src/main/java/feign/jackson/JacksonDecoder.java
+++ b/jackson/src/main/java/feign/jackson/JacksonDecoder.java
@@ -20,8 +20,9 @@ import com.fasterxml.jackson.databind.Module;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.RuntimeJsonMappingException;
 
+import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


adriancole:
Note: I am trying locally to see if I can find an alternative to reverting.

adriancole:
cc @bcrowcs @maximdim

PR #256, opened 2015-08-06
Fixes NPE when apache client rebuffers content

### LINKED ISSUE #255: NullPointerException with DEBUG logger level
Repro:

    1. init feign with:
        Feign.builder()
            .logger(new Slf4jLogger())
            .logLevel(feign.Logger.Level.FULL)

    2. set the level of Slf4jLogger to DEBUG

    3. call a method without response (204, No Content), then
    we would see the exception.

StackTrace:

    java.lang.NullPointerException: in
        at feign.Util.checkNotNull(Util.java:99)
        at feign.Util.toByteArray(Util.java:212)
        at feign.Logger.logAndRebufferResponse(Logger.java:94)
        at feign.slf4j.Slf4jLogger.logAndRebufferResponse(Slf4jLogger.java:62)
        at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:103)
        at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:71)
        at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:94)
        at com.sun.proxy.$Proxy27.deleteCredential(Unknown Source)
        at ...

Problem positioning:

    feign.Logger#logAndRebufferResponse(), line 94
    
        byte[] bodyData = Util.toByteArray(response.body().asInputStream());

    The call response.body().asInputStream() returns null.

Affected versions: 

    8.5.0, 8.6.0 (only these two versions were tested.)

adriancole:
When log level is full, the response body is rebuffered. The Apache
client had a bug where it allowed `toInputStream` to return null. This
fixes that bug and backfills tests for the other two clients.

Fixes #255

### DIFF
--- a/core/src/test/java/feign/client/DefaultClientTest.java
+++ b/core/src/test/java/feign/client/DefaultClientTest.java
@@ -34,6 +34,7 @@ import feign.Client;
 import feign.Feign;
 import feign.FeignException;
 import feign.Headers;
+import feign.Logger;
 import feign.RequestLine;
 import feign.Response;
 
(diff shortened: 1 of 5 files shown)


core/src/test/java/feign/client/DefaultClientTest.java
```diff
@@ -151,6 +152,39 @@ public void retriesFailedHandshake() throws IOException, InterruptedException {
     assertEquals(2, server.getRequestCount());
   }
 
+  @Test
+  public void safeRebuffering() throws IOException, InterruptedException {
+    server.enqueue(new MockResponse().setBody("foo"));
+
+    TestInterface api = Feign.builder()
+        .logger(new Logger(){
```
adriancole: j.u.Logger, SLF4J and others conditionally call logAndRebufferResponse based on the framework-specific log level. That's why there's an anonymous logger specified here.

PR #267, opened 2015-09-08
Un-abstracts processAnnotationOnClass so that we don't break api

adriancole:


adriancole:
slight revision to #266 as I noticed spring-cloud-netflix broke

PR #268, opened 2015-09-08
264 body decoding

### LINKED ISSUE #264: Body Expansion Incorrectly "Decoding" Unencoded Values?
Question/possible issue here.

Attempting to do a POST with a very simple body expansion and it seems like there's an issue line with the RequestTemplate body expansion.  Specifically here:

`body(urlDecode(expand(bodyTemplate, unencoded)));`

It seems like urlDecoding unencoded parameters can cause issues with certain strings.  For instance, if I have a '+' character in the argument being substituted, the urlDecode would swap the '+' for ' ' (this is the case I'm running into).  I imagine you would run into issues with '%<...>' in substituted strings as well.

I can't get my head around why just using `encoded`, the already encoded map, here wouldn't be the right solution.  It would expand into a fully encoded string (including anything encoded in the body template) and then properly decode to the full body.  

Is there a simple way to work around this during Feign building?  Swapping out the RequestTemplate for one that does the correct encoding doesn't seem very tenable -- looks like you'd have to swap all the way out in the contract and dupe a fair amount of code, since the RequestTemplate is instantiated inline with the MethodMetadata, and MethodMetadata is instantiated inline in the BaseContract.  Anything obvious I'm missing here?

ddteeter:
Fixes #264 

Switches body parameter substitution to use encoded parameters.  Resolves issue where some characters (+ %..) would end up double decoded if existing in parameters.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -229,7 +229,7 @@ public final class RequestTemplate implements Serializable {
     headers.clear();
     headers.putAll(resolvedHeaders);
     if (bodyTemplate != null) {
-      body(urlDecode(expand(bodyTemplate, unencoded)));
+      body(urlDecode(expand(bodyTemplate, encoded)));
     }
     return this;
   }
--- a/core/src/test/java/feign/RequestTemplateTest.java
+++ b/core/src/test/java/feign/RequestTemplateTest.java
@@ -198,6 +198,26 @@ public class RequestTemplateTest {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


adriancole:
denominator tests pass.. so I think we're good. nice fix!

mind squashing these into a single commit with message

```
Switches body parameter substitution to use encoded parameters

Resolves issue where some characters (+ %..) would end up double decoded if existing in parameters.

Fixes #264
```

adriancole:
gracias!

adriancole:
out in the wild, in v8.9.0

adriancole:
thanks for the help with this!

Issue #249, opened 2015-07-12
Allow for 3xx range headers


clatko:
Everything above 299 is considered an error, which is plainly incorrect.

To get around this I had to:
1) add an error decoder
2) throw a special exception for 3xx range
3) when catching runtime errors, check if it is the special 300 error
4) convert to a valid response entity

Will investigate replacing this line:

if (response.status() >= 200 && response.status() < 300) {
in SynchronousMethodHandler and PR'ing it. I just wonder why nobody else has come across this issue.

Should I fix it?

adriancole:
The Client implementation (ex Client.Default) is responsible for addressing
the 3xx range. It is considered an error when the Client propagates a 300
as it means it has exhausted its means to address redirects for example.

What scenario are you encountering where this became an issue (are you
using a non-default client?)

Best first action is to move the clarification about 300 to javadoc on the
Client interface.

clatko:
I am not using a non-default client. The redirect doesn't work because it crosses from http to https and java refuses to allow that.

This is a good thing.

I don't agree that if the client can't/doesn't handle a 300 it is an error. The proper response from the microservice is a 302 and I want feign to propagate this so the front end can do the redirect. With the default client using `setInstanceFollowRedirects(true)` you are assuming the 30x is handled and the only way to handle this is to get a 200 range response.

So to get the results I want, I have to either:

a) send a 200 back to feign from the microservice, intercept it and make make it a 302.
b) create my own client that uses `setInstanceFollowRedirects(false)`, then change the status code to something feign can swallow and break a bunch of RFCs in the process.

c) Modify this section of the code to something reasonable:
`if (response.status() >= 200 && response.status() < 300) {`

adriancole:
Hi, Chris.

My intent was to explain why this slipped for 3 years. My suggestion to
document that was not an alternative to changing things, just that would be
something smart to do and I could safely cherry-pick on pretty much all
branches. I didn't mean to sound defensive if I did.

I agree it should be fixed. Wanna raise a pull request with a mockwebserver
test that sends a redirect to an https endpoint?

adriancole:
ps another reason why this probably went unnoticed might be the scenario you described as going from http -> https. Since feign is usually used for apis, I suspect most have their Target as https in the first place.

adriancole:
Sorry.. I was on the plane when trying to comment to most of this, just catching up in detail now.

[...]

PR #269, opened 2015-09-09
Skips error handling when return type is Response

adriancole:
In unsuccessful scenarios, such as redirects, error handling converts a
`Response` into an exception. When a user defines a return type as
`Response`, we can assume they will want access to things like headers
even in an error scenario.

See #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 8.9
+* Skips error handling when return type is `Response`
+
 ### Version 8.8
 * Adds jackson-jaxb codec
 * Bumps dependency versions for integrations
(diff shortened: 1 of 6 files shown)


adriancole:
@sbuettner PTAL!

sbuettner:
Looks good! :+1:

adriancole:
@sbuettner 8.9.0 includes this.. lemme know if you are able to get things done now

sbuettner:
Is it published?

adriancole:
yep. just maven's search cache doesn't see it, yet
https://repo1.maven.org/maven2/com/netflix/feign/feign-core/8.9.0/

sbuettner:
Works! Thank you for your time. :smile:

PR #271, opened 2015-09-10
send empty response body for POST when no body parameter present

pnepywoda:
see
http://johnfeng.github.io/blog/2015/06/30/okhttp-updates-post-wouldnt-be-allowed-to-have-null-body/

### DIFF
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -18,6 +18,7 @@ package feign;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -143,7 +144,7 @@ public class ReflectiveFeign extends Feign {
         } else if (md.bodyIndex() != null) {
           buildTemplate = new BuildEncodedTemplateFromArgs(md, encoder);
         } else {
-          buildTemplate = new BuildTemplateByResolvingArgs(md);
+          buildTemplate = new BuildTemplateByResolvingArgs(md, encoder);
         }
         result.put(md.configKey(),
                    factory.create(key, md, buildTemplate, options, decoder, errorDecoder));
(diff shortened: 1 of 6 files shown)


core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -201,6 +204,11 @@ public RequestTemplate create(Object[] argv) {
 
     protected RequestTemplate resolve(Object[] argv, RequestTemplate mutable,
                                       Map<String, Object> variables) {
+      if ("POST".equals(metadata.template().method()) && argv == null) {
+        // write an empty string for the BODY to conform with okhttp 2.4.0+
+        // http://johnfeng.github.io/blog/2015/06/30/okhttp-updates-post-wouldnt-be-allowed-to-have-null-body/
+        encoder.encode("", String.class, mutable);
```
spencergibb: What happens with other clients? Will this work?

pnepywoda: What other clients? You mean ones that don't use ReflectiveFeign? We only use the JAXRSContract so that's the targeted fix that I made here. I didn't test out other contracts.

spencergibb: I mean non-OkHttpClient, such as ribbon or apache http client.

pnepywoda: Oh good point. My followup commit should fix it to only write an empty body for okhttp

spencergibb:
:+1: cool, thanks!

adriancole:
Looks safe. Few things; one is a nit.

* Why are we adding the encoder as a constructor param? It isn't used in
this PR.

* Add a test for Client.Default?

* update CHANGES? Probably there are folks who can undo similar logic
knowing this.
* new byte[0] is less code format sensitive than { }

Otherwise, LGTM!

adriancole:
ping! I'd like to cut a release today, so if you're ready, you can be in it!

pnepywoda:
Huh this is strange, it says there's a conflict in CHANGELOG.md but I can't figure out why or what the merge conflict actually is. @adriancole any ideas?

adriancole:
you're later that upstream.. no big deal, I can rebase this on the way in. Thanks!

adriancole:
in 8.10.0!

uschi2000:
Awesome stuff, thanks Adrian & Paul!

PR #280, opened 2015-10-16
add HystrixCommand support

spencergibb:
fixes gh-189

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,16 @@ MyService api = Feign.builder().client(RibbonClient.create()).target(MyService.c
 
 ```
 
+### Hystrix
+[HystrixFeign](https://github.com/Netflix/feign/tree/master/hystrix) configures circuit breaker support provided by [Hystrix](https://github.com/Netflix/Hystrix).
+
+To use Hystrix with Feign, add the Hystrix module to your classpath. Then use the `HystrixFeign` builder:
+
+```java
+MyService api = HystrixFeign.builder().target(MyService.class, "https://myAppProd");
+
+```
+
 ### SLF4J
 [SLF4JModule](https://github.com/Netflix/feign/tree/master/slf4j) allows directing Feign's logging to [SLF4J](http://www.slf4j.org/), allowing you to easily use a logging backend of your choice (Logback, Log4J, etc.)
 
--- a/core/src/main/java/feign/MethodMetadata.java
+++ b/core/src/main/java/feign/MethodMetadata.java
@@ -50,7 +50,7 @@ public final class MethodMetadata implements Serializable {
     return configKey;
   }
 
-  MethodMetadata configKey(String configKey) {
+  public MethodMetadata configKey(String configKey) {
     this.configKey = configKey;
     return this;
   }
@@ -59,7 +59,7 @@ public final class MethodMetadata implements Serializable {
     return returnType;
   }
 
-  MethodMetadata returnType(Type returnType) {
+  public MethodMetadata returnType(Type returnType) {
     this.returnType = returnType;
     return this;
   }
@@ -68,7 +68,7 @@ public final class MethodMetadata implements Serializable {
     return urlIndex;
   }
 
-  MethodMetadata urlIndex(Integer urlIndex) {
+  public MethodMetadata urlIndex(Integer urlIndex) {
     this.urlIndex = urlIndex;
     return this;
   }
@@ -77,7 +77,7 @@ public final class MethodMetadata implements Serializable {
     return bodyIndex;
   }
 
-  MethodMetadata bodyIndex(Integer bodyIndex) {
+  public MethodMetadata bodyIndex(Integer bodyIndex) {
     this.bodyIndex = bodyIndex;
     return this;
   }
@@ -89,7 +89,7 @@ public final class MethodMetadata implements Serializable {
     return bodyType;
   }
 
-  MethodMetadata bodyType(Type bodyType) {
+  public MethodMetadata bodyType(Type bodyType) {
     this.bodyType = bodyType;
     return this;
   }
--- a/hystrix/README.md
+++ b/hystrix/README.md
@@ -1,0 +1,40 @@
+Hystrix
+===================
+
+This module wraps Feign's http requests in [Hystrix](https://github.com/Netflix/Hystrix/), which enables the [Circuit Breaker Pattern](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern).
+
+To use Hystrix with Feign, add the Hystrix module to your classpath. Then, configure Feign to use the `HystrixInvocationHandler`:
+
+```java
+GitHub github = HystrixFeign.builder()
+        .target(GitHub.class, "https://api.github.com");
+```
+
+Methods that do *not* return [`HystrixCommand`](https://netflix.github.io/Hystrix/javadoc/com/netflix/hystrix/HystrixCommand.html) are still wrapped in a `HystrixCommand`, but `execute()` is automatically called for you.
+
+For asynchronous or reactive use, return `HystrixCommand<YourType>` rather than just `YourType`.
+
+```java
+interface YourApi {
+  @RequestLine("GET /yourtype/{id}")
+  HystrixCommand<YourType> getYourType(@Param("id") String id);
+  
+  @RequestLine("GET /yourtype/{id}")
+  YourType getYourTypeSynchronous(@Param("id") String id);
+}
+
+YourApi api = HystrixFeign.builder()
+                  .target(YourApi.class, "https://example.com");
+
+// for reactive
+api.getYourType("a").toObservable();
+
+// for asynchronous
+api.getYourType("a").queue();
+
+// for synchronous
+api.getYourType("a").execute();
+
+// or to apply hystrix to existing feign methods.
+api.getYourTypeSynchronous("a");
+``` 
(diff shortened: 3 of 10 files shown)


core/src/main/java/feign/InvocationHandlerFactory.java
```diff
@@ -33,6 +33,7 @@
   interface MethodHandler {
 
     Object invoke(Object[] argv) throws Throwable;
+    MethodMetadata getMethodMetadata();
```
spencergibb: Required to change the return type to the actual type to send on to the decoder. If the return type is `HystrixCommand<User>`, `User` is set as the return type.

core/src/main/java/feign/MethodMetadata.java
```diff
@@ -59,7 +59,7 @@ public Type returnType() {
     return returnType;
   }
 
-  MethodMetadata returnType(Type returnType) {
+  public MethodMetadata returnType(Type returnType) {
```
spencergibb: Maybe make the setters protected and I create a sub-class I can use.

core/src/main/java/feign/InvocationHandlerFactory.java
adriancole: interesting..

adriancole:
wondering out loud.. Seems the problem is trying to work within the invocation handler. What if you subclass `Feign.Builder`?

ReflectiveFeign's constructor has everything you need (I think) to ensure the right type gets into MethodMetadata in the first place.

Not that this would work, but.. new HystrixedFeignBuilder()...not setting invocation factory.. build()

spencergibb:
So move the mangling the methodMetadata to the new builder and automatically set the invocation handler there?

adriancole:
or maybe add a method to InvocationHandlerFactory, exposing the setters public as you mentioned.

```
MethodMetadata InvocationHandlerFactory.decorate(MethodMetadata in)
```

adriancole:
Either way, I suppose.. with Hystrix you don't need a new builder per-se.. gets more complicated with things like RxJava which need to set multiple builder methods to work. Just thinking if we should continue with invocationhandlerfactory as a builder method, or basically say, make a new builder if you need to set a invocationhandlerfactory. wdyt?

spencergibb:
Not sure. I could knock both out and see what they look like. Then we could compare?

adriancole:
yeah.. somewhere in the scope of build wire in the ability to decorate
methodmetadata somewhere between contract.parseAndValidatateMetadata
and where it is actually used. Ex. if you override the builder, you
could make a filtering contract, which manipulates the methodmetadata.

ex.

```
public Builder contract(Contract contract) {
   this.contract = new FirstTypeIsReturnValContract(contract);
   return this;
}

FirstTypeIsReturnValContract implements Contract {

override parseAndValidatateMetadata
filter return and toss the enclosing parameterized type

```

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -0,0 +1,92 @@
...
+      for (Map.Entry<Method, MethodHandler> entry : this.dispatch.entrySet()) {
+        if (HystrixCommand.class.isAssignableFrom(entry.getKey().getReturnType())) {
+          MethodMetadata methodMetadata = entry.getValue().getMethodMetadata();
+          Type type = methodMetadata.returnType();
+          Type actualType = ((ParameterizedType) type).getActualTypeArguments()[0];
+          methodMetadata.returnType(actualType);
+        }
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
jdamick: shouldn't it logged not printed?

adriancole: heh, I think this is going to get buffed out

core/src/main/java/feign/InvocationHandlerFactory.java
adriancole: still needed?

spencergibb: Nope, good catch.

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+    HystrixCommand<Object> hystrixCommand = new HystrixCommand<Object>(setter) {
+      @Override
+      protected Object run() throws Exception {
+        try {
+          return HystrixInvocationHandler.this.dispatch.get(method).invoke(args);
+        } catch (Exception e) {
+          throw e;
+        } catch (Throwable t) {
+          throw new Exception(t);
```
adriancole: `throw (Error) t`, as that's the only possible outcome besides Exception

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -62,6 +62,10 @@ private SynchronousMethodHandler(Target<?> target, Client client, Retryer retrye
     this.decoder = checkNotNull(decoder, "decoder for %s", target);
   }
 
+  public MethodMetadata getMethodMetadata() {
```
spencergibb: removing

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
spencergibb: will do

hystrix/README.md
```diff
@@ -0,0 +1,40 @@
...
+// for reactive
+api.getYourType("a").toObservable();
+
+// for asynchronous
+api.getYourType("a").queue();
+
+// for synchronous
+api.getYourType("a").execute();
+
+// OR
```
adriancole: or to apply hystrix to existing feign methods.

adriancole: something like..

`// plain methods will inherit hystrix synchronous behavior`

hystrix/src/main/java/feign/hystrix/HystrixDelegatingContract.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.List;
+
+import com.netflix.hystrix.HystrixCommand;
+
+import feign.Contract;
+import feign.MethodMetadata;
+
+public class HystrixDelegatingContract implements Contract {
```
adriancole: s/public/final unless you think folks should be extending this or using it directly

hystrix/src/main/java/feign/hystrix/HystrixFeign.java
```diff
@@ -0,0 +1,24 @@
+package feign.hystrix;
+
+import feign.Contract;
+import feign.Feign;
+
+public class HystrixFeign {
```
adriancole: final this and the builder? maybe needs javadoc saying that this allows your interfaces to return a HystrixCommand and decorates plain methods.

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.netflix.hystrix.HystrixCommand;
+import com.netflix.hystrix.HystrixCommandGroupKey;
+import com.netflix.hystrix.HystrixCommandKey;
+
+import feign.InvocationHandlerFactory;
+import feign.InvocationHandlerFactory.MethodHandler;
+import feign.Target;
+
+public class HystrixInvocationHandler implements InvocationHandler {
```
adriancole: s/public/final unless you plan to use this outside the builder entrypoint

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -0,0 +1,76 @@
...
+      }
+    };
+
+    if (HystrixCommand.class.isAssignableFrom(method.getReturnType())) {
+      return hystrixCommand;
+    }
+    return hystrixCommand.execute();
+  }
+
+  public static final class Factory implements InvocationHandlerFactory {
```
adriancole: again probably not needed to be public

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -0,0 +1,97 @@
...
+    @Headers("Accept: application/json")
+    HystrixCommand<String> command();
+
+    @RequestLine("GET /")
+    @Headers("Accept: application/json")
+    HystrixCommand<Integer> intCommand();
+
+    @RequestLine("GET /")
+    @Headers("Accept: application/json")
+    String get();
```
adriancole: one more plain `List<String>` just in case :P

README.md
```diff
@@ -152,6 +152,16 @@ MyService api = Feign.builder().client(RibbonClient.create()).target(MyService.c
 
 ```
 
+### Hystrix
+[HystrixFeign](https://github.com/Netflix/feign/tree/master/hystrix) creates a builder that configures circuit breaker support provided by [Hystrix](https://github.com/Netflix/Hystrix).
```
adriancole: s/creates a builder that //

builder isn't the headline

hystrix/README.md
```diff
@@ -0,0 +1,40 @@
...
+// for reactive
+api.getYourType("a").toObservable();
+
+// for asynchronous
+api.getYourType("a").queue();
+
+// for synchronous
+api.getYourType("a").execute();
+
+// or to apply hystrix to existing feign methods.
```
adriancole: nice

adriancole:
:shipit:

PR #287, opened 2015-10-28
Proof of concept for Visitor-pattern-based handling of Responses

uschi2000:
Hi @adriancole ,

This pull-request is a proof-of-concept for making response handling more flexible. (Note that I haven't fixed up the tests, so it doesn't compile.) The motivation is as follows: Our servers often return 404s for non-existing resources and we'd like to surface those as Optional#absent (Guava or Java8) in clients, rather than as exceptions. Given the current architecture of response handling in SynchronousResponseHandler this is not possible since the configured Decoder is only invoked for [200, 300) error codes. This PR shows how to use a Visitor pattern to make response handling more flexible and configurable.

I'd like to know if you'd in principle be happy to consider this or a similar approach, or if you are opposed based on some fundamental consideration. In the former case, I'm happy to get APIs and tests in good shape according to your suggestions. Back-compat of the Feign.Builder class will be an issue.

Cheers!
  Robert

### DIFF
--- a/core/src/main/java/feign/ResponseHandler.java
+++ b/core/src/main/java/feign/ResponseHandler.java
@@ -1,0 +1,26 @@
+package feign;
+
+public interface ResponseHandler {
+
+  /**
+   * If this handler can handle the given {@link Response}, returns the decoded response into {@code result} and
+   * returns true; else returns false. Throws in case of internal errors or in case the given response should be
+   * surfaced as an error, e.g., when the response has error code 404.
+   */
+  boolean handle(Response response, MethodMetadata metadata, ResultHolder result) throws Throwable;
+
+  /**
+   * Mutable Object wrapper for holding/returning the result of a handler; null is a valid result.
+   */
+  class ResultHolder {
+    private Object result = null;
+
+    public Object getResult() {
+      return result;
+    }
+
+    public void setResult(Object result) {
+      this.result = result;
+    }
+  }
+}
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -21,7 +21,6 @@ import java.util.concurrent.TimeUnit;
 
 import feign.InvocationHandlerFactory.MethodHandler;
 import feign.Request.Options;
-import feign.codec.DecodeException;
 import feign.codec.Decoder;
 import feign.codec.ErrorDecoder;
 
@@ -41,13 +40,12 @@ final class SynchronousMethodHandler implements MethodHandler {
   private final Logger.Level logLevel;
   private final RequestTemplate.Factory buildTemplateFromArgs;
   private final Options options;
-  private final Decoder decoder;
-  private final ErrorDecoder errorDecoder;
+  private final Iterable<ResponseHandler> responseHandlers;
   private SynchronousMethodHandler(Target<?> target, Client client, Retryer retryer,
                                    List<RequestInterceptor> requestInterceptors, Logger logger,
                                    Logger.Level logLevel, MethodMetadata metadata,
                                    RequestTemplate.Factory buildTemplateFromArgs, Options options,
-                                   Decoder decoder, ErrorDecoder errorDecoder) {
+                                   Iterable<ResponseHandler> responseHandlers) {
     this.target = checkNotNull(target, "target");
     this.client = checkNotNull(client, "client for %s", target);
     this.retryer = checkNotNull(retryer, "retryer for %s", target);
@@ -58,8 +56,7 @@ final class SynchronousMethodHandler implements MethodHandler {
     this.metadata = checkNotNull(metadata, "metadata for %s", target);
     this.buildTemplateFromArgs = checkNotNull(buildTemplateFromArgs, "metadata for %s", target);
     this.options = checkNotNull(options, "options for %s", target);
-    this.errorDecoder = checkNotNull(errorDecoder, "errorDecoder for %s", target);
-    this.decoder = checkNotNull(decoder, "decoder for %s", target);
+    this.responseHandlers = checkNotNull(responseHandlers, "responseHandlers for %s", target);
   }
 
   @Override
@@ -103,23 +100,16 @@ final class SynchronousMethodHandler implements MethodHandler {
         response =
             logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
       }
-      if (Response.class == metadata.returnType()) {
-        if (response.body() == null) {
-          return response;
+
+      // Return the result produced by the first accepting handler.
+      ResponseHandler.ResultHolder result = new ResponseHandler.ResultHolder();
+      for (ResponseHandler handler : responseHandlers) {
+        if (handler.handle(response, metadata, result)) {
+          return result.getResult();
         }
-        // Ensure the response body is disconnected
-        byte[] bodyData = Util.toByteArray(response.body().asInputStream());
-        return Response.create(response.status(), response.reason(), response.headers(), bodyData);
       }
-      if (response.status() >= 200 && response.status() < 300) {
-        if (void.class == metadata.returnType()) {
-          return null;
-        } else {
-          return decode(response);
-        }
-      } else {
-        throw errorDecoder.decode(metadata.configKey(), response);
-      }
+      throw new FeignException("Failed to handle response, no applicable handler found.");
+
     } catch (IOException e) {
       if (logLevel != Logger.Level.NONE) {
         logger.logIOException(metadata.configKey(), logLevel, e, elapsedTime);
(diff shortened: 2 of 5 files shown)


core/src/main/java/feign/ResponseHandler.java
```diff
@@ -0,0 +1,26 @@
...
+   * If this handler can handle the given {@link Response}, returns the decoded response into {@code result} and
+   * returns true; else returns false. Throws in case of internal errors or in case the given response should be
+   * surfaced as an error, e.g., when the response has error code 404.
+   */
+  boolean handle(Response response, MethodMetadata metadata, ResultHolder result) throws Throwable;
+
+  /**
+   * Mutable Object wrapper for holding/returning the result of a handler; null is a valid result.
+   */
+  class ResultHolder {
```
spencergibb: Maybe add a `boolean handles(Response response, MethodMetadata metadata)` and have `handle` return an Object.

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -103,23 +100,16 @@ Object executeAndDecode(RequestTemplate template) throws Throwable {
...
-        return Response.create(response.status(), response.reason(), response.headers(), bodyData);
-      }
-      if (response.status() >= 200 && response.status() < 300) {
-        if (void.class == metadata.returnType()) {
-          return null;
-        } else {
-          return decode(response);
+
+      // Return the result produced by the first accepting handler.
+      ResponseHandler.ResultHolder result = new ResponseHandler.ResultHolder();
```
spencergibb: Yeah, this makes me like `ResultHolder` even less.

adriancole:
My suggestion would be to have a look at the hystrix support. If we added a plugin arch, it should be able to address existing use cases of extensibility. IOTW, we don't really want to do this twice and I think the hystrix code merged recently probably shows how to achieve your concerns. If it doesn't, it would be good to note the gaps. https://github.com/Netflix/feign/tree/master/hystrix

uschi2000:
Cheers, will have a look.

uschi2000:
@adriancole , I may not understand right what you mean. When I look at the implementation of the Hystrix module, it looks like it's wrapping the reflection proxy in order to wrap the value returned from remote into a new HystrixCommand. This technique doesn't quite work for the my case described above since I need to change the way the MethodHandler itself works; seeing it's return value is not enough. Maybe I'm missing something?

adriancole:
I guess what I was thinking about is that you are looking to supply a fallback value (ex absent on 404). This requires coordination in the same way hystrix does. Hystrix has a richer fallback system, so if anything I would expect more code there.

For example, to acheive the opinion you want, you are teaching feign optional (codec) and also how to interpret exceptions (ex 404 should fallback to absent is possible today with an invocation handler). 

Coordinating multiple aspects is what hystrix is doing as well.

By first using the least code, we can understand how much pain this causes and how this leads to a cure. I'm hesitant to add concepts that don't also increase the clarity of similar code such as hystrix which also handles fallbacks.

Did this explanation help or hurt?

uschi2000:
@adriancole  thanks for your suggestions. An implementation analogous to the Hystrix one looks roughly like this:
````
final class OptionalAwareInvocationHandler implements InvocationHandler {
    private final Map<Method, MethodHandler> dispatch;

    OptionalAwareInvocationHandler(Map<Method, MethodHandler> dispatch) {
        this.dispatch = Util.checkNotNull(dispatch, "dispatch");
    }

    @Override
    public Object invoke(final Object proxy, final Method method, final Object[] args) throws Throwable {
        Type type = method.getGenericReturnType();
        if (type instanceof ParameterizedType && ((ParameterizedType) type).getRawType().equals(Optional.class)) {
            try {
                return Optional.of(dispatch.get(method).invoke(args));
            } catch (NotFoundException e) {
                return Optional.absent();
            }
        } else {
            return dispatch.get(method).invoke(args);
        }
    }

    static final class Factory implements InvocationHandlerFactory {
        @Override
        public InvocationHandler create(Target target, Map<Method, MethodHandler> dispatch) {
            return new OptionalAwareInvocationHandler(dispatch);
        }
    }
}
````

````
public final class OptionalAwareFeign {
    private OptionalAwareFeign() {}

    public static Builder builder() {
        return new Builder();
    }

    public static final class Builder extends Feign.Builder {
        public Builder() {
            invocationHandlerFactory(new OptionalAwareInvocationHandler.Factory());
            contract(new OptionalDelegatingContract(new Contract.Default()));
        }

        @Override
        public Feign.Builder contract(Contract contract) {
            return super.contract(new OptionalDelegatingContract(contract));
        }
    }
}
````

A few things feel a little off about this:
- Extending Feign.Builder is a bit off a hack, both in terms of implementation and usability
- Implementing the Optional functionality at the level of the InvocationHandler is a little cumbersome: I first let "vanilla feign" do its work by delegating, then catch a subset of errors (NotFoundException, in this case) and turn them into Optional#absent. I believe it would be much more transparent if this functionality were implemented at the decoder level.
- IMO error handling (ErrorDecoder.class) and decoding (Decoder.class) are the same abstract concept: You see an HTTP Response and need to surface an appropriate Java "answer". I don't think it's necessary to force users to see exceptions for anything outside [200-299) by delegating to an ErrorDecoder otherwise; this is the correct default, but there is no architectural reasons users should not be allowed to override this.

Bottom-line: I can run with the above work-around for now, but would love to see a more first-class way of allowing users to configure response decoding. Happy to help out with code as well, let me know.

adriancole:
Thanks for the input!

@spencergibb @jdamick any other ideas on this? I've been highly resistant
to introducing a fallback handler in feign as it is a half-fix when
compared to hystrix. We did use this often in jclouds, almost always to
turn 404s in to absent or empty list.

The 404 scenario is common enough that I could see us entertaining an
option to either widen what decoder can interpret to *any* status (which is
imho instant tech debt), or maybe special casing 404 somehow. Ex. an
internal or package visible variable to allow decoder to interpret 404..

uschi2000:
Adrian, do I understand right that feign in your opinion *should* be opinionated about interpreting HTTP codes? I don't (yet) see the downside of providing sane defaults (as per the current behaviour), but allowing users to customize as they see fit.

adriancole:
the harm is that the interface by which we support customization is
something we have to deal with for years. It is good software design to
keep narrowest possible interface to support the highest amount of use
cases. This is what feign tries to do.

Interfaces such as those addressing fallbacks often limit other types of
work. For example, we at one point did have a fallback interface. This was
removed as it wasn't compatible with async invocation.

It is the case that people ask to treat non-success range as success. I
usually offer to them alternatives, such as make a client wrapper that
returns a proper http code, when the api doesn't. Usually, they are very
edge case, and not something most are interested in.

Then, there's 404 -> some empty val. This is something I agree is worth
looking into, but not if it means opening the entire status range. It
simply doesn't require that in order to support it.

When we introduce a new interface, we have to make sure it is worth it. The
clutter of several classes to support coercing 404 -> X is overkill for the
problem.

Libraries who go for widest interface to support narrow decisions often are
just lengthening the rope to hang themselves with. They also lose the
product context, like what this tool even does. Ex. the actual requirement
is addressing 404, not all codes.

it may seem like just adding several types to support 404 isn't a problem
in isolation. It could be the kiss of death, though. Few like to debug,
answer questions, or maintain unnecessarily complex code. It is really
important to not swamp the project.

jdamick:
perhaps the error decoder could give the option to not throw.. it seems like if there was a ThrowingErrorDecoder vs. Non that would more simply solve this problem?

adriancole:
Honestly, if given a choice, I prefer this to adding a new interface or a
special 404 hook. Basically, it means reverting this I think
https://github.com/Netflix/feign/pull/12

adriancole:
problem is that lots of folks have ErrorDecoders already.. and the
fallback capable impl is not compatible with the non.

adriancole:
we're chatting on gitter about this

uschi2000:
Thanks for sharing your philosophy, Adrian. It would be great to see 404->Optional handling and I trust you will find the right way of supporting it in feign. Until then, I will go with the InvocationHandler route.

adriancole:
ok here's the gist. I think we should introduce 404 -> something via a decoder flag.

next in the long answers :)

We have one pattern for integration of systems that have a special return value. `feign-hystrix` is a great example, of this, but it also extends to similar things like `Try<X>`, `Single<X>`, etc. Basically, anything that has custom error handling which needs to be taken into consideration when creating a response.

I don't think 404 falls into this bucket, or needs to. If we had a way to address 404, which is semantically defined as absent, we gain a lot.

Firstly, we have a fast-path for addressing absence which doesn't accidentally also catch bad request, redirect, etc. In other words, it is a very safe policy: it doesn't overlap with retries or other behavior in feign.

Secondly, we don't create a cliff, where folks looking for error handling eventually realize they can't address all fallbacks with only an http response code (bc exceptions can happen far before you get a response from the server). This is a slippery slope that leads to things like Hystrix, so might as well encourage that for *real* errors.

Most importantly, addressing 404 is extremely high value. You are just the last to mention it. For example, I would use this in denominator. Moreover, it doesn't require us to introduce or break an interface. This means all users, such as spring-cloud, can easily opt-into this. It could substantially reduce their code in the same way it can reduce your code, but without new types.

uschi2000:
Great, thanks a lot!

PR #290, opened 2015-10-31
Adds Feign.Builder.decode404() to reduce boilerplate for empty semantics

adriancole:
This adds the `Feign.Builder.decode404()` flag which indicates decoders
should process responses with 404 status. It also changes all
first-party decoders (like gson) to return well-known empty values by
default. Further customization is possible by wrapping or creating a
custom decoder.

#### On why we aren't opening all codes

Prior to this change, we used custom invocation handlers as the way to
add fallback values based on exception or return status. `feign-hystrix`
uses this to return `HystrixCommand<X>`, but the general pattern applies
to anything that has a type representing both success and failure, such
as `Try<X>` or `Observable<X>`.

As we define it here, 404 status is not a retry or fallback policy, it
is just empty semantics. By limiting Feign's special processing to 404,
we gain a lot with very little supporting code.

If instead we opened all codes, Feign could easily turn bad request,
redirect, or server errors silently to null. This sort of configuration 
issue is hard to troubleshoot. 404 -> empty is a very safe policy vs 
all codes.

Moreover, we don't create a cliff, where folks seeking fallback policy 
eventually realize they can't if only given a response code. Fallback
systems like Hystrix address exceptions that occur before or in lieu of
a response. By special-casing 404, we avoid a slippery slope of half-
implementing Hystrix.

Finally, 404 handling has been commonly requested: it has a clear use-
case, and through that value. This design supports that without breaking
compatibility, or impacting existing integrations such as Hystrix or 
Ribbon.

See #238 #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 8.12
+* Adds `Feign.Builder.decode404()` to reduce boilerplate for empty semantics.
+
 ### Version 8.11
 * Adds support for Hystrix via a `HystrixFeign` builder.
 
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -82,8 +82,7 @@ public abstract class Feign {
 
   public static class Builder {
 
-    private final List<RequestInterceptor>
-        requestInterceptors =
+    private final List<RequestInterceptor> requestInterceptors =
         new ArrayList<RequestInterceptor>();
     private Logger.Level logLevel = Logger.Level.NONE;
     private Contract contract = new Contract.Default();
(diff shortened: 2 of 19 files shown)


adriancole:
Thanks to @uschi2000 @nblair, who both asked for this and offered ideas about it, also @spencergibb and @jdamick, who were involved in recent brainstorming

adriancole:
@FrEaKmAn your work is also related.

nblair:
I like this approach, :+1: 

Thanks!

spencergibb:
:+1:

uschi2000:
Wow, that was quick. Cheers Adrian!

adriancole:
quick is relative: I'm sure there are folks who think this took too long
(as asked earlier) :)  I'll take the cheers, regardless!

PR #293, opened 2015-11-02
SynchronousMethodHandler no longer wraps Decoder#decode unchecked exc…

uschi2000:
…eptions in DecodeException

Removing exception wrapping adds flexibility to the interplay of Decoder and ErrorDecoder:
A Decoder can now delegate to an ErrorDecoder and the original ErrorDecoder exception gets
thrown rather than a Feign-specific DecodeException. An example use-case is a Jackson/Gson
implementation of special 404-handling of Optional<Foo> methods: when the Feign client has
decode404() enabled, then the Decoder is in charge of dispatching 404 to Optional#absent
where applicable, or to a delegate ErrorDecoder otherwise; consistent exception handling
requires that the exception produced by the ErrorDecoder does not wrapped.

### DIFF
--- a/core/src/test/java/feign/FeignTest.java
+++ b/core/src/test/java/feign/FeignTest.java
@@ -36,6 +36,7 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import feign.Target.HardCodedTarget;
@@ -388,6 +389,23 @@ public class FeignTest {
   }
 
   @Test
+  public void decoderCanThrowUnwrappedExceptionInDecode404Mode() throws Exception {
+    server.enqueue(new MockResponse().setResponseCode(404));
+    thrown.expect(NoSuchElementException.class);
+
+    TestInterface api = new TestInterfaceBuilder()
+        .decode404()
+        .decoder(new Decoder() {
+          @Override
+          public Object decode(Response response, Type type) throws IOException {
+            assertEquals(404, response.status());
+            throw new NoSuchElementException();
+          }
+        }).target("http://localhost:" + server.getPort());
+    api.post();
+  }
+
+  @Test
   public void okIfEncodeRootCauseHasNoMessage() throws Exception {
     server.enqueue(new MockResponse().setBody("success!"));
     thrown.expect(EncodeException.class);
@@ -591,6 +609,11 @@ public class FeignTest {
 
     TestInterfaceBuilder errorDecoder(ErrorDecoder errorDecoder) {
       delegate.errorDecoder(errorDecoder);
+      return this;
+    }
+
+    TestInterfaceBuilder decode404() {
+      delegate.decode404();
       return this;
     }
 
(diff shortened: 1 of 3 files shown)


uschi2000:
@adriancole

adriancole:
Hmm I didn't realize that the decode method you removed here already attempted to not double-wrap FeignException.

This solution will break folks. I originally understood that you tried to throw a FeignException, but that it became wrapped in a DecodeException.

That seems not the case. It seems that you want to throw arbitrary exceptions when decode404 is flagged. I could buy that if you kept it contained to the branch where that flag is enabled.

Can you revert deleting the decode method and only call decoder.decode directly when the decode404 flag is set? Also, add a test to make sure it works!

Thx

uschi2000:
OK, take another look please, @adriancole

core/src/main/java/feign/Feign.java
```diff
@@ -140,6 +140,10 @@ public Builder decoder(Decoder decoder) {
      * {@link Util#emptyValueOf}. To customize further, wrap an existing
      * {@link #decoder(Decoder) decoder} or make your own.
      *
+     * In order to allow {@link Decoder}s to delegate error handling to {@link ErrorDecoder}s,
```
adriancole: I appreciate the thought, but I don't think this will reduce confusion. There's no introduction of what wrapping exceptions is about.

If anything, we'd add a section in the javadoc of Decoder about exception handing, mentioning that runtime exceptions are wrapped as DecodeExceptions except when decode404 is set. I hate that this is even possible, so really probably just put off about us adding the complexity of allowing this.

adriancole:
Not a fan of this change, but I see what you're trying to do and think it is relatively harmless.

I'd like the documentation around this feature moved from Feign into a subpoint on Decoder, as it only applies to those making custom decoders, and only under the circumstance of splitting hairs on 404.

core/src/test/java/feign/FeignTest.java
```diff
@@ -388,6 +388,23 @@ public Object decode(Response response, Type type) throws IOException {
   }
 
   @Test
+  public void decoderCanThrowUnwrappedExceptionInDecode404Mode() throws Exception {
```
adriancole: let's make this more of a story.

`decoderCanThrowUnwrappedExceptionInDecode404Mode` doesn't tell what we want, it tells us what we don't want.
how about `decode404AllowsCustomExceptions`

Instead of curiously returning "success!" to demonstrate 404, return 404?

Then, in the decode method, assert response.code is 404 (to reinforce your point). Then, throw an IllegalStateException or NoSuchElementException, which is more like what you're looking for.

When a test is written like this, future me's will surely delete it, because it makes no sense to have a feature like this. When a test show's the intent, it is less likely to be deleted or abandoned in a future version. Basically, think of your test as an absentee advocate for your feature!

core/src/main/java/feign/Feign.java
uschi2000: done

core/src/test/java/feign/FeignTest.java
uschi2000: done

adriancole:
:clap: 

Thanks, Robert

adriancole:
8.12.1 is on the way out now

uschi2000:
Cheers Adrian. Responsive as always :)

PR #302, opened 2015-12-15
Allow responses to be streamed.

cakofony:
Never expands >8kb responses into memory

It seems less dangerous to avoid closing Response objects than to explode large responses in memory

### DIFF
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -32,6 +32,8 @@ import static feign.Util.ensureClosed;
 
 final class SynchronousMethodHandler implements MethodHandler {
 
+  private static final long MAX_RESPONSE_BUFFER_SIZE = 8192L;
+
   private final MethodMetadata metadata;
   private final Target<?> target;
   private final Client client;
@@ -101,6 +103,7 @@ final class SynchronousMethodHandler implements MethodHandler {
     }
     long elapsedTime = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
 
+    boolean shouldClose = true;
     try {
       if (logLevel != Logger.Level.NONE) {
         response =
@@ -108,6 +111,11 @@ final class SynchronousMethodHandler implements MethodHandler {
       }
       if (Response.class == metadata.returnType()) {
         if (response.body() == null) {
+          return response;
+        }
+        if (response.body().length() == null ||
+                response.body().length() > MAX_RESPONSE_BUFFER_SIZE) {
+          shouldClose = false;
           return response;
         }
         // Ensure the response body is disconnected
@@ -131,7 +139,9 @@ final class SynchronousMethodHandler implements MethodHandler {
       }
       throw errorReading(request, response, e);
     } finally {
-      ensureClosed(response.body());
+      if (shouldClose) {
+        ensureClosed(response.body());
+      }
     }
   }
 
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -195,7 +195,7 @@ public final class ApacheHttpClient implements Client {
 
       @Override
       public Integer length() {
-        return entity.getContentLength() < 0 ? (int) entity.getContentLength() : null;
+        return entity.getContentLength() >= 0 ? (int) entity.getContentLength() : null;
       }
 
       @Override
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -60,6 +60,7 @@ public class ApacheHttpClientTest {
     assertThat(response.headers())
         .containsEntry("Content-Length", asList("3"))
         .containsEntry("Foo", asList("Bar"));
+    assertThat(response.body().length()).isEqualTo(3);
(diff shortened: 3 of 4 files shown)


core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -101,13 +103,20 @@ Object executeAndDecode(RequestTemplate template) throws Throwable {
...
             logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
       }
       if (Response.class == metadata.returnType()) {
-        if (response.body() == null) {
+        if (response.body() == null || response.body().isRepeatable()) {
+          return response;
+        }
+        if (response.body().length() == null ||
+                response.body().length() > MAX_RESPONSE_BUFFER_SIZE ||
+                response.body().length() < 0) {
```
adriancole: I think we should forbid this being -1, as it is redundant with the semantics of null. ex. enforce in InputStreamBody   ctor as -1 should be a bug. Have you ever seen this assigned to -1 in feign?

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -195,7 +195,7 @@ Response toFeignResponse(HttpResponse httpResponse) throws IOException {
 
       @Override
       public Integer length() {
-        return entity.getContentLength() < 0 ? (int) entity.getContentLength() : null;
+        return entity.getContentLength() >= 0 ? (int) entity.getContentLength() : null;
```
adriancole: can you avoid revert the import change? it drowns out the actual bugfix here

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -33,11 +33,13 @@
 import feign.RequestLine;
 import feign.Response;
 
-import static feign.Util.UTF_8;
-import static feign.assertj.MockWebServerAssertions.assertThat;
 import static java.util.Arrays.asList;
+
 import static org.junit.Assert.assertEquals;
 
+import static feign.Util.UTF_8;
```
adriancole: similarly, please revert the import change so we know what's different (if anything)

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -54,19 +56,44 @@ public void parsesRequestAndResponse() throws IOException, InterruptedException
...
-        .hasPath("/?foo=bar&foo=baz&qux=")
-        .hasHeaders("Foo: Bar", "Foo: Baz", "Qux: ", "Accept: */*", "Content-Length: 3")
-        .hasBody("foo");
+    TestInterface api = Feign.builder()
+            .client(new ApacheHttpClient())
+            .target(TestInterface.class, "http://localhost:" + server.getPort());
+
+    Response response = api.post("testing");
+    try {
+      assertThat(response.status()).isEqualTo(200);
```
adriancole: too bad we don't have a feign Response assertj class, right? this would look much nicer chained. (ps if you feel like it, make one, else no biggie)

core/src/main/java/feign/SynchronousMethodHandler.java
cakofony: Good call, I have not seen a negative value there.
I also think that checking isRepeatable() is wrong in the above condition, it may have to get resolved before being closed.  I'll remove that part of the conditional.

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -54,19 +56,44 @@ public void parsesRequestAndResponse() throws IOException, InterruptedException
...
+      assertThat(response.reason()).isEqualTo("OK");
+      assertThat(response.body().length()).isNull();
+      assertThat(response.body().asInputStream())
+              .hasContentEqualTo(new ByteArrayInputStream("foo".getBytes(UTF_8)));
+    } finally {
+      response.close();
```
adriancole: I suspect you don't need to do this, as the mockwebserver rule will implicitly close this, right?

Just the indent + all the assertThat things makes this test hard to read.

https://github.com/square/okhttp/blob/master/mockwebserver/src/main/java/com/squareup/okhttp/mockwebserver/MockWebServer.java#L151

adriancole:
looking good. let me know when you are ready for realz

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
cakofony: fixed

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
cakofony: That's true, once the InputStream is completely read (via hasContentEqualTo) the connection will be released.

Fixed :-)

cakofony:
Cleaned up the tests, I think we're good to go :-)
Thanks!

adriancole:
oops one last thing add a note in the CHANGELOG for 8.13

Thanks tons for making a compatible way to address this feature!

adriancole:
nm I'll work this out on the way in

cakofony:
Sorry about that, fell asleep. Thanks for reviewing/merging so quickly!
Any idea when to expect 8.13 to be published to maven?

adriancole:
sorry this never made it out. I've now released it in 8.16.2

PR #305, opened 2015-12-23
Document and expose HystrixDelegatingContract

adriancole:
HystrixDelegatingContract is reusable when developers like @marcingrzejszczak make custom invocation handlers.


spencergibb:
:+1:

adriancole:
on the way out as 8.13.1

Issue #298, opened 2015-12-01
Configuring a fallback in feign clients


codependent:
It would be great if we could configure a Hystrix fallback in Feign clients. Any plans to implement this funcionality?

adriancole:
can you explain how this request is different than feign-hystrix? (probably
with code is best). Also there have been many requests around hystrix, so
if you do an issue search and find something similar, this will help too!

codependent:
Hi! I searched the repository for something related and didn't find anything.

This is related to an [issue](https://github.com/spring-cloud/spring-cloud-netflix/issues/680) I opened in Spring Cloud Netflix repo: I am using their Feign integration throw the use of `@FeingClient` annotation:
```
@FeignClient("stories-microservice")
public interface StoryClient {

    @RequestMapping(method = RequestMethod.GET, value = "/stories", params="random")
    String getStory(@RequestParam("random") boolean random);
```
I have a rest controller that calls this client:
```
@RestController
@RequestMapping("/api")
@EnableEurekaClient
@EnableFeignClients
@EnableCircuitBreaker
@RefreshScope
@SpringBootApplication
public class StoryTellerApiApplication {

    @Autowired
    private StoryClient sgc;

    @RequestMapping(value="/stories", params="random=true")
    public String getRandomStory(HttpServletResponse response) {
        String randomStory = sgc.getStory(true);
```
 The point is that there is no way of configuring a Hystrix callback for Feign clients. Under the hood feign-hystrix is used:

`org.springframework.cloud.netflix.feign.FeignClientFactoryBean`;

```
    Feign.Builder builder = HystrixFeign.builder()
	// required values
	.logger(logger)
	.encoder(get(factory, Encoder.class))
	.decoder(get(factory, Decoder.class))
	.contract(get(factory, Contract.class));
```

 I have been looking for a way of defining hystrix callback methods in the native HystrixFeignBuilder but it seems there isn't any way to do it.

spencergibb:
@adriancole this if for defining a Hystrix fallback when there are errors. It was intentionally left out because I didn't know how to define a fallback and it wasn't needed to get simple hystrix support working.

dnathanson:
I was wondering about this, too.  Maybe something with an @Fallback annotation that can be added to methods that points to static methods on an extension of the client interface. 

```java
public interface FooClient {
      void doSomething(String withArg);
}
 
@FeignClient       
public interface FeignFooClient extends FooClient {

    @Override
[...]

PR #307, opened 2016-01-02
Add support for Observable return type to Hystrix module

christopherlakey:
Support Observable return types by automatically calling `toObservable()`.  This makes it possible to declare a pure RxJava Feign API that is portable between the current Hystrix module and a future RxJava module.

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -28,6 +28,8 @@ import com.netflix.hystrix.HystrixCommandKey;
 import feign.InvocationHandlerFactory;
 import feign.InvocationHandlerFactory.MethodHandler;
 import feign.Target;
+import rx.Observable;
+import rx.Single;
 
 final class HystrixInvocationHandler implements InvocationHandler {
 
@@ -62,6 +64,12 @@ final class HystrixInvocationHandler implements InvocationHandler {
 
     if (HystrixCommand.class.isAssignableFrom(method.getReturnType())) {
       return hystrixCommand;
+    } else if (Observable.class.isAssignableFrom(method.getReturnType())) {
+      // Create a cold Observable
+      return hystrixCommand.toObservable();
+    } else if (Single.class.isAssignableFrom(method.getReturnType())) {
+      // Create a cold Observable as a Single
+      return hystrixCommand.toObservable().toSingle();
     }
     return hystrixCommand.execute();
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 8.14
+* Add support for RxJava Observable and Single return types via the `HystrixFeign` builder.
+
 ### Version 8.13
 * Never expands >8kb responses into memory
 * Bumps dependency versions, most notably Gson 2.5 and OkHttp 2.7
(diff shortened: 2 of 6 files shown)


adriancole:
Good idea. I'd update the PR description with your rationale (which is to sketch a portability path between this and a future rxjava direct integration).

Question about this though.. Should we also do Single, as that's more semantical for rpc in some folks minds?

Lastly, probably worth updating CHANGELOG and the javadoc somewhere

christopherlakey:
Good idea.  I just updated the README.md and I'll update the javadocs for `HystrixDelegatingContract` and `HystrixFeign`.

For the CHANGELOG, should I assume version 8.14.0?

christopherlakey:
@adriancole `rx.Single` looks like a good fit for Feign.  I'll take a look at Single and can add support for it as well.

I think Observable should also be supported since Single is still marked as beta and Observable is likely more interoperable for those already using `toObservable()`.

spencergibb:
Seems like a good idea.

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -62,6 +63,8 @@ protected Object run() throws Exception {
 
     if (HystrixCommand.class.isAssignableFrom(method.getReturnType())) {
       return hystrixCommand;
+    } else if (Observable.class.isAssignableFrom(method.getReturnType())) {
+      return hystrixCommand.toObservable();
```
spencergibb: There is also `observe()` what has different semantics from `toObservable()`. If  `toObservable()` is what you want, you'll need to document it.

adriancole: Good idea!

christopherlakey: Good point. I returned a "cold" Observable to allow the consumer more control over execution and it definitely needs to be documented.

Is there a case to make this configurable or return a hot observable instead? One advantage of a "hot" observable is that you don't need to subscribe if you're not interested in the result. This might be the case when performance an update style request.

christopherlakey: Something to consider is portability with a RxJava implementation based on something like RxNetty.  As best I can tell, the Observables in RxNetty are "cold".  

```The request is executed every time {@link HttpClientRequest}, or {@link Observable} returned by {@code write*Content} is subscribed and is the only way of executing the request.```

My knowledge of Hystrix is limited.  Does the cold Observable (`toObervable()`) execute the command each time you subscribe, or does it use a replay, similar to `observe()`?

Comment (adriancole) [inline hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java]: Great idea looking at RxNetty.  To avoid a 3rd compat concern, I'd probably
defer to cold unless @NiteshKant has cause to believe RxNetty will warm up
in the future.

**Comment by adriancole** [comment]
one last suggestion about doc/testing the coldness. after that, squash for merge!

**Comment by christopherlakey** [comment]
Asserts have been added to prove that the returned observables are indeed cold:

```
    Observable<List<String>> observable = api.listObservable();
    assertThat(observable).isNotNull();
    assertThat(server.getRequestCount()).isEqualTo(0);
```

The README has been updated and commits have also been squashed.

adriancole:
made my day. Thanks so much for the great work, here!

christopherlakey:
Thanks for Feign.  It's a great API and has saved me a lot of work.

Any idea when 8.14 will be released?

adriancole:
I'll take a quick stab at #298 and cut after that, so likely today (weds)

adriancole:
after this.. https://github.com/Netflix/feign/pull/308

PR #308, opened 2016-01-06
Adds fallback implementation configuration to the HystrixFeign builder

### LINKED ISSUE #298: Configuring a fallback in feign clients

adriancole:
Fallbacks are known values, which you return when there's an error invoking an http method.
For example, you can return a cached result as opposed to raising an error to the caller. To use
this feature, pass a safe implementation of your target interface as the last parameter to `HystrixFeign.Builder.target`.

Here's an example:

```java
// When dealing with fallbacks, it is less tedious to keep interfaces small.
interface GitHub {
  List<String> contributors(@Param("owner") String owner, @Param("repo") String repo);
}

// This instance will be invoked if there are errors of any kind.
GitHub fallback = (owner, repo) -> {
  if (owner.equals("Netflix") && repo.equals("feign")) {
    return Arrays.asList("stuarthendren"); // inspired this approach!
  } else {
  }
};

GitHub github = HystrixFeign.builder()
                            ...
                            .target(GitHub.class, "https://api.github.com", fallback);
```

Credit to the idea goes to @stuarthendren!

Fixes #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Version 8.14
 * Add support for RxJava Observable and Single return types via the `HystrixFeign` builder.
+* Adds fallback implementation configuration to the `HystrixFeign` builder
 
 ### Version 8.13
 * Never expands >8kb responses into memory
--- a/hystrix/README.md
+++ b/hystrix/README.md
@@ -48,4 +48,33 @@ api.getYourType("a").execute();
 
 // or to apply hystrix to existing feign methods.
 api.getYourTypeSynchronous("a");
-``` 
+```
+
+### Fallback support
+
+Fallbacks are known values, which you return when there's an error invoking an http method.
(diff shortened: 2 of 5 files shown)


adriancole:
ps @spencergibb at the level of abstraction of feign, it doesn't care how the fallback instance is created. so, it could be created by a dynamic proxy or whatever (which I suspect would be the way to integrate with custom annotation processing)

adriancole:
will merge this in 8hrs unless there's blocking feedback, as there are folks waiting on 8.14

adriancole:
scrubbed covariants after seeing this! https://gist.github.com/JakeWharton/14547311b1f145cb0695

adriancole:
this is now released as 8.14 cc @christopherlakey

PR #313, opened 2016-01-13
Allow HystrixFeign.Builder to be used incrementally.

spencergibb:


hystrix/README.md
```diff
@@ -74,7 +74,7 @@ GitHub fallback = (owner, repo) -> {
   }
 };
 
-GitHub github = HystrixFeign.builder()
-                            ...
-                            .target(GitHub.class, "https://api.github.com", fallback);
+HystrixFeign.Builder builder = HystrixFeign.builder();
+... 
+GitHub github = builder.target(GitHub.class, "https://api.github.com", fallback);
```
spencergibb: Because the change is in a terminal method you need it's original type :disappointed:. One option is to move create a `fallback()` method which could retain the fluent style. Another is to create an `AbstractBuilder<B>` that returns `B` from each method allowing downstream builders to have custom terminal methods.

hystrix/src/main/java/feign/hystrix/HystrixFeign.java
```diff
@@ -79,137 +76,26 @@ public InvocationHandler create(Target target, Map<Method, MethodHandler> dispat
...
-    }
-
-    /**
-     * @see feign.Feign.Builder#client
-     */
-    public Builder client(Client client) {
-      delegate.client(client);
+    @Override
+    public Feign.Builder invocationHandlerFactory(InvocationHandlerFactory invocationHandlerFactory) {
+      //TODO: throw exception, log warning or do nothing?
```
spencergibb: Since the method can no longer be hidden, what do you think should be done here?

adriancole: unsupportedoperationexception

hystrix/src/main/java/feign/hystrix/HystrixFeign.java
```diff
@@ -79,137 +76,26 @@ public InvocationHandler create(Target target, Map<Method, MethodHandler> dispat
...
       return this;
     }
 
     /**
-     * @see feign.Feign.Builder#retryer
-     */
-    public Builder retryer(Retryer retryer) {
```
adriancole: I'd switch all these to just covariant overrides, as that allows you to chain to the new methods. I had done this in the first place, ironically enough.

ex. kill the javadocs as they are implicitly the same. add 

```
@Override
public Builder logger(Logger logger) {
  return (Builder) super.logger(logger);
} 
```

Rationale is I don't want to add anything new to the base builder (like a paramterized interface) as in the future, as I think extending builders digs holes like these. Eventhough covariant overrides are more complex bytecode, there's not many options to work with.

adriancole:
I had this in my intellij history, so yeah :P

adriancole:
ps the feeling against a base `builder<X extends builder>` is more-or-less that this became unwieldy in other projects (like jclouds). For example a type doing a static method now has an awkward `Builder<?>` which makes overrides even weirder to resolve. I don't like covariance either, just it contains the complexity as an internal detail which users won't see.

PR #323, opened 2016-01-27
Handle invalid HTTP received from server

carlossg:
If response was empty it caused a NPE

```
java.lang.NullPointerException: reason
  at feign.Util.checkNotNull(Util.java:104)
  at feign.Response.<init>(Response.java:49)
  at feign.Response.create(Response.java:60)
  at feign.Client$Default.convertResponse(Client.java:171)
  at feign.Client$Default.execute(Client.java:72)
  at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:95)
  at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:74)
  at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:94)
```

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -15,6 +15,8 @@
  */
 package feign;
 
+import static java.lang.String.format;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -149,6 +151,12 @@ public interface Client {
     Response convertResponse(HttpURLConnection connection) throws IOException {
       int status = connection.getResponseCode();
       String reason = connection.getResponseMessage();
+
+      if (status < 0 || reason == null) {
+        // invalid response
+        throw new IOException(format("Invalid HTTP executing %s %s", connection.getRequestMethod(),
+            connection.getURL()));
+      }
 
       Map<String, Collection<String>> headers = new LinkedHashMap<String, Collection<String>>();
       for (Map.Entry<String, List<String>> field : connection.getHeaderFields().entrySet()) {


adriancole:
thx for the PR.. so in this case, we are wrong, I think. It is valid (albeit super weird) to have an empty reason. It smells of error, though! http://tools.ietf.org/html/rfc7230#section-3.1.2

This is hard to test as I'm not sure we could get MockWebServer to return a null reason either..

Regardless, I might throw an IOException instead, as we are assuming there is a transport problem.

core/src/main/java/feign/Client.java
```diff
@@ -150,6 +157,11 @@ Response convertResponse(HttpURLConnection connection) throws IOException {
       int status = connection.getResponseCode();
       String reason = connection.getResponseMessage();
 
+      if (status < 0 || reason == null) {
+        // invalid response
+        return null;
```
adriancole: so basically, just throw an IOE here.

core/src/main/java/feign/Client.java
```diff
@@ -15,6 +15,8 @@
  */
 package feign;
 
+import static java.lang.String.*;
```
adriancole: nit: explicitly import format, as we generally don't use wildcards.

carlossg:
In my case the server is returning nothing through the connection due to server error. I haven't checked but I guess java.net is also returning status code -1 per the javadocs

> Returns null if none could be discerned from the responses (the result was not valid HTTP).

https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#getResponseMessage()

> Returns -1 if no code can be discerned from the response (i.e., the response is not valid HTTP).

https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#getResponseCode()

Any exception is fine

carlossg:
updated and rebased

PR #324, opened 2016-02-03
Apache HTTP client hates to have the Content-Length set

dsyer:
It always sets the header itself and if we set it as well it barfs
so it is better just to avoid setting it at all.

### DIFF
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -118,10 +118,9 @@ public final class ApacheHttpClient implements Client {
         hasAcceptHeader = true;
       }
 
-      if (headerName.equalsIgnoreCase(Util.CONTENT_LENGTH) &&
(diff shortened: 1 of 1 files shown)


dsyer:
Example:

```
Caused by: org.apache.http.ProtocolException: Content-Length header already present
	at org.apache.http.protocol.RequestContent.process(RequestContent.java:96)
	at org.apache.http.protocol.ImmutableHttpProcessor.process(ImmutableHttpProcessor.java:132)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:182)
	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:88)
	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184)
	... 118 more
```

adriancole:
thx! lemme know how urgent this is to release

dsyer:
Would be good to have it this week for Spring Cloud Netflix milestone. But if not we always have the other clients.

PR #335, opened 2016-02-25
Add QueryMap annotation

nmiyake:


core/src/main/java/feign/MethodMetadata.java
```diff
@@ -109,4 +112,8 @@ public RequestTemplate template() {
   public Map<Integer, Class<? extends Expander>> indexToExpanderClass() {
     return indexToExpanderClass;
   }
+
+  public Set<Integer> queryMapIndices() {
```
adriancole: I'd simplify this by supporting only one `queryMapIndex`, similar to how we only support a single body or url parameter.

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -196,7 +198,37 @@ public RequestTemplate create(Object[] argv) {
           }
         }
       }
-      return resolve(argv, mutable, varBuilder);
+
+      RequestTemplate template = resolve(argv, mutable, varBuilder);
+
+      // add query map parameters after initial resolve so that they take
```
adriancole: good comment

core/src/main/java/feign/QueryMap.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Retention;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * A template parameter applied to {@linkplain RequestLine}.
+ * The parameter should refer to a Map that contains query parameters,
+ * where the keys are the parameter names and the values are the parameter
+ * values.
```
adriancole: add an example?
```
<pre>{@code
// real-world, but simplified example here.
}</pre>
```

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -196,7 +198,37 @@ public RequestTemplate create(Object[] argv) {
...
+      // precedence over any predefined values
+      return addQueryMapQueryParameters(argv, template);
```
adriancole: not sure if this should be last method in resolve or not.. looks a bit random here.

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -196,7 +198,37 @@ public RequestTemplate create(Object[] argv) {
...
+    }
+
+    private RequestTemplate addQueryMapQueryParameters(Object[] argv, RequestTemplate mutable) {
+    	for (int currQueryMapIndex : metadata.queryMapIndices()) {
```
adriancole: single index will allow you to early exit.. ex. `if (metadata.queryMapIndex == null) return`

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -196,7 +198,37 @@ public RequestTemplate create(Object[] argv) {
...
+      	  Object value = argv[currQueryMapIndex];
```
adriancole: when you are ready, use formatting from other code..

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -196,7 +198,37 @@ public RequestTemplate create(Object[] argv) {
...
+      	  if (value instanceof Map) {
```
adriancole: break the contract if the parameter is not a Map.. then you don't have to check it here

core/src/test/java/feign/DefaultContractTest.java
```diff
@@ -15,11 +15,10 @@
  */
 package feign;
 
-import com.google.gson.reflect.TypeToken;
-
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
+import static feign.assertj.FeignAssertions.assertThat;
```
adriancole: when readying for prime time, revert any accidental import changes

adriancole:
good work! key changes are to only support a single parameter, and also enforce the parameter is indeed a Map (via the contract) else is just polish.

nmiyake:
Thanks for the feedback -- pushed another commit to address the issues you flagged.

adriancole:
Great work. Add a single-sentence description in the CHANGES file for the
next unreleased minor. Also, consider if README would be better mentioning
this in a short note.

nmiyake:
Updated CHANGES and README, squashed commits and rebased.

nmiyake:
Not sure what you mean -- there's a short section called "Dynamic Query Parameters" in the README.md, and the Javadoc in https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/QueryMap.java#L26 covers most of the details. There is also test usage at https://github.com/OpenFeign/feign/blob/deda96a4c5c5966d67d0e17d9d8c460faf67dbe6/core/src/test/java/feign/FeignTest.java#L291.

You can certainly submit a PR to add more documentation as you deem appropriate, but I think that this feature is fairly well-documented as far as Feign features go.

anirudh83:
Thanks a lot for the links, I could not find those, so I thought may be I can help, anyways, thanks for the links, cheers.

anirudh83:
Hey Thanks for the links, I could not find those, that helps.

cheers,
Anirudh

PR #337, opened 2016-02-26
Allow the injection of a custom InvocationHandlerFactory

cchacin:


cchacin:
@adriancole I'm not sure if those changes are in line with the goals, but I think this is a way to use a custom HystrixCommand to wrap the calls.

adriancole:
Not really in favor of this. It is less dangerous to copy/paste and make
your own HystrixFeign. The invocation handler is coupled to how hystrix is
used internally.

By exposing something that was intentionally marked unsupported, we would
now have bugs related to this to add to the issues list.

What are you trying to accomplish?

cchacin:
@adriancole I have a couple of configured HystrixCommand and I want to use it with feign instead of use the HystrixInvocationHandler.

Do you think it's better copy/paste?

cchacin:
What about to receive in the ctor a HystrixCommand and pass it to the HystrixInvocationHandler?, and by default use the current behavior

adriancole:
Yeah let's focus on the feature! Invocation handler isnt exactly a scalpel
:)

Sadly I am mostly away this weekend, but rephrasing as an issue (in this
case customizing the HystrixCommand) would be a good start.

Ex is this for all feign commands or some? What does your invocation
handler look like?

In your local codebase, copy/paste fork etc until there's resolution on it.

adriancole:
so just looking at the code here.. most of what HystrixFeign does is set HystrixDelegatingContract and also HystrixInvocationHandler. Most of the code is in HystrixInvocationHandler, and that's what it seems you'd like to affect.

What if you just used normal feign contract and set HystrixDelegatingContract and custom invocation hander there? Would that work?

Alternatively, what is it about HystrixCommand that you are changing? I know some folks want to change the naming policy, for example. Is there something reusable in what you're doing that others might need?

adriancole:
@nmiyake fyi: I'll wait until weds or when this completes for the next release, whichever is sooner

cchacin:
@adriancole 
> What if you just used normal feign contract and set HystrixDelegatingContract and custom invocation hander there? Would that work?

I'll try that and I'll let you know if that works.

> Alternatively, what is it about HystrixCommand that you are changing? I know some folks want to change the naming policy, for example. Is there something reusable in what you're doing that others might need?

I wanted to create my own HystrixCommand and/or be able to configure roperties like:

```yaml
groupKey: "something"
key: "somethingElse"
requestVolumeThreshold: 1
enabled: true
errorThresholdPercentage: 10
sleepWindowMilliseconds: 10000
timeoutMilliseconds: 100
```

adriancole:
HystrixFeign is a work in progress, but most of the code is in the invocation handler. Changing the invocation handler to a new one is very much like having a new module, as the hystrix invocation handler is most of the lines in HystrixFeign.

I recommend using the normal Feign builder to do what you want, particularly while the current implementation evolves.

Ex.
```java
Feign.builder()
         .contract(new HystrixDelegatingContract())
         .invocationHandlerFactory(yourCustomOne)
         ...
````

Please report back if the above doesn't work.

Also, see #342, soon to be [HACKING](./HACKING.md) which explains a bit the rationale around change cc @marcingrzejszczak who also asked about swapping the invocation handler. Generally speaking, things not designed for extensibility aren't likely to advertise extensibility.

PR #338, opened 2016-02-27
refactor param expander instantiation

mbenson:
As discussed on feign's Gitter chat, a change like this would allow custom Feign `Contract` implementations to use alternate strategies for supplying `Param.Expander` instances.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -24,6 +24,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import feign.Param.Expander;
+
+import static feign.Util.checkNotNull;
 import static feign.Util.checkState;
 import static feign.Util.emptyToNull;
 
@@ -312,4 +315,27 @@ public interface Contract {
       return result;
     }
   }
+
+  class DelegatingContract implements Contract, Param.Expander.Factory {
+    protected final Contract delegate;
+
+    public DelegatingContract(Contract delegate) {
+      super();
+      this.delegate = checkNotNull(delegate, "delegate");
+    }
+
+    @Override
+    public List<MethodMetadata> parseAndValidatateMetadata(Class<?> targetType) {
+      return this.delegate.parseAndValidatateMetadata(targetType);
+    }
+
+    @Override
+    public <E extends Expander> E getInstance(Class<E> expanderType) {
+      if (delegate instanceof Param.Expander.Factory) {
+          return ((Param.Expander.Factory) delegate).getInstance(expanderType);
+      }
+      return Param.DefaultExpanderFactory.INSTANCE.getInstance(expanderType);
+    }
+  }
+
 }
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -28,6 +28,7 @@ import java.util.Map.Entry;
 
 import feign.InvocationHandlerFactory.MethodHandler;
 import feign.Param.Expander;
+import feign.Param.Expander.Factory;
 import feign.Request.Options;
 import feign.codec.Decoder;
 import feign.codec.EncodeException;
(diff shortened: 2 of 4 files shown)


core/src/main/java/feign/Contract.java
```diff
@@ -312,4 +315,27 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
       return result;
     }
   }
+
+  class DelegatingContract implements Contract, Param.Expander.Factory {
```
spencergibb: Seems a little strange to me. Why can't Contact have an expander factory with accessors? Composition over inheritance.

mbenson: Hi @spencergibb . My thought was that since Feign is Java 6-based I didn't want to change the `Contract` interface and break downstream subclasses.

adriancole:
I thought I noted this in one of the conversations some place, but.. here goes just in case:

I'd like to contain the hackery to the contract, and remove any runtime failures by placing them up front. What that means is a volatile map of param to expander instance, which the contract is required to hydrate.

Ex. in MethodMetadata
```java
  private volatile Map<Integer, Expander> indexToExpander = new LinkedHashMap<Integer, Expander>();

  /** After {@link #indexToExpanderClass} is populated, this is called for each entry. */
  public MethodMetadata putExpander(int index, Expander expander) {
    indexToExpander.put(index, expander);
    return this;
  }

  public Map<Integer, Expander> indexToExpander() {
    return indexToExpander;
  }
```

A contract would perform above however it chooses. Ex a spring-powered Contract wouldn't use newInstance, but we also wouldn't add an Injector type to those who do use newInstance.

Make sense?

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -160,21 +167,16 @@ public String toString() {
...
     private final Map<Integer, Expander> indexToExpander = new LinkedHashMap<Integer, Expander>();
 
-    private BuildTemplateByResolvingArgs(MethodMetadata metadata) {
+    private BuildTemplateByResolvingArgs(MethodMetadata metadata, Factory paramExpanderFactory) {
       this.metadata = metadata;
+      checkNotNull(paramExpanderFactory, "paramExpanderFactory");
       if (metadata.indexToExpanderClass().isEmpty()) {
         return;
       }
       for (Entry<Integer, Class<? extends Expander>> indexToExpanderClass : metadata
```
adriancole: ex move the instantiation logic to DefaultContract, and just assign `this.indexToExpander = md.indexToExpander`
A spring contract wouldn't use newInstance, but we don't need a factory type either.

mbenson: Hi @adriancole ... Indeed I did misunderstand you in gitter. You suggest moving the instantiation logic to `Contract.Default`, but would it be better to add this to `BaseContract`? In any case, this still doesn't maintain compatibility for any `Contract` implementations, if any, that exist in the wild and do not inherit from `BaseContract`. For this reason it might still be a good idea to, someplace, maintain a backup of the original `#newInstance()` invocation against the old `indexToxpanderClass` mapping. Any thoughts on these ideas?

adriancole: great point on back compat..

In this case, I'd make this simpler.. make MethodMetadata.indexToExpander
nullable. Document it as if unset, this will be instantiated via
newInstance.

If MethodMetadata.indexToExpander is set (by a Contract) by the time
ReflectiveFeign is trying to assign indexToExpander, use it. Otherwise,
default to the existing logic.

Make sense?

mbenson: That's pretty simple and probably effective. For the sake of discussion, it would be possible to support `indexToExpander` and `indexToExpanderClass` simultaneously, preferring the former, but in practice deprecating `indexToExpanderClass` should mean that any `Contract` implementation using the new form would use it exclusively. The only exception to that might be if there is some composite `Contract` out there, but such an object could just as easily handle the merging of `expanderClass`es with `expander`s. Unless I hear differently I will implement per your thoughts above... however, until you tell me otherwise I am going to approach as `indexToExpander` being empty rather than `null` because that feels to me more  in keeping with the feel of the `MethodMetadata` API (no mutator method for `indexToExpanderClass`). Unless you want to do something wild like return `null` from `#indexToExpander()` if `#indexToExpanderClass()` is not empty, but that starts to suggest other things that become more and more complicated.

adriancole: I think the null signal here is different than empty. Null means
indexToExpander hasn't been processed. If empty, it has been processed, and
happens to be empty. Sorry I didn't make this clear earlier.

I wouldn't deprecate indexToExpanderClass, as the transient hint on
indexToExpander suggests something important.

All non-transient fields on MethodMetadata are relatively simple, they
don't hold values that might be hard to serialize. indexToExpander will,
which is why it is transient.

adriancole:
I was holding up a release on this feature. Apologies for clashing with any in-flight work. I've obviated it here https://github.com/Netflix/feign/pull/343

mbenson:
Np, I hadn't had time to redo it yet. Thanks for getting the support in
place!

PR #342, opened 2016-03-05
Adds HACKING file, to demystify change to Feign

adriancole:
Over the years, there's been a number of times where I've repeated rationale around change on a case-by-case basis. This is a summary of thoughts that I think those who want to change Feign dramatically should bear in mind.


spencergibb:
:+1:

PR #343, opened 2016-03-05
Supports runtime injection of Param.Expander

adriancole:
This supports runtime injection of Param.Expander. Implementing
contracts will assign `MethodMetadata.indexToExpander` using configured
values. When `MethodMetadata.indexToExpander` is unset, Feign has the
existing behavior, which is to newInstance each `indexToExpanderClass`.

Fixes #338


spencergibb:
:+1:

PR #346, opened 2016-03-08
[#345] Hacky solution to wrong headers resolution

### LINKED ISSUE #345: Custom content-types are broken
Hi!

This commit: https://github.com/Netflix/feign/commit/7e25bbbfb269dcde20bc5107b682fc98a4c6812d#diff-31aaa08ee890c87021503798d38d031fR218

Has broken compatibility of vendor content-types. A content-type like this: `application/vnd.io.spring.cloud.test.v1+json` gets resolved to `application/vnd.io.spring.cloud.test.v1 json` (`+` -> ` `).

Check https://github.com/spring-cloud/spring-cloud-netflix/pull/884 for a sample.

I see the point in the release notes. If I register Feign request mapping like this:

```
@RequestMapping(method = RequestMethod.POST,
				consumes = "application/vnd.io.spring.cloud.test.v1%2Bjson",
				produces = "application/vnd.io.spring.cloud.test.v1%2Bjson",
				value = "/complex")
		String moreComplexContentType(String body);
```

the tests pass. However I think it's completely unintutive and bad practice. I have to have two sets of strings, one for normal mapping and one for Feign. Also it's a breaking change that enforces all users to encode their headers which I don't thing should take place when 8.14.4 -> 8.15.0 change takes place. I could expect it to happen in 8.14.4 -> 9.0.0 change.

@ahus1 - would you mind having a look at this? thanks :)

I've created a pathetic quick hacky fix PR - https://github.com/Netflix/feign/pull/346/files ;) You can take just the test from it.

marcingrzejszczak:
fixes #345

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -215,7 +215,8 @@ public final class RequestTemplate implements Serializable {
     for (String field : headers.keySet()) {
       Collection<String> resolvedValues = new ArrayList<String>();
       for (String value : valuesOrEmpty(headers, field)) {
-        String resolved = urlDecode(expand(value, encoded));
+        String expandedValueWithEncodedPlus = expand(value.replace("+", "%2B"), encoded); // special case + as that's commonly used in vnd media types.
+        String resolved = urlDecode(expandedValueWithEncodedPlus);
         resolvedValues.add(resolved);
       }
       resolvedHeaders.put(field, resolvedValues);
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -162,15 +162,15 @@ public class ApacheHttpClientTest {
 
       api.noPostBody();
   }
-  
+
   @Test
   public void noResponseBodyForPut() {
       server.enqueue(new MockResponse());
-      
+
       TestInterface api = Feign.builder()
               .client(new ApacheHttpClient())
               .target(TestInterface.class, "http://localhost:" + server.getPort());
-      
+
       api.noPutBody();
   }
     @Test
@@ -186,6 +186,20 @@ public class ApacheHttpClientTest {
         assertThat(server.takeRequest()).hasMethod("POST")
                 .hasPath("/path/current%20documents/resource")
                 .hasBody("foo");
+    }
+
+    @Test
+    public void getWithVndHeader() throws IOException, InterruptedException {
+        server.enqueue(new MockResponse().setBody("foo"));
+
+        TestInterface api = Feign.builder()
+                .client(new ApacheHttpClient())
+                .target(TestInterface.class, "http://localhost:" + server.getPort());
+
+        String response = api.vnd();
+
+        assertThat(server.takeRequest()).hasMethod("GET")
+                .hasHeaders("Accept: application/vnd.github.v3+json");
     }
 
   interface TestInterface {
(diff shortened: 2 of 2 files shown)


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -215,7 +215,8 @@ public RequestTemplate resolve(Map<String, ?> unencoded) {
     for (String field : headers.keySet()) {
       Collection<String> resolvedValues = new ArrayList<String>();
       for (String value : valuesOrEmpty(headers, field)) {
-        String resolved = urlDecode(expand(value, encoded));
+        String expandedValueWithEncodedPlus = expand(value, encoded).replaceAll("\\+", "__PLUS__");
```
adriancole: does this do the same thing?
```bash
String resolved = urlDecode(expand(value, encoded));
resolved = resolved.replace("%2B", "+"); // special case + as that's commonly used in vnd media types.
```

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -188,6 +188,20 @@ public void postWithSpacesInPath() throws IOException, InterruptedException {
                 .hasBody("foo");
     }
 
+    @Test
```
adriancole: probably you want a test closer to the code, too.. ex one in RequestTemplateTest

adriancole:
@ahus1 I think some version of this hack is good enough (I've seen similar things in the past for %20). wdyt?

core/src/main/java/feign/RequestTemplate.java
marcingrzejszczak: Actually it doesn't. decode just changes `+` into ` ` (space). I thought that it will be exactly what you wrote.

adriancole: you mind reader, you :)

nit: I think "replace" doesn't do regex, so you'd not have to escape the "+" iirc. Otherwise, add a comment like " // special case + as that's commonly used in vnd media types." as that helps set hack-context.

ahus1:
@adriancole @marcingrzejszczak I am fine passing a plus sign "as is". 

But instead of doing two replace operations, I would have done one replace operation on the original string of "+" to "%2B".

Otherwise we would get into trouble if someone uses "__PLUS__" inside a header.

marcingrzejszczak:
hi @ahus1 . I tried that - if I change + to %2B then it gets converted to a space anyways...

marcingrzejszczak:
Argh... instead of `%2B` I used `%20`... your suggestion works fine

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
marcingrzejszczak: Hmm... it's quite consistent to have a test here. There are already quite a few header tests in this class. I've just added one more.

ahus1:
As read it you are first expanding, then replacing. I think it should be the other way around, as you don't want to replace any plus coming in from an (encoded) variable?

marcingrzejszczak:
Yup - you're right. Fixed it.

ahus1:
+1 with icing on the top.

adriancole:
Had a sleep thought and wondering if we need to be URL encoding post
parameters in the first place. Maybe we were since the infrastructure in
the code wants us to (since it was made for path and query). Might be worth
a thought or TODO to remove URL encoding from this part of the code.

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
adriancole: Clearly we want a test with the class effected! what happens if apache is deleted or moved out of the repo? the actual change occurred to RequestTemplate.

tests in http clients are often to ensure the client isn't interfering. For example, some http clients affect headers that are set. When making a change to feign, it is best to place the test where it is most relevant and in scope. I have no reason to believe that somehow apache is going to be a better place to test a change to RequestTemplate than RequestTemplateTest. If you somehow know better, please offer that reason.

adriancole: added https://github.com/Netflix/feign/commit/b29f9d8aea5a326069275d9787eebe26a09930e3 to hopefully explain the intent of client tests

PR #326, opened 2016-02-10
make replacements header values work like in the body

ahus1:
This tries to make replacement work in the headers like it works in the body. See #288 for a discussion. You might need to URL-encode literal values for `{` or `%` in your existing code.

Old behaviour:

````java
      @Headers("Authorization: {bearer}"
      // will replace -> Authorization: 37dahdbcpadf

      @Headers("Authorization: Bearer {bearer}")
      // will not replace -> Authorization: Bearer {bearer}

      @Headers("Authorization: Bearer %7Bbearer%7B")
      // will not url decode -> Authorization: Bearer %7Bbearer%7B

      @Headers("Authorization: %25"
      // will not url decode -> Authorization: %25````
````

New behavior

````java
      @Headers("Authorization: {bearer}"
      // will replace (AS BEFORE) -> Authorization: 37dahdbcpadf

      @Headers("Authorization: Bearer {bearer}")
      // WILL replace -> Authorization: Bearer 37dahdbcpadf

      @Headers("Authorization: %7Bbearer%7B"
      // WILL url decode-> Authorization: Bearer {bearer}

      @Headers("Authorization: %25"
      // WILL url decode -> Authorization: %
````

I've also added some test cases. This PR describes the behaviour I was expecting when using Headers for the first time. But it might be a breaking change for existing users, especially that curly braces now need to encoded if used literally.

Please feel free to comment and discuss, I will update the PR as necessary

PR #348, opened 2016-03-09
Fixes vnd header regression by changing Headers encoding

### LINKED ISSUE #345: Custom content-types are broken

adriancole:
I mistakenly advised `@Headers` to follow the encoding rules of `@Body`.
This was a a mistake as in both cases, url encoding is a bad choice, if
the only goal is to prevent accidental variable expansion. For example,
url encoding interferes a lot with content, including messing with '+'
characters, such as exist in "Accept: application/vnd.github.v3+json"

This changes `@Headers` to only address the problem, which where a '{'
literal is desired in a header value. The solution offered here is to
simply repeat "{{" when you desire a '{' literal. For example, if your
header value needs to be literally "{{variable}}", you'd encode it as
"{{{{variable}}".

The impact of this change is limited to those who have already started
using v8.15, and a fast release will occur after merge to limit that.

See #326
Fixes #345
Closes #346

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -60,7 +60,6 @@ public final class RequestTemplate implements Serializable {
   private boolean decodeSlash = true;
 
   public RequestTemplate() {
-
   }
 
   /* Copy constructor. Use this when making templates. */
@@ -128,9 +127,19 @@ public final class RequestTemplate implements Serializable {
     for (char c : template.toCharArray()) {
       switch (c) {
         case '{':
+          if (inVar) {
+            // '{{' is an escape: write the brace and don't interpret as a variable
+            builder.append("{");
+            inVar = false;
+            break;
+          }
           inVar = true;
           break;
         case '}':
+          if (!inVar) { // then write the brace literally
+            builder.append('}');
+            break;
+          }
           inVar = false;
           String key = var.toString();
           Object value = variables.get(var.toString());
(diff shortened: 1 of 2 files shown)


adriancole:
@marcingrzejszczak @ahus1 I think this is a more sustainable approach to the problem. In hind sight, I wish I did this on the body annotation

spencergibb:
:+1:

PR #349, opened 2016-03-09
Adds commit log guidelines

adriancole:
I've noticed myself repeating instructions around the change log. This places guidance in the CONTRIBUTING section, as that's slightly more relevant than HACKING as it already includes policy such as license and code style.


adriancole:
@marcingrzejszczak fyi I'm logging this here as I think formally I was telling people individually.

PR #352, opened 2016-03-09
Wrong IOExceptionhandling when combining HystrixFeign and RibbonClient

einachim:


ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -69,13 +69,23 @@ public Response execute(Request request, Request.Options options) throws IOExcep
...
-        throw IOException.class.cast(e.getCause());
-      }
+      checkForIOException(e);
       throw new RuntimeException(e);
     }
   }
 
+  static void checkForIOException(Throwable throwable) throws IOException {
+
+    final List<Throwable> list = new ArrayList<Throwable>(); // for detecting cycles
```
adriancole: this is pretty edge case :) even guava doesn't guard against this. I might dumb it down, by removing this guard (and associated tests)

ribbon/src/main/java/feign/ribbon/RibbonClient.java
```diff
@@ -69,13 +69,23 @@ public Response execute(Request request, Request.Options options) throws IOExcep
       return lbClient(clientName).executeWithLoadBalancer(ribbonRequest,
           new FeignOptionsClientConfig(options)).toResponse();
     } catch (ClientException e) {
-      if (e.getCause() instanceof IOException) {
-        throw IOException.class.cast(e.getCause());
-      }
+      checkForIOException(e);
```
adriancole: maybe rename to propagateFirstIOException

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+ * limitations under the License.
+ */
+package feign.ribbon;
+
+import java.io.IOException;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class RibbonClientCheckForIOExceptionTest {
```
adriancole: rename to PropagateFirstIOExceptionTest? it is really less important that it is living in RibbonClient

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private Throwable throwable;
+
+    @Before
+    public void setUp() {
+
+        throwable = null;
+    }
+
+    @Test(expected = IOException.class)
```
adriancole: nit google code format (or I can fix formatting on the way in)

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+package feign.ribbon;
+
+import java.io.IOException;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class RibbonClientCheckForIOExceptionTest {
+
+    private Throwable throwable;
```
adriancole: this will make your life easier:
```java
  @Rule
  public ExpectedException thrown = ExpectedException.none();
```

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @Before
+    public void setUp() {
+
+        throwable = null;
+    }
+
+    @Test(expected = IOException.class)
+    public void assureThatIOExceptionIsFound() throws IOException {
+
+        givenAThrowableThatIsAnIOException();
```
adriancole: we don't use given then when in unit tests.. inlining will allow less flipping up and down the file. plus, this method is stateless, probably most straight-forward to just pass the parameter that your test method name describes.

```java
thrown.expect(IOException.class)

RibbonClient.checkForIOException(new IOException);
```

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+        whenCheckForIOExceptionIsCalled();
+    }
+
+    @Test(expected = IOException.class)
+    public void assureThatNestedIOExceptionIsFound() throws IOException {
```
adriancole: similarly, except here I'd literally javadoc saying that this actually happened!

```
/** Happened in practice; a blocking observable wrapped the connect exception in a runtime exception */
```
```java
thrown.expect(ConnectException.class)

RibbonClient.checkForIOException(new ClientException(new RuntimeException(new ConnectException())));
```

ribbon/src/test/java/feign/ribbon/RibbonClientCheckForIOExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    public void assureCyclesWillNotEndInAnEndlessLoop() throws IOException {
+
+        givenAThrowableWithCycles();
+        whenCheckForIOExceptionIsCalled();
+    }
+
+    private void givenAThrowableWithCycles() {
```
adriancole: since this is only about ribbon, make the exceptions concrete. ex always start with a clientException (unless instantiating that is somehow difficult)

adriancole:
If you're up to polishing this, let's proceed. else let me know and I can make an impl based on your work.

If you want to proceed, it would be easiest to make a branch off master, since git will be easier that way.

Thanks!

ribbon/src/main/java/feign/ribbon/RibbonClient.java
einachim: it is handled in apache commons lang. the guava Throwables methods (at least in version 14 included by the project) are marked as Beta.

ribbon/src/main/java/feign/ribbon/RibbonClient.java
einachim: +1 (i am really bad in finding names)

einachim:
It would be nice if you make an impl based on the pull request.

adriancole:
polished in https://github.com/OpenFeign/feign/pull/441

PR #354, opened 2016-03-11
Ignore static methods when validating interface

### LINKED ISSUE #312: Feign fails if static methods are not annotated
Hi,
Java 8 allows static methods on interfaces. The way I want to use this is to slightly remove boilerplate - instead of having an interface, and a static creator for it somewhere else, I'd ideally like to have the static creator inside the interface, like

    public interface Foo {
        @RequestLine("GET foo")
        String getFoo();

        @RequestLine("GET bar")
        String getBar(@Param("baz") Baz baz);

        static Foo create(String url, Feign.Builder builder) {
            builder.encoder(new FooEncoder());
            return builder.attach(Foo.class, url);
        }
    }

This (with Feign) requires the method to be annotated with a @RequestLine and @Params, like

    @RequestLine("GET IDoNotThinkThisShouldBeHere")
    static Foo create(@Param("NeitherHere") String url, @Param("NorHere") Feign.Builder builder) {
        builder.encoder(new FooEncoder());
        return builder.attach(Foo.class, url);
    }

Feign validates that the method is annotated correctly, even though any Feign implementation can never be called (as it is static). I think that Feign should ignore static methods on interfaces. The primary drawback of this is that it would be untestable - presumably the test suite must run on Java versions <= 8, where no implementation could compile.

I'd be happy to code this up if it sounds reasonable?

oillio:
This removes the check and exception for when static methods are found on an Interface being processed
fixes #312

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -9,3 +9,8 @@ dependencies {
     testCompile 'com.google.code.gson:gson:2.5' // for example
     testCompile 'org.springframework:spring-context:4.2.5.RELEASE' // for example
 }
+
+configure(compileTestJava) {
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ notifications:
     on_failure: always  # options: [always|never|change] default: always
     on_start: false     # default: false
 jdk:
-- oraclejdk7
+- oraclejdk8
 install: ./installViaTravis.sh
 script: ./buildViaTravis.sh
 cache:
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -17,6 +17,7 @@ package feign;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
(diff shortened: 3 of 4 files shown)


spencergibb:
@adriancole is animal-sniffer (or equivalent) configured that makes sure no java8 bleeds into main?

spencergibb:
otherwise, :+1:

oillio:
Animalsniffer ignores test classes by default, so no config change was needed.  I tested it out and adding 1.8 components to a main class does cause animalsiffer to fail.

Looks like Travis is not configured to handle 1.8 however.

oillio:
fixed.

core/build.gradle
```diff
@@ -9,3 +9,8 @@ dependencies {
     testCompile 'com.google.code.gson:gson:2.5' // for example
     testCompile 'org.springframework:spring-context:4.2.5.RELEASE' // for example
 }
+
+configure(compileTestJava) {
+    sourceCompatibility = 1.8
```
adriancole: there's an existing issue on source level, and looks like our last call was to re-evaluate when JDK 9 was out...
https://github.com/Netflix/feign/issues/289

adriancole: whoops compileTest! np

PR #357, opened 2016-03-15
Created decoder for jsonpath

velo:
Hey guys,

I made this decoder to meet my needs, and though could be useful for the project.

Is that something you guys consider interesting?

The nice thing about it is that I can decode rely complex json files in a simplified way using jsonPath
https://github.com/jayway/JsonPath

### DIFF
--- a/jsonpath/build.gradle
+++ b/jsonpath/build.gradle
@@ -1,0 +1,13 @@
+apply plugin: 'java'
+
+sourceCompatibility = 1.6
+
+dependencies {
+    compile project(':feign-core')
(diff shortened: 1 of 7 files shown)


adriancole:
hi, there. thanks for raising a PR. I kindof wish we had an org for integrations like this, so that they can incubate in a place where folks can find them (and therefore ask for them to be released etc). At the moment, I think this is too early for a new codec project. Do you have an account or org you can put this into, ex calling that repo feign-jsonpath?

velo:
yes I do, will do that them.  thanks

adriancole:
closed via https://github.com/Netflix/feign/issues/359

PR #360, opened 2016-03-16
Added HystrixGroupKey annotation

### LINKED ISSUE #334: allow anotation of hystrix commandKey and groupKey for interfaced feign methods
There are defaults for cmdKesy = methodname and groupKey=classname.
 If hytsrix cfg in application.yml is linked via this default names and method or class names get refactored the configuration is delinked without recognizing. Therefore explict labeling via custom commandKey and groupKey anotated above the Feign (interface) method would help to explicit link that and thre are no problems with code refactorings.

This is related to https://github.com/Netflix/feign/issues/189#issuecomment-95688652

caarlos0:
Added `@HystrixConfig` and `@HystrixCommandConfig` annotations so we can set a custom group and command key for each service, as well timeouts. We could add more config in the future.

If none is set, defaults to the previous behavior.

Syntax will be like:

```java
    @HystrixConfig(key = "MyKey", timeout = 150)
    interface Service {
        @RequestLine("GET /status")
        HystrixCommand<String> status();

        @HystrixCommandConfig(key = "Slow", timeout = 15000)
        @RequestLine("GET /slower")
        HystrixCommand<String> slower();
    }
```

this should also fix #334

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixCommandConfig.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixCommandConfig.java
@@ -1,0 +1,26 @@
+package feign.hystrix;
+
+import static java.lang.annotation.ElementType.METHOD;
+import java.lang.annotation.Retention;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import java.lang.annotation.Target;
+
+/**
+ * Adds the ability to customize hystrix command's settings:
+ * <blockquote>
+ * <pre>
+ *     interface MyService {
+ *         @HystrixCommandConfig(key = "MyCustomCommandKey", timeout = 2000)
+ *         @RequestLine("GET /blah")
+ *         String blah();
+ *     }
+ * </pre>
+ * </blockquote>
+ */
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface HystrixCommandConfig {
+    String key() default "";
+
+    int timeout() default -1;
+}
--- a/hystrix/src/main/java/feign/hystrix/HystrixConfig.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixConfig.java
@@ -6,10 +6,10 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 import java.lang.annotation.Target;
 
 /**
- * Adds the ability to define a custom hystrix key name:
+ * * Adds the ability to customize hystrix's settings:
  * <blockquote>
  * <pre>
- *     @HystrixGroupKey("MyCustomKey")
+ *     @HystrixConfig(key = "MyCustomKey", timeout = 2000)
  *     interface MyService {
  *         // methods...
  *     }
@@ -18,6 +18,8 @@ import java.lang.annotation.Target;
  */
 @Target(TYPE)
 @Retention(RUNTIME)
-public @interface HystrixGroupKey {
-    String value();
+public @interface HystrixConfig {
+    String key() default "";
+
+    int timeout() default -1;
 }
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -68,7 +68,7 @@ final class HystrixInvocationHandler implements InvocationHandler {
           return super.getFallback();
         }
         try {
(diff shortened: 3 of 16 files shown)


caarlos0:
ping?

adriancole:
sorry, I meant to comment. For example, what about the command key? For example, javanica annotations looks like this: 

`@HystrixCommand(groupKey="UserGroup", commandKey = "GetUserByIdCommand")`

I think there are others who've asked about setting hystrix group key. Can you look through similar issues and ping them on this issue? I'd like this type of feature to be diversely supported.

caarlos0:
@adriancole makes sense, I guess I can join all configs in one single annotation. Will do that and ping you back (and find other issues related).
Thanks :dancer:

adriancole:
ps for an implementation I'd just start with renaming the annotation and adding groupKey, commandKey.

However, this is kiting a bit, assuming that this is a good way to go. can you ping others who have contributed to hystrix here in the past, like @spencergibb, to see if they agree with this path?

caarlos0:
@adriancole I did it, but I think I found some thread-safety issues, wasn't able to find where though... can you give me some advice about it?

hystrix/src/main/java/feign/hystrix/HystrixCommandConfig.java
```diff
@@ -0,0 +1,26 @@
...
+ *     }
+ * </pre>
+ * </blockquote>
+ */
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface HystrixCommandConfig {
+    String key() default "";
+
+    int timeout() default -1;
```
adriancole: I've fallen down this trap before, and wouldn't recommend doing it again. timeout values often have far more to do with the network than the api, and this annotation puts this in code vs config.

hystrix/src/main/java/feign/hystrix/HystrixConfig.java
```diff
@@ -0,0 +1,25 @@
...
+ *     @HystrixConfig(key = "MyCustomKey", timeout = 2000)
+ *     interface MyService {
+ *         // methods...
+ *     }
+ * </pre>
+ * </blockquote>
+ */
+@Target(TYPE)
+@Retention(RUNTIME)
+public @interface HystrixConfig {
```
adriancole: I don't want to support multiple annotations at this point.. can we choose one?

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -106,6 +102,34 @@ protected Object getFallback() {
     return hystrixCommand.execute();
   }
 
+  private HystrixCommand.Setter config(final Object proxy, final Method method) {
```
adriancole: probably what we want to do, is make something like MethodMetadata for Hystrix. That way, parsing annotations is separate from using them. For example, we don't want to replicate javanica. We can have a simple way to specify command and group names, and leave any further settings to either out-of-band configuration or another annotation system.

hystrix/src/main/java/feign/hystrix/HystrixCommandConfig.java
caarlos0: makes sense, so, how can I change the timeout of a hystrix command if it was already "built" by feign?

hystrix/src/main/java/feign/hystrix/HystrixConfig.java
caarlos0: I guess we can...

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
caarlos0: I'm not familiar with javanica..

Not sure I understand your idea... can you elaborate?

apischan:
Whether this pull request is dead?

adriancole:
/me pokes with a stick

Seems like it. We should generally close dead PRs since they distract folks.

caarlos0:
Yeah, I was kind of waiting for a reply here: https://github.com/OpenFeign/feign/pull/360#discussion_r58198701

caarlos0:
but, np, will close it.

adriancole:
I meant something like this:
https://github.com/OpenFeign/feign/issues/434#issuecomment-239609848

PR #366, opened 2016-03-23
Update README with documentation on setting headers

nmiyake:
Document setting headers at api level using @Headers
or at client level using RequestInterceptor or Target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,78 @@ client.xml("denominator", "secret"); // <login "user_name"="denominator" "passwo
 client.json("denominator", "secret"); // {"user_name": "denominator", "password": "secret"}
 ```
 
+### Headers
+Feign supports settings headers on requests either as part of the api or as part of the client
+dending on the use case.
+
+#### Set headers using apis
+In cases where specific interfaces or calls should always have certain header values set, it
+makes sense to define headers as part of the api.
+
+Static headers can be set on an api interface or method using the `@Headers` annotation.
+
+```java
+@Headers("Accept: application/json")
+interface BaseApi<V> {
+  @Headers("Content-Type: application/json")
+  @RequestLine("PUT /api/{key}")
+  void put(@Param("key") String, V value);
+}
+```
+
+Methods can specify dynamic content for static headers using using variable expansion in `@Headers`.
+
+```java
+ @RequestLine("POST /")
+ @Headers("X-Ping: {token}")
+ void post(@Param("token") String token);
+```
+
+These approaches specify specific header entries as part of the api without requiring any customizations
+when buildling Feing clients. It is not currently possible to customize the header entries themselves
+on a per-request basis at the api level.
+
+#### Setting headers per target
+In cases where headers should differ for the same api based on different endpoints or where per-request
+customization is required, headers can be set as part of the client using a `RequestInterceptor` or a
+`Target`.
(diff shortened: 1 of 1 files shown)


nmiyake:
@adriancole

README.md
```diff
@@ -246,6 +246,70 @@ client.xml("denominator", "secret"); // <login "user_name"="denominator" "passwo
...
+
+```java
+ @RequestLine("POST /")
+ @Headers({
+  "X-Foo: Bar",
+  "X-Ping: {token}"
+ }) void post(@Param("token") String token);
+```
+
+These approaches specify specific header entries as part of the api without requiring any cutomizations
```
adriancole: cutomizations

README.md
```diff
@@ -246,6 +246,70 @@ client.xml("denominator", "secret"); // <login "user_name"="denominator" "passwo
...
+when buildling Feing clients. It is not currently possible to customize the header entries themselves
+on a per-request basis at the api level.
+
+#### Set headers using client
```
adriancole: Setting headers per target

README.md
```diff
@@ -246,6 +246,70 @@ client.xml("denominator", "secret"); // <login "user_name"="denominator" "passwo
...
+
+```java
+  static class ForwardedForTarget extends HardCodedTarget {
+    public ForwardedForTarget(Class type, String url) {
+      super(type, url);
+    }
+
+    @Override
+    public Request apply(RequestTemplate input) {
+      input.header("X-Forwarded-For", "origin.host.com");
```
adriancole: not a very realistic target..

Target's main use has been adding authentication headers or anything tha varies on the url.
Have a look at denominator for examples. https://github.com/Netflix/denominator/find/master

README.md
```diff
@@ -246,6 +246,70 @@ client.xml("denominator", "secret"); // <login "user_name"="denominator" "passwo
...
+      return super.apply(input);
+    }
+  }
+  ...
+  Bank bank = Feign.builder()
+          .decoder(accountDecoder)
+          .target(new ForwardedForTarget(Bank.class, "https://api.examplebank.com"));
+```
+
+These approaches depend on the custom `RequestInterceptor` or `Target` being set on the Feign
```
adriancole: Maybe reword it, as many things are implicit, and this is not different than things like encoders, or anything else in feign builder. What's important is the calling thread.

When setting dynamic headers with  `RequestInterceptor` or `Target`, bear in mind that `apply` is called as a side effect of an api call: that can be made from any thread.

In simplest case, all callers get the same value. For example, an authentication token is set in a thread-safe fashion and shared for all callers.

Another use case is context-specific values. For example, an tracing id is set in a thread-local upstream to Feign. The target or request interceptor reads that context and adds a trace header accordingly.

adriancole:
looks good. main recommendation is to use concrete examples here. I think authentication token and trace ids explain the thread-safety concern very well. Plus, in most cases, you'd never want the user to have to explicitly provide a auth token or trace context.. this reinforces common implicit uses and the underlying point about thread safety.

hope this helps

nmiyake:
Thanks for the feedback -- made the example more realistic (setting an auth token and thread-local trace id) and provided more specific examples in documentation.

adriancole:
Nicely done.

Issue #353, opened 2016-03-11
Feign Hystrix fallback can't access method in packaged interface 


liuzhengyang:
If apiType is a package interface, HystrixCommand can not invoke the fallback method, so the hystrix fallback will fail too.
Here is my sample code.
```
interface Github {
	@RequestLine("GET /")
	String get();
}

public class Main2 {
	public static Github fallback = () -> "i am fallback";

	public static void main(String[] args) {
		Github target = HystrixFeign.builder().target(Github.class, "http://api.github.com/ss", fallback);
		System.out.println(target.get());
	}
}
```
And the exception:
```
Caused by: java.lang.IllegalAccessException: Class feign.hystrix.HystrixInvocationHandler$1 can not access a member of class com.lzy.main.p2.Github with modifiers "public abstract"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Method.invoke(Method.java:490)
	at feign.hystrix.HystrixInvocationHandler$1.getFallback(HystrixInvocationHandler.java:75)
	... 39 common frames omitted
com.netflix.hystrix.exception.HystrixRuntimeException: get timed-out and fallback failed.
	at com.netflix.hystrix.AbstractCommand$16.call(AbstractCommand.java:816)
	at com.netflix.hystrix.AbstractCommand$16.call(AbstractCommand.java:790)
	at rx.internal
```
I try to fix this by using method.setAccessible.But it does not look very graceful.
https://github.com/liuzhengyang/feign/commit/b4fdd67f5567530736686179b1f18790604cb828

adriancole:
interesting. I wonder if we can do this set accessible thing up-front, for
example, in the builder or the contract?

caarlos0:
are the two classes in the same package (Github and Main2)?

adriancole:
Good point. If these classes are in different jars I don't think set
accessible will save you iirc.

liuzhengyang:
@caarlos0，yes. In my example, they are in the same file.

liuzhengyang:
If i rename the package name of HystrixBuilderTest class  from `feign.hystrix` to `feign.hystrixtest`, some test cases will fail.
After i add this code snatch to `HystrixInvocationHandler`. We can passes test cases.
```
[...]

PR #367, opened 2016-03-23
Set method access for hystrix package-private fallback method.

liuzhengyang:
For interfaces with package-private access, hystrix fallback calls would fail. Address this by caching methods to reduce the amount of reflection calls.
Address #353.

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -22,6 +22,7 @@ import com.netflix.hystrix.HystrixCommandKey;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.Map;
 
 import feign.InvocationHandlerFactory;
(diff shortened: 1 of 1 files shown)


liuzhengyang:
@adriancole , please take a look whether this is suitable. I am not clear how to set accessible in contract or builder.

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -38,11 +39,23 @@
...
     this.dispatch = checkNotNull(dispatch, "dispatch");
     this.fallback = fallback;
+    this.toFallback = toFallbackMethod(dispatch);
+  }
+
+  private Map<Method, Method> toFallbackMethod(Map<Method, MethodHandler> dispatch) {
+    Map<Method, Method> result = new HashMap<Method, Method>();
+    for (Map.Entry<Method, MethodHandler> entry : dispatch.entrySet()) {
+      Method key = entry.getKey();
+      key.setAccessible(true);
```
adriancole: setAccessible is mutating, so you don't need to save this off

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -38,11 +39,23 @@
   private final Target<?> target;
   private final Map<Method, MethodHandler> dispatch;
   private final Object fallback; // Nullable
+  private final Map<Method, Method> toFallback;
 
   HystrixInvocationHandler(Target<?> target, Map<Method, MethodHandler> dispatch, Object fallback) {
     this.target = checkNotNull(target, "target");
     this.dispatch = checkNotNull(dispatch, "dispatch");
     this.fallback = fallback;
+    this.toFallback = toFallbackMethod(dispatch);
```
adriancole: do you need to save this? What happens if you just do this instead?
```
this.fallback = fallback;
// make sure we can invoke fallbacks on package-private interfaces
for (Method method : dispatch.keySet()) {
  method.setAccessible(true);
}
```

liuzhengyang: I do have tried this way at first. But it doesn't take effect. The method injected in InvocationHandler is not the same as method in dispatch, though they are equals. So i bypass this problem by saving a map.

adriancole: OK in that case, it truly deserves a comment! can you add one?

liuzhengyang: Of course

liuzhengyang:
Comments added. While it's strange that the jenkins job failed.https://netflixoss.ci.cloudbees.com/job/NetflixOSS/job/feign/job/feign-pull-requests/264/console.

adriancole:
looks fine. I'll squash these commits on the way in.

adriancole:
closed via 2ef3bc80c33a57e31f363d0263b0969b7faf2908

adriancole:
Thank you!

liuzhengyang:
exciting

PR #369, opened 2016-03-28
Support default methods on interface

### LINKED ISSUE #289: Move language level to 1.7
Android supports Java language level 1.7 features, even if we should still enforce use of 1.6 apis via animal-sniffer. This should reduce clutter in declarations.

This implies minSdkVersion 19

See https://github.com/square/wire/pull/370

oillio:
With this change, interfaces targeted by Feign may have static or default methods (if using Java 8+).

These allows Feign clients to contain logic that is not expressly defined by the underlying API.

For example, static methods make it easy to specify common client build configurations; default methods can be used to compose queries or define default parameters.
```java
interface GitHub {
  List<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);

  @RequestLine("GET /users/{username}/repos?sort={sort}")
  List<Repo> repos(@Param("username") String owner, @Param("sort") String sort);

  default List<Repo> repos(String owner) {
    return repos(owner, "full_name");
  }

  /**
   * Lists all contributors for all repos owned by a user.
   */
  default List<Contributor> contributors(String user) {
    MergingContributorList contributors = new MergingContributorList();
    for(Repo repo : this.repos(owner)) {
      contributors.addAll(this.contributors(user, repo.getName()));
    }
    return contributors.mergeResult();
  }

  static GitHub connect() {
    return Feign.builder()
                .decoder(new GsonDecoder())
                .target(GitHub.class, "https://api.github.com");
  }
}
```

See issue #312 for more background.
fixes #289

### DIFF
--- a/core/src/main/java/feign/DefaultMethodHandler.java
+++ b/core/src/main/java/feign/DefaultMethodHandler.java
@@ -1,0 +1,62 @@
+package feign;
+
+import feign.InvocationHandlerFactory.MethodHandler;
+import org.jvnet.animal_sniffer.IgnoreJRERequirement;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+/**
+ * Handles default methods by directly invoking the default method code on the interface.
+ * The bindTo method must be called on the result before invoke is called.
+ */
+@IgnoreJRERequirement
+final class DefaultMethodHandler implements MethodHandler {
+  // Uses Java 7 MethodHandle based reflection.  As default methods will only exist when
+  // run on a Java 8 JVM this will not affect use on legacy JVMs.
+  // When Feign upgrades to Java 7, remove the @IgnoreJRERequirement annotation.
+  private final MethodHandle unboundHandle;
+
+  // handle is effectively final after bindTo has been called.
+  private MethodHandle handle;
+
+  public DefaultMethodHandler(Method defaultMethod) {
+    try {
+      Class<?> declaringClass = defaultMethod.getDeclaringClass();
+      Field field = Lookup.class.getDeclaredField("IMPL_LOOKUP");
+      field.setAccessible(true);
+      Lookup lookup = (Lookup) field.get(null);
+
+      this.unboundHandle = lookup.unreflectSpecial(defaultMethod, declaringClass);
+    } catch (NoSuchFieldException ex) {
+      throw new IllegalStateException(ex);
+    } catch (IllegalAccessException ex) {
+      throw new IllegalStateException(ex);
+    }
+  }
+
+  /**
+   * Bind this handler to a proxy object.  After bound, DefaultMethodHandler#invoke will act as if it was called
+   * on the proxy object.  Must be called once and only once for a given instance of DefaultMethodHandler
+   */
+  public void bindTo(Object proxy) {
+    if(handle != null) {
+      throw new IllegalStateException("Attempted to rebind a default method handler that was already bound");
+    }
+    handle = unboundHandle.bindTo(proxy);
+  }
+
+  /**
+   * Invoke this method.  DefaultMethodHandler#bindTo must be called before the first
+   * time invoke is called.
+   */
+  @Override
+  public Object invoke(Object[] argv) throws Throwable {
+    if(handle == null) {
+      throw new IllegalStateException("Default method handler invoked before proxy has been bound.");
+    }
+    return handle.invokeWithArguments(argv);
+  }
+}
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -18,12 +18,7 @@ package feign;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
 
 import feign.InvocationHandlerFactory.MethodHandler;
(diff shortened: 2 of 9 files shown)


adriancole:
I think we can make this work without raising the language level. For example, I think you can get by with using an animal sniffer ignore annotation. I'll make a note for something to try.

We suggested holding off until september to raise the language level, so let's see if we can honor that.

core/src/main/java/feign/Util.java
```diff
@@ -21,10 +21,7 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Reader;
-import java.lang.reflect.Array;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
-import java.lang.reflect.WildcardType;
+import java.lang.reflect.*;
```
adriancole: no wildcard imports please

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -93,10 +96,22 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         return hashCode();
       } else if ("toString".equals(method.getName())) {
         return toString();
+      } else if (Util.isDefault(method)) {
+        return invokeDefault(proxy,method, args);
       }
       return dispatch.get(method).invoke(args);
     }
 
+    private Object invokeDefault(Object proxy, Method method, Object[] args) throws Throwable {
```
adriancole: Since this is guarded by Util.isDefault, we can assume that the language level isn't 6 when we call this.

You should be able to get by with `@IgnoreJRERequirement  // Should only be invoked on Java 7+.` on this method

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -93,10 +96,22 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         return hashCode();
       } else if ("toString".equals(method.getName())) {
         return toString();
+      } else if (Util.isDefault(method)) {
+        return invokeDefault(proxy,method, args);
```
adriancole: nit formatting space between methods

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -93,10 +96,22 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
       }
       return dispatch.get(method).invoke(args);
     }
 
+    private Object invokeDefault(Object proxy, Method method, Object[] args) throws Throwable {
+      Class<?> declaringClass = method.getDeclaringClass();
+      Field field = Lookup.class.getDeclaredField("IMPL_LOOKUP");
+      field.setAccessible(true);
+      Lookup lookup = (Lookup) field.get(null);
+      return lookup.unreflectSpecial(method, declaringClass)
```
adriancole: benchmarking might find this method is expensive. if so, we might try to cache the MethodHandle lookups in the ctor, similar to `dispatch`. Ack this might be tricky to have work in JRE 6.

Ex you might have to use a field like `Map<Method, Object> defaultMethods`, and then do a containsKey on it and cast out.. like..

```java
} else if (defaultMethods.containsKey(method)) {
  return invokeDefault(method);
}
--snip--

@IgnoreJRERequirement // Should only be invoked on Java 7+
Object invokeDefault(Method defaultMethod) throws Exception {
  MethodHandle handle = (MethodHandle) defaultMethods.get(defaultMethod);
  return handle.invokeWithArguments(args);
}
...


```

adriancole:
nice work. I made some suggestions that hopefully can rollback the language level and signature updates. I don't see anything here that needs a language level update. The only signature issue that's obvious seems guarded enough to work with an animal sniffer exclusion.

If I'm not noticing something, please speak up!

oillio:
I think that will work.  I have never intentionally used later model classes when targeting a legacy JVM, but I don't see any reason why it won't work.

I will work on caching the MethodHandles tomorrow.

adriancole:
cool I'll hold 8.16 till then

oillio:
OK, check out the new code.
Using a separate map for the default method handles was getting pretty messy as it would necessitate a change to InvocationHandlerFactory and the mess with Map<Method,Object>.

I think I found a better way by overloading MethodHandler.  Let me know what you think.
Adding the proxy call to MethodHandler#invoke made everything pretty clean.  If you think that will cause problems, I should be able to get the proxy into my MethodHandler using a Promise or something.  It won't be nearly as pretty/

core/src/main/java/feign/InvocationHandlerFactory.java
```diff
@@ -32,7 +32,7 @@
    */
   interface MethodHandler {
 
-    Object invoke(Object[] argv) throws Throwable;
+    Object invoke(Object proxy, Object[] argv) throws Throwable;
```
adriancole: unfortunately we can't do this as it will break api

adriancole: ex this will at least break spring-cloud-netflix and possibly others.. that would mean a major version bump

oillio:
OK, refactored to pass in proxy using promise.
How's this look?

core/src/main/java/feign/Promise.java
```diff
@@ -0,0 +1,25 @@
+package feign;
+
+public class Promise<T> {
```
adriancole: promise is a loaded term, although I get what you are doing. how about using an AtomicReference instead? that way, we don't need to test as much.

adriancole: actually I've another idea..

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -57,15 +57,21 @@
   public <T> T newInstance(Target<T> target) {
     Map<String, MethodHandler> nameToHandler = targetToHandlersByName.apply(target);
     Map<Method, MethodHandler> methodToHandler = new LinkedHashMap<Method, MethodHandler>();
+    Promise<Object> proxyPromise = new Promise<Object>();
+
     for (Method method : target.type().getMethods()) {
       if (method.getDeclaringClass() == Object.class) {
         continue;
+      } else if(Util.isDefault(method)) {
+        methodToHandler.put(method, new DefaultMethodHandler(method, proxyPromise));
```
adriancole: here's another idea.. since this is mutable, skip the promise/ref thing. after you create the handler, loop through methodToHandler and bind the proxy before returning. this should be more thread safe.

ex. 
```java

boolean supportsDefault = false;
for (Method method : ...
...
} else if(Util.isDefault(method)) {
  supportsDefault = true;
...
T proxy = ...
if (supportsDefault) { // guards against execution in old JREs
  for (MethodHandler handler: methodToHandler.values()) {
    if (handler instanceof DefaultMethodHandler) {
      ((DefaultMethodHandler)handler).bindTo(proxy);
    }
  }
}
```

core/src/main/java/feign/DefaultMethodHandler.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@IgnoreJRERequirement
+public class DefaultMethodHandler implements MethodHandler {
+  // Uses Java 7 MethodHandle based reflection.  As default methods will only exist when
+  // run on a Java 8 JVM this will not affect use on legacy JVMs.
+  // When Feign upgrades to Java 7, remove the @IgnoreJRERequirement annotation.
+  private final MethodHandle unboundHandle;
+  private final Promise<Object> proxy;
+
+  private MethodHandle handle;
```
adriancole: once other things happen, you can comment this is effectively final

core/src/main/java/feign/ReflectiveFeign.java
oillio: Sounds good.  I'll write it up this evening.

adriancole: great. some touches that might make me be able to merge right away:

* make DefaultMethodHandler package private and final (we don't want
people trying to mutate or extend this)
* update CHANGES with a note about supporting default methods
* put a cool example in README about what this allows you to do! ex
Github api = Github.connect();

oillio:
OK, dropped the promise and added documentation.

adriancole:
Thanks for all the hard work! nit storm over.. time to ship

PR #376, opened 2016-04-15
8.x

didvae:


adriancole:
I think you might have mistakenly opened this PR. It looks like an attempt to merge a fork's changes into 8.x. If you've something to propose, it would be easier to understand in a branch and ideally a single commit.

PR #377, opened 2016-04-19
Allows for multiple inheritance in clients

bandrzejczak:
In `feign.Client` lines 51 through 56, there were checks to ensure no multiple inheritance in interfaces used to build clients. After removing those lines it turned out, that multiple inheritance actually works just fine. To ensure that, I've added a test class `MultipleInheratanceTest`.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -19,11 +19,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


adriancole:
There are subtleties to multiple inheritance ex clashing method signatures,
how hystrix interops etc. Please open this as an issue. If many people ask
for it, and the implementation comes with tests, we could consider it, but
I'd say it would be unlikely.

adriancole:
cleaning the desk.. since this hasn't been mentioned in a while, closing. Again, folks can later mention interest in this feature as they'd do a search and find the closed PR. If that happened a lot, we could re-consider it.

PR #387, opened 2016-04-28
Added rx.Completable support

### LINKED ISSUE #379: Hystrix RxJava Completable support
Currently when I'm doing a HystrixFeign client and I didn't expect a body, I use Observable<Void> but It sends a 'null' event. I would expect just a onComplete event if everything works ok, or onError otherwise.

I think behaivior cannot be changed, because it will break backward compatibility, but there is another recommended way to deal with this?

Maybe a good idea add support to Completable in the future, when someone removes that ugly Experimental annotation :)

padilo:
- Added ```rx.Completable``` support
- Upgraded hystrix-core to have a newer version of rxjava with ```Completable``` in theory it can be upgraded to 1.5.x as it should be backward compatible.

Fixes #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Version 8.17
+* Adds support to RxJava Completable via `HystrixFeign` builder with fallback support
+* Upgraded hystrix-core to 1.4.26
+
 ### Version 8.16
 * Adds `@HeaderMap` annotation to support dynamic header fields and values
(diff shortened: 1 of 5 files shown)


adriancole:
nice work. maybe add to CHANGES for 8.17?

padilo:
Thank you, done. Tell me if anything more is needed :)

adriancole:
nope. all good!

PR #390, opened 2016-05-05
Upgrade to OkHttp 3.2 and adjust tests to work with new version

bjsousa:
Upgrading to OkHttp3 will allow applications with Netflix-Feign to use enhanced OkHttpClient capabilities, especially the newly introduced [Builder method](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html).

In addition to import and compile statement changes, method name changes were needed in multiple tests to work with the new dependency version.  In the OkHttpClient class, clone() was replaced with newBuilder() as a way to change client settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 8.18
+* Upgrades dependency version for OkHttp/MockWebServer 3.2.0
+
 ### Version 8.17
 * Adds support to RxJava Completable via `HystrixFeign` builder with fallback support
 * Upgraded hystrix-core to 1.4.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,7 @@
-### Version 8.18
-* Upgrades dependency version for OkHttp/MockWebServer 3.2.0
-
 ### Version 8.17
 * Adds support to RxJava Completable via `HystrixFeign` builder with fallback support
 * Upgraded hystrix-core to 1.4.26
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+### Version 8.18
```
adriancole: 8.17 hasn't been cut, yet, so you can put this there

adriancole:
LGTM, just nit on changelog.

Thanks for taking time to do this!

bjsousa:
No problem!  Feign is a huge help to us in a project we're working on at UW-Madison.

tburch:
Thanks for this update!

tburch:
@adriancole Any idea when you'll do the next release that includes this PR?

adriancole:
Thanks for asking! v8.17.0 is on the way

j-baker:
We've got a whole load of diamond dependencies with Feign and feign-okhttp, and this makes Feign 8.17.0 binary-incompatible with 8.16.0. This is causing us plenty of grief.

Since okhttp3 uses a different package name and so peacefully coexists with okhttp2, any chance we could create a feign-okhttp3 package, put the results of this PR in there, keep the old code in here, and publish an 8.18.0 which reverts the breaking change to feign-okhttp?

If this sounds like an appropriate course of action, I'd be happy to contribute it.

adriancole:
@j-baker right. go for it!

PR #394, opened 2016-05-17
add support for OkHttp 3.x

tburch:
I'd like to propose adding support for OkHttp 3.x since that's now the latest stable version from Square.


tburch:
Missed that #390 was already open...

PR #403, opened 2016-06-17
Add support for expansion of @Param lists

dsyer:
The existing support for expanders in method parameters is
limited to converting a single value. The change applies the
expander individually to each item in a collection or array,
thus making it useful for multi-valued query parameters, for
instance. The old behaviour is preserved because no existing
expanders would have been converting collections to strings
(probably).

### DIFF
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -227,8 +227,6 @@ public class ReflectiveFeign extends Feign {
     private Object expandElements(Expander expander, Object value) {
       if (value instanceof Iterable) {
         return expandIterable(expander, (Iterable) value);
-      } else if (value.getClass().isArray()) {
-        return expandIterable(expander, Arrays.asList((Object[])value));
       }
       return expander.expand(value);
     }
@@ -236,9 +234,7 @@ public class ReflectiveFeign extends Feign {
     private List<String> expandIterable(Expander expander, Iterable value) {
       List<String> values = new ArrayList<String>();
       for (Object element : (Iterable) value) {
-        if (element==null) {
-          values.add("");
-        } else {
+        if (element!=null) {
           values.add(expander.expand(element));
         }
       }
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -202,7 +202,7 @@ public class ReflectiveFeign extends Feign {
         Object value = argv[entry.getKey()];
         if (value != null) { // Null values are skipped.
           if (indexToExpander.containsKey(i)) {
-            value = indexToExpander.get(i).expand(value);
+            value = expandElements(indexToExpander.get(i), value);
           }
           for (String name : entry.getValue()) {
             varBuilder.put(name, value);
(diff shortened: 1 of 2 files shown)


dsyer:
I had to disable animal sniffer to get the build to work locally. If it works for other people I'm interested in knowing what I did wrong.

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -224,6 +224,27 @@ public RequestTemplate create(Object[] argv) {
       return template;
     }
 
+    private Object expandElements(Expander expander, Object value) {
```
adriancole: For consistency I think we should check instance of Iterable not, Collection or array. Currently, we don't have any special handling for arrays, while we do for Iterable (HeaderMap and QueryMap)

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -224,6 +224,27 @@ public RequestTemplate create(Object[] argv) {
...
+        return expandCollection(expander, Arrays.asList((Object[])value));
+      }
+      return expander.expand(value);
+    }
+
+    private List<String> expandCollection(Expander expander, Collection value) {
+      List<String> values = new ArrayList<String>();
+      for (Object element : (Collection) value) {
+        if (element==null) {
+          values.add("");
```
adriancole: eek.. I'd probably skip null vs add an empty string. (ex users can be told to add empty strings if they want empty strings). Either way, should be tested.

core/src/test/java/feign/FeignTest.java
```diff
@@ -218,6 +218,18 @@ public void customExpander() throws Exception {
   }
 
   @Test
+  public void customExpanderListParam() throws Exception {
+    server.enqueue(new MockResponse());
+
+    TestInterface api = new TestInterfaceBuilder().target("http://localhost:" + server.getPort());
+
+    api.expandList(Arrays.asList(new Date(1234l)));
```
adriancole: if folks are interested in this change, I'd test multiple values and also skipping of null. that way, the key behaviors are locked and understood (even if debatable)

core/src/main/java/feign/ReflectiveFeign.java
dsyer: I disagree. Null is the only sensible placeholder I can think of for "empty but not a string".

core/src/main/java/feign/ReflectiveFeign.java
dsyer: Done

core/src/test/java/feign/FeignTest.java
dsyer: Done

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -224,6 +224,27 @@ public RequestTemplate create(Object[] argv) {
       return template;
     }
 
+    private Object expandElements(Expander expander, Object value) {
+      if (value instanceof Iterable) {
+        return expandIterable(expander, (Iterable) value);
+      } else if (value.getClass().isArray()) {
+        return expandIterable(expander, Arrays.asList((Object[])value));
```
adriancole: why are we adding array support?

adriancole:
I'm really against weird rules like:

Ex. if you have a collection that permits nulls, and it isn't a collection of strings, put null and we'll put in an empty string for you!

I'm also against expanding contracts from iterable to also array params inconsistently (also at all)

the bug surface area is increasing for no apparent reason. We've not had a single request for null element collections, nor array params. Feign doesn't kite ahead of requirements as it keeps the code less dodgy longer. If we want to change the change policy of feign, we can in open feign provided others besides me are there to fix all of the new bugs we make possible.

dsyer:
I think you're being unnecessarily defensive, but I'm happy to rip out the array params and null handling features (see last commit).

adriancole:
it is just optimizing for the least code and tests for the functionality.
I've always been defensive on feign

adriancole:
thanks for doing this despite my grumbling. cherry-picking into master, too

Issue #373, opened 2016-04-12
Move to a new org


adriancole:
Netflix was the original sponsor of Feign, but it isn't used internally anymore. Plus, since we aren't in an org, we have no place to host things like jaxrs2 or other optional repositories.

I think we have sufficient activity to carry-on, so should we move on to a new org?

kennedyoliveira:
@adriancole, off topic but just for curiosity, what is Netflix using instead of Feign?

adriancole:
Due to freedom+responsibility, I imagine there are multiple answers to
this. One thing I noticed recently was Ribbon
<https://github.com/Netflix/ribbon/blob/master/ribbon-examples/src/main/java/com/netflix/ribbon/examples/rx/proxy/MovieService.java>,
which has declarative annotations. There are probably non-netflix projects
in-use, too. Folks currently working at Netflix are free to chime in, as
they'd know better than me what's working for them.

adriancole:
Perhaps corny, but what if we used https://github.com/feigntown? Basically, it would be the place where all the feign projects hangout.

xxlabaza:
Sounds cool. What shall I do to get in?

adriancole:
It needs to be created first, and probably needs a few more acks before we
proceed.

pmvilaca:
As feign isn't used internally anymore, will Netflix continue supporting it or it'll managed by the community?

spencergibb:
@pmvilaca, it wasn't ever heavily used at Netflix. @adriancole created it while he was there for another project (denominator) and has been maintained by him since he left. Moving out of Netflix will allow for more community involvement.

adriancole:
https://github.com/FeignTown it is then!

We'll keep this open until it is functional

adriancole:
so first steps is we need to get some infrastructure in place. For example, publishing requires a bunch of work on the sonatype and bintray side of things, then we'd need to update the build to use it.

First decisions are about group id and bintray id. For example, we could use feign.io, but we should probably do the DNS registration (esp helpful if someone offers to steward that domain and grant access to admins of the org).

[...]

PR #404, opened 2016-06-17
Migrate to maven move to org io.github.openfeign

spencergibb:
/cc @adriancole for your perusing pleasure :-)


spencergibb:
Still missing any publishing framework, but we could adapt what is done for openzipkin.

dsyer:
Why is it "feigntown"? (It's just hard to remember.)

spencergibb:
@dsyer feign is taken. Open to suggestions. OpenFeign?

spencergibb:
See #373 for adrian's suggestion.

adriancole:
We should probably migrate this to the new org before merging this. before that, we need sonatype etc setup.

benchmark/src/main/java/feign/benchmark/RealRequestBenchmarks.java
```diff
@@ -51,7 +51,7 @@ public void setup() {
     });
     server.start();
     client = new OkHttpClient();
-    client.setRetryOnConnectionFailure(false);
+// FIXME openfeign   client.setRetryOnConnectionFailure(false);
```
adriancole: elaborate? :)

jackson-jaxb/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+  </developers>
+  <licenses>
+    <license>
+      <name>The Apache Software License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
+    </license>
+  </licenses>
+  <properties>
+    <nebula_Manifest_Version>1.0</nebula_Manifest_Version>
```
adriancole: wow.. we need all these things?

dsyer: They could go in a parent pom. But you need them to comply with Sonatype's rules about Maven Central artifacts (which are not always enforced, but it's good citizens would do it).

adriancole: I was referring to all the nebula stuff.. sorry should have indicated downwards

benchmark/src/main/java/feign/benchmark/RealRequestBenchmarks.java
spencergibb: `setRetryOnConnectionFailure` doesn't exist on `okhttp3.OkHttpClient`.

jackson-jaxb/pom.xml
spencergibb: Sorry, I missed one pom. I removed them all from the others.

adriancole:
closed via #416 thanks for the help!

PR #409, opened 2016-06-29
Fixed clearing query param with unsafe characters in name

edio:
queries map uses encoded names as keys, but unencoded name was passed to `remove()`


adriancole:
Looks great. thanks!

PR #415, opened 2016-07-01
XXE fixes in JAXBDecoder and SAXDecoder classes 

gursev:
Thank you for the feedback. Pull request as per the feedback.


adriancole:
thanks! I'll raise a follow-up with tests (if it ends up easy enough)

PR #416, opened 2016-07-02
Adds maven project under group id io.github.openfeign

adriancole:
This will allow us to fork into the openfeign org and test

based on #404


spencergibb:
👍

PR #418, opened 2016-07-08
Lower case headers in response and use TreeMap to allow case insensitive access

a-k-g:
As per the discussion on https://gitter.im/Netflix/feign

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   * Previously the OkhttpClient would throw an exception, and ApacheHttpClient
     would report a wrong, possibly negative value
 * Adds support for encoded query parameters in `@QueryMap` via `@QueryMap(encoded = true)`
+* Keys in `Response.headers` are now lower-cased. This map is now case-insensitive with regards to keys,
+  and iterates in lexicographic order.
+  * This is a step towards supporting http2, as header names in http1 are treated as case-insensitive
+    and http2 down-cases header names.
 
 ### Version 8.17
 * Adds support to RxJava Completable via `HystrixFeign` builder with fallback support
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -24,8 +24,10 @@ import java.io.Reader;
 import java.nio.charset.Charset;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.Locale;
 import java.util.Map;
+import java.util.TreeMap;
 
 import static feign.Util.UTF_8;
 import static feign.Util.checkNotNull;
@@ -47,10 +49,7 @@ public final class Response implements Closeable {
     checkState(status >= 200, "Invalid status code: %s", status);
     this.status = status;
     this.reason = reason; //nullable
-    LinkedHashMap<String, Collection<String>> copyOf =
-        new LinkedHashMap<String, Collection<String>>();
-    copyOf.putAll(checkNotNull(headers, "headers"));
-    this.headers = Collections.unmodifiableMap(copyOf);
+    this.headers = Collections.unmodifiableMap(caseInsensitiveCopyOf(headers));
     this.body = body; //nullable
   }
 
@@ -92,6 +91,9 @@ public final class Response implements Closeable {
     return reason;
   }
 
+  /**
+   * Returns a case-insensitive mapping of header names to their values.
+   */
   public Map<String, Collection<String>> headers() {
     return headers;
   }
@@ -241,5 +243,18 @@ public final class Response implements Closeable {
     public String toString() {
       return decodeOrDefault(data, UTF_8, "Binary data");
     }
+  }
+
+  private static Map<String, Collection<String>> caseInsensitiveCopyOf(Map<String, Collection<String>> headers) {
+    Map<String, Collection<String>> result = new TreeMap<String, Collection<String>>(String.CASE_INSENSITIVE_ORDER);
+
+    for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
+      String headerName = entry.getKey();
+      if (!result.containsKey(headerName)) {
+        result.put(headerName.toLowerCase(Locale.ROOT), new LinkedList<String>());
+      }
+      result.get(headerName).addAll(entry.getValue());
+    }
+    return result;
   }
 }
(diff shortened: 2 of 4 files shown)


core/src/main/java/feign/Response.java
```diff
@@ -92,6 +91,11 @@ public String reason() {
     return reason;
   }
 
+  /**
+   * Map of header names to collection of values. This is a TreeMap which uses a case-insensitive
```
adriancole: no need to talk about implementation (in case it one day changes), you can just say..

```
/** Returns a case-insensitive mapping of header names to their values. */
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * Content-Length response bodies with lengths greater than Integer.MAX_VALUE report null length
   * Previously the OkhttpClient would throw an exception, and ApacheHttpClient
     would report a wrong, possibly negative value
+* The `headers` map in the `Response` object now lower-cases header names. It is now a `TreeMap` with
```
adriancole: maybe something like, which helps with motivation and talks about impact vs impl.

```
* `Response.headers` is now case-insensitive with regards to keys, and iterates in lexicographic order.
  * This better supports http2, which down-cases header names.
```

core/src/main/java/feign/Response.java
```diff
@@ -242,4 +246,19 @@ public String toString() {
...
+
+  private static Map<String, Collection<String>> caseInsensitiveCopyOf(Map<String, Collection<String>> headers) {
+    Map<String, Collection<String>> result = new TreeMap<String, Collection<String>>(String.CASE_INSENSITIVE_ORDER);
+
+    for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
+      String headerName = entry.getKey();
+      Collection<String> values = result.get(headerName);
+      if (values == null) {
+        values = new LinkedList<String>();
+        result.put(headerName.toLowerCase(Locale.US), values);
```
adriancole: I think `Local.ROOT` is probably a better choice.

core/src/main/java/feign/Response.java
```diff
@@ -242,4 +246,19 @@ public String toString() {
       return decodeOrDefault(data, UTF_8, "Binary data");
     }
   }
+
+  private static Map<String, Collection<String>> caseInsensitiveCopyOf(Map<String, Collection<String>> headers) {
+    Map<String, Collection<String>> result = new TreeMap<String, Collection<String>>(String.CASE_INSENSITIVE_ORDER);
+
+    for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
+      String headerName = entry.getKey();
```
adriancole: nit: can save a line by moving the defense up.

ex.
```java
String headerName  = entry.getKey();
if (!result.containsKey(headerName)) {
  result.put(headerName.toLowerCase(Locale.ROOT), new LinkedList<String>());
}
result.get(headerName).addAll(entry.getValue());
```

core/src/main/java/feign/Response.java
```diff
@@ -242,4 +246,19 @@ public String toString() {
       return decodeOrDefault(data, UTF_8, "Binary data");
     }
   }
+
+  private static Map<String, Collection<String>> caseInsensitiveCopyOf(Map<String, Collection<String>> headers) {
```
adriancole: tests would be less crazy, if you make this package protected and test it directly.

core/src/test/java/feign/ResponseTest.java
```diff
@@ -32,4 +38,40 @@ public void reasonPhraseIsOptional() {
...
+  @Test
+  public void headerValuesWithSameNameOnlyVaryingInCaseAreMerged() {
+    Map<String, Collection<String>> headersMap = new LinkedHashMap<>();
+    headersMap.put("Set-Cookie", Lists.newArrayList("Cookie-A=Value", "Cookie-B=Value"));
+    headersMap.put("set-cookie", Lists.newArrayList("Cookie-C=Value"));
+
+    Response response = Response.create(200, null, headersMap, new byte[0]);
+
+    List<String> expectedHeaderValue = Lists.newArrayList("Cookie-A=Value", "Cookie-B=Value", "Cookie-C=Value");
+    assertThat(response.headers().get("set-cookie")).isEqualTo(expectedHeaderValue);
```
adriancole: assertj has map assertions!

you should probably be able to do 

```java
assertThat(response.headers()).containsOnly(
  entry("set-cookie", asList("Cookie-A=Value", "Cookie-B=Value", "Cookie-C=Value")
)
```

core/src/test/java/feign/ResponseTest.java
```diff
@@ -32,4 +38,40 @@ public void reasonPhraseIsOptional() {
...
+                                        Collections.singletonMap("Content-Type", valueList),
+                                        new byte[0]);
+    assertThat(response.headers().get("content-type")).isEqualTo(valueList);
+    assertThat(response.headers().get("Content-Type")).isEqualTo(valueList);
+  }
+
+  @Test
+  public void headerValuesWithSameNameOnlyVaryingInCaseAreMerged() {
+    Map<String, Collection<String>> headersMap = new LinkedHashMap<>();
+    headersMap.put("Set-Cookie", Lists.newArrayList("Cookie-A=Value", "Cookie-B=Value"));
```
adriancole: nit: using the built-in `Arrays.asList` for things like this is more consistent with other code and.. less chars :P

adriancole:
the change seems desirable. I made some style comments, but there's nothing concerning.

core/src/main/java/feign/Response.java
a-k-g: I'm thinking of leaving this at is now. Testing it this way seems a bit cumbersome, but it's safer.

a-k-g:
Rebased this PR to resolve merge conflicts on the changelog. Would be nice to have this merged soon.

Thanks again.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   * Previously the OkhttpClient would throw an exception, and ApacheHttpClient
     would report a wrong, possibly negative value
 * Adds support for encoded query parameters in `@QueryMap` via `@QueryMap(encoded = true)`
+* Keys in `Response.headers` are now lower-cased. This map is now case-insensitive with regards to keys,
+  and iterates in lexicographic order.
+  * This is a step towards supporting http2, as header names in http1 are treated as case-insensitive
```
adriancole: even better than my suggestion. nice!

adriancole:
Thanks to you!

PR #423, opened 2016-07-22
Disables fallback when there's no fallback class configured

adriancole:
I'm not sure if this will solve the below issue, but it seems like we
should disable fallbacks when we know we don't have one.

See https://github.com/spring-cloud/spring-cloud-netflix/issues/1204


adriancole:
cc @dsyer @ryanjbaxter here, before and after it doesn't seem like any behavioral difference (just a message). However, there might be some subtleties around usage not in this test case that could be better now?

adriancole:
actually this change totally messes with other tests which should be unrelated!

adriancole:
at the moment I think this change is distracting as Hystrix is programmed to understand the default behavior, which is the throw unsupported when there's no fallback

PR #425, opened 2016-08-02
Introducing pre and post processors

marcingrzejszczak:
with this change you are able to
- add a Pre processor - code that will ALWAYS be applied before sending the request and any request interceptors are executed
- add a Post processor - code that will ALWAYS be applied after receiving the response or when an exception was thrown

that will be useful to solve:
https://github.com/spring-cloud/spring-cloud-sleuth/issues/362 and https://github.com/spring-cloud/spring-cloud-sleuth/issues/202


spencergibb:
At a glance it looks straight forward.

core/src/main/java/feign/RequestPreProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+ * For example: <br>
+ * <pre>
+ * public void apply(RequestTemplate input) {
+ *     // do sth meaningful before the request was built and sent
+ * }
+ * </pre>
+ * <br> <br><b>Configuration</b><br> <br> {@code RequestPreProcessor} are configured via {@link
+ * Feign.Builder#requestPreProcessors}.
+ */
+public interface RequestPreProcessor {
```
adriancole: why are we adding something with the same interface as RequestInterceptor?

core/src/main/java/feign/RequestPostProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+ * Feign.Builder#requestPostProcessors}.
+ */
+public interface RequestPostProcessor {
+
+  /**
+   * Called for every request. Add data using methods on the supplied {@link RequestTemplate}.
+   *
+   * @param e - RetryableException if the request is to be retried. Can be null if there was no exception.
+   */
+  void apply(RequestTemplate template, RetryableException e);
```
adriancole: what's the use case for accessing the request template after a request was invoked when the exception isn't retryable? The name RequestPostProcessor  sounds generalized, but the method looks very special-case

marcingrzejszczak: Yeah I thought that basing on the request data you could different stuff. The thing that comes to my mind ATM is that maybe a better thing would be to pass the `Response` too to have all the information.

So the concept is that especially in Sleuth you need to close the span once you have no exception and continue / detach a span once you tried to retry. (Actually I think I'll have to play around more with this in Sleuth to ensure that this is what we need).

core/src/main/java/feign/RequestPreProcessor.java
marcingrzejszczak: I didn't have a better idea for the name ;) Actually we could make RequestInterceptor be executed earlier and we could remove this thing.

adriancole:
I think the concrete use case of these changes need to be present in unit tests, particularly since this is similar to other interceptors that already exist. Making a wild guess about intent as the issues pasted don't explain this design...

It seems you need to have some context-specific behavior when retries occur. One quick way to address that would be to add `Retryer.Factory.create(template)`, which would be an alternative to `retryer.clone()`. Retryer is stateful anyway, so caching the template would allow the special-casing without affecting the control flow.

ex.
```java
interface Retryer {
  interface Factory {
    Retryer create(RequestTemplate template);
  }

  // default implementation which makes the old Feign.Builder portable.
  class CloneFactory implements Factory {
    Retryer source;

    Retryer create(RequestTemplate ignored) {
      return source.clone();
    }
--snip--
```

marcingrzejszczak:
Yeah maybe you're right. I'll try to put in the test something similar to what I want to achieve in Sleuth. Thanks for the hint

adriancole:
:thumbsup:

marcingrzejszczak:
@adriancole @spencergibb - updated the code with tests / updated javadocs

marcingrzejszczak:
Related PR - https://github.com/spring-cloud/spring-cloud-sleuth/pull/367

adriancole:
fyi still a bit of a zombie, so will review this tomorrow

core/src/main/java/feign/Request.java
```diff
@@ -33,22 +33,37 @@
...
+                               byte[] body, Charset charset, boolean retried) {
+    return new Request(method, url, headers, body, charset, retried);
   }
 
   private final String method;
   private final String url;
   private final Map<String, Collection<String>> headers;
   private final byte[] body;
   private final Charset charset;
+  private final boolean retried;
```
adriancole: was looking at this and also usage in sleuth. Seems a strange place to add state. The template is about how to create a request. The retryer is the place to deal with side-effects about a failed attempt.

adriancole:
looking at your usage in sleuth, it looks like you need to know:
* that a request completed (so that you can make a side-effect)
* if a retry happened ( so that a side-effect doesn't happen twice )

I'm not a fan of this interceptor, but on-board with finding a way to get this to work.

marcingrzejszczak:
We've gone with a different approach that makes this obsolete - https://github.com/spring-cloud/spring-cloud-sleuth/pull/372 .

Issue #431, opened 2016-08-05
Prefer method with Exception argument for fallbacks


tjuchniewicz:
Maybe this is not standard case but it would simplify a lot when we use declarative Feign Clients using Spring Cloud tools. 

Please see idea described here: https://github.com/spring-cloud/spring-cloud-netflix/issues/1117

adriancole:
can you please be specific about what change you want here? it is less effortful than having to look at other repos and makes your issue more relevant to this codebase.

tjuchniewicz:
Now when executing fallback, Feign expects exactly the same method signature as a target API. Proposal is that it should support (prefer if exists) the same method signature plus Exception to easily get exception information in fallback method.

adriancole:
So concretely, if the method on the api interface was get() the fallback
interface might have get(Throwable cause) or similarly post(Foo foo) ->
post(Foo foo, Throwable cause)

Is that right?

tjuchniewicz:
Right.

ryanjbaxter:
I like this idea.  @tjuchniewicz why not create a pull request?

spencergibb:
We knew going in that our integration was going to be basic to start.

adriancole:
https://github.com/OpenFeign/feign/pull/443

PR #432, opened 2016-08-08
Prefer method with Throwable argument for fallbacks if such method exists

tjuchniewicz:
Fixes gh-431

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -24,7 +24,10 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import feign.InvocationHandlerFactory;
@@ -110,7 +113,31 @@ final class HystrixInvocationHandler implements InvocationHandler {
           return super.getFallback();
         }
         try {
-          Object result = fallbackMethodMap.get(method).invoke(fallback, args);
+          Method fallbackMethod = fallbackMethodMap.get(method);
+          Method enhancedFallbackMethod = null;
+          Class<?>[] getEnhancedArgsClasses = getEnhancedArgsClasses(args);
+          try {
+            enhancedFallbackMethod = fallback.getClass().getMethod(fallbackMethod.getName(), getEnhancedArgsClasses);
+          } catch (NoSuchMethodException e) {
+            // no such method
+          } catch (SecurityException e) {
+            throw new AssertionError(e);
+          }
+          
+          Object result;
+          if (enhancedFallbackMethod != null) {
+            Object[] argsPlusException;
+            if (args == null) {
+              argsPlusException = new Object[] { getFailedExecutionException() };
+            } else {
+              argsPlusException = Arrays.copyOf(args, args.length + 1);
+              argsPlusException[argsPlusException.length - 1] = getFailedExecutionException();
+            }
+            result = enhancedFallbackMethod.invoke(fallback, argsPlusException);
+          } else {
+            result = fallbackMethod.invoke(fallback, args);
+          }
+          
           if (isReturnsHystrixCommand(method)) {
             return ((HystrixCommand) result).execute();
           } else if (isReturnsObservable(method)) {
@@ -147,6 +174,18 @@ final class HystrixInvocationHandler implements InvocationHandler {
       return hystrixCommand.toObservable().toCompletable();
     }
     return hystrixCommand.execute();
+  }
+  
+  protected Class<?>[] getEnhancedArgsClasses(Object[] args) {  
+    List<Class<?>> classes = new ArrayList<Class<?>>();
+     
+    if(args != null) {
+      for (int i = 0; i < args.length; i++) {
+        classes.add(args[i].getClass());
+      }
+    }
+    classes.add(Throwable.class);
+    return classes.toArray(new Class<?>[]{});
   }
 
   private boolean isReturnsCompletable(Method method) {
--- a/hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
+++ b/hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
@@ -109,6 +109,30 @@ public class HystrixBuilderTest {
     assertThat(command).isNotNull();
     assertThat(command.execute()).containsExactly("fallback");
   }
+  
+  @Test
+  public void hystrixEnhancedFallback() {
+    server.enqueue(new MockResponse().setResponseCode(500));
+
+    TestInterface api = target();
+
+    HystrixCommand<String> command = api.enhanced();
+
+    assertThat(command).isNotNull();
+    assertThat(command.execute()).isEqualTo("fallback with exception");
+  }
+  
+  @Test
+  public void hystrixEnhancedWithArgFallback() {
+	server.enqueue(new MockResponse().setResponseCode(500));
+	  
+	TestInterface api = target();
+	  
+	HystrixCommand<String> command = api.enhancedWithArg("sample text");
+	  
+	assertThat(command).isNotNull();
+	assertThat(command.execute()).isEqualTo("fallback with exception");
+  }
 
   // When dealing with fallbacks, it is less tedious to keep interfaces small.
   interface GitHub {
@@ -592,6 +616,14 @@ public class HystrixBuilderTest {
     @RequestLine("GET /")
     @Headers("Accept: application/json")
     HystrixCommand<String> command();
+    
+    @RequestLine("GET /")
+    @Headers("Accept: application/json")
+    HystrixCommand<String> enhanced();
+    
+    @RequestLine("GET /{strParam}")
+    @Headers("Accept: application/json")
+    HystrixCommand<String> enhancedWithArg(@Param("strParam") String strParam);
 
     @RequestLine("GET /")
     @Headers("Accept: application/json")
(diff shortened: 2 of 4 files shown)


spencergibb:
How about a test?

tjuchniewicz:
I will add tests.

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -109,6 +109,30 @@ public void hystrixCommandListFallback() {
...
+
+    HystrixCommand<String> command = api.enhancedWithArg("sample text");
+
+    assertThat(command).isNotNull();
+    assertThat(command.execute()).isEqualTo("fallback with exception");
+  }
+  
+  @Test
+  public void hystrixEnhancedWithArgFallback() {
+	  server.enqueue(new MockResponse().setResponseCode(500));
```
spencergibb: formatting

spencergibb:
There are a few formatting issues, should also update the docs.

spencergibb:
Almost there. There are still some tabs, where feign uses two spaces.

tjuchniewicz:
Please take a look now.

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -110,7 +113,31 @@ protected Object getFallback() {
           return super.getFallback();
         }
         try {
-          Object result = fallbackMethodMap.get(method).invoke(fallback, args);
+          Method fallbackMethod = fallbackMethodMap.get(method);
+          Method enhancedFallbackMethod = null;
```
adriancole: the word enhanced doesn't help explain anything. how about something like `fallbackMethodWithThrowable`?

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -109,6 +109,30 @@ public void hystrixCommandListFallback() {
...
+    TestInterface api = target();
+
+    HystrixCommand<String> command = api.enhanced();
+
+    assertThat(command).isNotNull();
+    assertThat(command.execute()).isEqualTo("fallback with exception");
+  }
+  
+  @Test
+  public void hystrixEnhancedWithArgFallback() {
```
adriancole: let's be consistent. please scrub the word enhanced everywhere as it is a filler term. Choose "WithArgFallback" or something else, and be consistent in both the names of locals and also in test classes and method.

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -592,6 +616,14 @@ private TestInterface targetWithoutFallback() {
     @RequestLine("GET /")
     @Headers("Accept: application/json")
     HystrixCommand<String> command();
+    
+    @RequestLine("GET /")
+    @Headers("Accept: application/json")
+    HystrixCommand<String> enhanced();
```
adriancole: no need to duplicate this method, just use `command()`

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -592,6 +616,14 @@ private TestInterface targetWithoutFallback() {
...
+    
+    @RequestLine("GET /{strParam}")
+    @Headers("Accept: application/json")
+    HystrixCommand<String> enhancedWithArg(@Param("strParam") String strParam);
```
adriancole: no need to rename the method when overloading.. just call it `command()`

Having fewer changes draws attention to what's actually changed. For example, the method "command()" refers to the return type which is a hystrix "command" as opposed to a completable.

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -717,5 +749,33 @@ public String get() {
     public Completable completable() {
       return Completable.complete();
     }
+
+    @Override
+    public HystrixCommand<String> enhanced() {
```
adriancole: I'd define a new fallback object that says what you are doing..

```bash
class FallbackPrefersThrowableArg {
  @Override
  public HystrixCommand<String> command() {
    throw new AssertionError("command(throwable) should have been called");
  }

  @Override
  public HystrixCommand<String> command(Throwable throwable) {
    return new ...
--snip--
```

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -110,7 +113,31 @@ protected Object getFallback() {
...
+          Class<?>[] getEnhancedArgsClasses = getEnhancedArgsClasses(args);
```
adriancole: lots of late bugs possible here. I'd move this to where the dispatch table is being created. 

https://github.com/OpenFeign/feign/blob/6a6271727ad14dd776e3c3fb029f15fdf95c9c4a/hystrix/src/main/java/feign/hystrix/HystrixFeign.java#L44

validation will be a pain, as people can easily name things subltely wrong etc, with wrong visibility, maybe Exception arg instead of Throwable. They'd silently fail and fill our issues list stack overflow etc with questions. See if you can tighten this up, for example, failing fast the way that contract does. These sort of problems are possible when an interface isn't being implemented.

I'll discuss an alternative as a summary

adriancole:
Whatever we do increases the complexity of the implementation, particularly when an interface isn't being implemented. A lot of the simplicity of feign is that we can lean on Java's type system for its strengths. This loose association is something I've done before in other libraries and led to many little bugs.

Looking at the implementation, I'd really prefer a factory pattern instead. Something that takes the strengths we already have (like interfaces) can reduce the type of bugs we get. For example, something like..

```java
interface FallbackFactory<T> {
   T create(Throwable cause);

   final class Default<T> implements FallbackFactory<T> {
      final T constant;

      Default(T constant) {
        this.constant = constant;
      }

      @Override
      public T create(Throwable ignored) {
         return constant;
      }
   }
}
```

When someone wants to have context of a throwable, they'd implement `T` with a constructor parameter of the cause. They'd make a `FallbackFactory<T>` whose create method calls that constructor.

Thoughts?

spencergibb:
How would you use that in practice?

adriancole:
@spencergibb are you asking what would we do with the throwable in practice? or how is writing a fallback object different, when it has a throwable field?

adriancole:
Here's an example derived from the other.

```java
  interface GitHub {
    List<String> contributors(@Param("owner") String owner, @Param("repo") String repo);
  }

  FallbackFactory<GitHub2> fallbackFactory = throwable -> (owner, repo) -> {
    if (owner.equals("Netflix") && repo.equals("feign")) {
      return Arrays.asList("stuarthendren"); // inspired this approach!
    } else if (throwable instanceof IllegalArgumentException) {
      return Arrays.asList("cops");
    } else {
    }
  };
```

adriancole:
and to configure, we'd overload the HystrixFeign so that we can pass a factory instead of a single fallback instance.. like this..

```java
    public <T> T target(Target<T> target, final FallbackFactory<T> fallbackFactory) {
```

We could dispatch the old signature to the old one with by delegating with FallbackFactory.Default like this...

```java
    public <T> T target(Target<T> target, final T fallback) {
      return target(target, new FallbackFactory.Default<T>(fallback));
    }
```

adriancole:
sorry about the ramble, but the user-facing change is that advanced users (those concerned with the source exception), can use the factory pattern mentioned with no api impact to the existing users who don't need to know the cause.

spencergibb:
I guess I still don't see how that matches up to individual methods (I may be tired and incoherent though). 😜

adriancole:
so since the throwable is a field, all methods can see it and access it.

I don't really know the use case of the throwable as all the related issues are vague. People saying "I need this" but no concrete story around it.

For example, I'd love to hear something like.. I get X exception in my service. Knowing that, I can choose to respond X instead of Y. If we had a realistic use case, I could make the example more concrete.

spencergibb:
>  I get X exception in my service. Knowing that, I can choose to respond Y instead of Z. If we had a realistic use case, I could make the example more concrete.

That's exactly right.

ryanjbaxter:
Thought I would leave my two cents on a use case here....

I ran into a problem with Feign + Hystrix where it was actually Hystrix throwing an exception because there was no fallback defined and not actually a problem with my client per se.  Just logging the exception would have helped me immensely in knowing the problem :)

adriancole:
regardless of this change, we could certainly log the exception!

assuming we go forward with this change, we can also log the exception in
the example (or in the default FallbackFactory)

ryanjbaxter:
Sure, just throwing that out there :)

tjuchniewicz:
Logging is one the reason but not strong enough to implement this feature. I agree wtih @adriancole  that stronger argument is needed. 

I'm trying to get this information from my teammates...

I agree that loose association may result in bugs but my approach I believe is more user friendly for Feign declarative usage using `@FeignClient` from Spring Cloud Netflix. Especially when we have more than one method in the interface. But looks good if this feature is used occasionally.

adriancole:
Feign's design, including fallbacks already prefers smaller interfaces. ex the below is literally in the javadoc of the hystrix feign builder

"     * // When dealing with fallbacks, it is less tedious to keep interfaces small."

The advice to use the compiler instead of more reflective dispatch is congruent with Feign's design which generally favors less runtime magic. For example, the whole "contract" thing of parsing things up front, and the RequestTemplate trying to keep a sane an unreflective view of the template state for a method were examples of earnest attempts to reduce the amount of late errors.

As time went by there have been features, especially around partial matches etc, that have definitely made some of feign high maintenance and buggy. It is already the case I'm barely interested in fixing these sorts of bugs, and I really don't want to be completely turned off of Feign. Long story short, I really can't get behind adding magic methods by convention. I'll help with getting the feature in, but not this way.

adriancole:
https://github.com/OpenFeign/feign/pull/443

PR #433, opened 2016-08-08
add context-path support in LoadBalancingTarget

mikexliu:


mikexliu:
I can make the commit message a bit better.
In:
`Target.java`

We have:
```java
/* base HTTP URL of the target. For example, {@code https://api/v2}. */
```

It seems like LoadBalancingTarget should support the "api/v2" (or any other path) as well. But it's stripped out in the constructor.

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
```diff
@@ -45,26 +45,38 @@
...
     this.scheme = checkNotNull(scheme, "scheme");
     this.name = checkNotNull(name, "name");
+    this.path = "";
+    this.lb = AbstractLoadBalancer.class.cast(getNamedLoadBalancer(name()));
+  }
+  
+  protected LoadBalancingTarget(Class<T> type, String scheme, String name, String path) {
+    this.type = checkNotNull(type, "type");
+    this.scheme = checkNotNull(scheme, "scheme");
+    this.name = checkNotNull(name, "name");
```
mikexliu: I didn't want to change existing implementation, but it seems like `url()` should actually return scheme + path, not name (according to `Target.java` docs)

mikexliu:
@adriancole Can you review this? Seems to be a simple missing feature.

ribbon/src/test/java/feign/ribbon/LoadBalancingTargetTest.java
```diff
@@ -71,10 +71,43 @@ public void loadBalancingDefaultPolicyRoundRobin() throws IOException, Interrupt
...
+    try {
+      LoadBalancingTarget<TestInterface>
+              target =
+              LoadBalancingTarget.create(TestInterface.class, "http://" + name + "/servers");
+      TestInterface api = Feign.builder().target(target);
+    
+      api.get();
+      api.get();
+    
+      assertEquals(1, server1.getRequestCount());
```
adriancole: I think you should do an assertion on the path received by the server. ex server1.takeRequest().getPath()

ribbon/src/test/java/feign/ribbon/LoadBalancingTargetTest.java
```diff
@@ -71,10 +71,43 @@ public void loadBalancingDefaultPolicyRoundRobin() throws IOException, Interrupt
...
+      assertEquals(1, server2.getRequestCount());
+      // TODO: verify ribbon stats match
```
adriancole: no need to copy/paste TODOs

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
```diff
@@ -45,26 +45,38 @@
 
   private final String name;
   private final String scheme;
+  private final String path;
   private final Class<T> type;
   private final AbstractLoadBalancer lb;
+  
+  @Deprecated
```
adriancole: add javadoc saying when this will be removed.

ex.

```
/**
 * @deprecated will be removed in Feign 10
 */
```

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
```diff
@@ -45,26 +45,38 @@
...
   }
 
   /**
-   * creates a target which dynamically derives urls from a {@link com.netflix.loadbalancer.ILoadBalancer
+   * Creates a target which dynamically derives urls from a {@link com.netflix.loadbalancer.ILoadBalancer
    * loadbalancer}.
    *
    * @param type       corresponds to {@link feign.Target#type()}
-   * @param schemeName naming convention is {@code https://name} or {@code http://name} where name
+   * @param scheme     naming convention is {@code https://name/api/v2} or {@code http://name/context} where name
```
adriancole: schemeNamePath - this literally is saying what's encoded

adriancole: or just call it url (since we parse it into those things..). At any rate, keep the base example in the javadoc (ie without a path) along with the other ones.

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
adriancole: feel free to fix that. technically it will still be wrong as it is an unresolved url, but you're right to say it will be better.

adriancole:
Made some notes.. please update CHANGES, too. and should be good for the next version of feign!

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
```diff
@@ -79,7 +94,7 @@ public String name() {
 
   @Override
   public String url() {
-    return name;
+    return scheme + path;
```
mikexliu: @adriancole 
Thanks for the comments! I changed this to be `scheme + path` but and all the tests pass but not sure if it's okay to change existing functionality (even if it is "technically" more correct).

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ### Version 9.1
 * Allows query parameters to match on a substring. Ex `q=body:{body}`
+* Support for full path in `LoadBalancingTarget`.
```
adriancole: sorry 9.1 is already out.. can you move to 9.2?

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
adriancole: sounds fine to me

CHANGELOG.md
adriancole: ps it is best to keep details to a minimum in the change log. Details are good for your actual commit message.

Ex. for the changelog:

* Supports context path when using Ribbon `LoadBalancingTarget`

For the commit message:
```
Supports context path when using Ribbon LoadBalancingTarget

Before, `LoadBalancingTarget` stripped out the path and only used
`URI.getScheme()` and `URI.getHost()` to generate the `Request`.

Now, it will add `URI.getPath()` to the `Request` as well; this is
useful if you want to interact with endpoints with a context-path.
```

mikexliu:
Don't merge yet.

ribbon/src/main/java/feign/ribbon/LoadBalancingTarget.java
```diff
@@ -79,7 +94,7 @@ public String name() {
 
   @Override
   public String url() {
-    return name;
+    return String.format("%s://%s", scheme, path);
```
mikexliu: Just found in my own testing this isn't just scheme + path, needed the "://". :P

Thanks!

mikexliu:
@adriancole 
I believe this is ready now. Sorry for the last minute fix!

adriancole:
Beautiful

PR #435, opened 2016-08-09
Makes request info available to logResponseAndRebuffer method

nfiacco:
Added the 'Request request' parameter to the abstract logResponseAndRebuffer method.
* Allows developer to check response code before logging HTTP requests

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -75,8 +75,8 @@ public abstract class Logger {
     log(configKey, "---> RETRYING");
   }
 
-  protected Response logAndRebufferResponse(String configKey, Level logLevel, Response response,
-                                            long elapsedTime) throws IOException {
+  protected Response logAndRebufferResponse(String configKey, Level logLevel, Request request,
+                                            Response response, long elapsedTime) throws IOException {
     String reason = response.reason() != null && logLevel.compareTo(Level.NONE) > 0 ?
         " " + response.reason() : "";
     log(configKey, "<--- HTTP/1.1 %s%s (%sms)", response.status(), reason, elapsedTime);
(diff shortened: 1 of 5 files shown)


nfiacco:
Use case: only logging on error message, and being able to include request info when doing so

core/src/main/java/feign/Logger.java
```diff
@@ -75,8 +75,8 @@ void logRetry(String configKey, Level logLevel) {
     log(configKey, "---> RETRYING");
   }
 
-  protected Response logAndRebufferResponse(String configKey, Level logLevel, Response response,
-                                            long elapsedTime) throws IOException {
+  protected Response logAndRebufferResponse(String configKey, Level logLevel, Request request,
```
adriancole: this is an api breaking change

adriancole:
hey there.. we can't break api in a minor release, as it would pin people to versions before this change...

It is possible that we could add the request to the response, similar to how it is done in okhttp
https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Response.java#L84

however, we'd have to make it nullable until version 10 because again we can't have things breaking (ex custom clients wouldn't know to set this)

What part of the request are you logging, by the way?

nfiacco:
I was hoping to be able to access the url, form data, http method, etc.

I do like the idea of adding this stuff to the Response object. Do you mean that people might see the new field and think it's already set?

adriancole:
ahh I think we can solve this!

Firstly, it is high-time Response had a builder. If you don't mind
helping with this in a separate PR, it will surely clean up a lot,
including the deafening number of factory methods on response. Please
use https://github.com/openzipkin/zipkin/blob/b6da02686249ac2df81daf98315f80c572142e29/zipkin/src/main/java/zipkin/Annotation.java
as an example of a clean builder that includes a toBuilder() function.
Change all call-sites that use Response factories to use the builder,
and deprecate the factories for removal in version 10. For example,
change Logger.logAndRebufferResponse to return
response.toBuilder().body(bodyData).build();

Next, add Response.request, with similar javadoc as the one in OkHttp.
Note in a comment that we will add a null check in version 10, which
may require changes to custom feign.Client or Loggers.  Make sure that
all feign's packaged clients set request (ideally in a unit test).

Finally, lock down the two places where we take a response from
potentially custom code:

response = client.execute(request, options);

response =
    logger.logAndRebufferResponse(metadata.configKey(), logLevel,
response, elapsedTime);


below each, add something like this:

// ensure the request is set. TODO: remove in feign 10
response = response.toBuilder().request(request).build();



Make sense?

nfiacco:
Ah yes I'm a big fan of the builder paradigm-- and now that you point it out there really is an absurd number of Response.create methods. I'm definitely happy to help, especially if it means I get to use it!

adriancole:
on deprecated I mean both the annotation and the javadoc ref. in the javadoc ref you can say that this will be removed in version 10

nfiacco:
Okay about to make an initial pr, still some stuff we should iterate on

PR #439, opened 2016-08-15
Reduce logging overhead

schlosna:
* Avoid overhead when logging is disabled.
* Reuse a single StringBuilder when constructing method tag.

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -178,7 +178,9 @@ public abstract class Logger {
 
     @Override
     protected void log(String configKey, String format, Object... args) {
-      logger.fine(String.format(methodTag(configKey) + format, args));
+      if (logger.isLoggable(java.util.logging.Level.FINE)) {
+        logger.fine(String.format(methodTag(configKey) + format, args));
+      }
     }
 
     /**
--- a/slf4j/src/main/java/feign/slf4j/Slf4jLogger.java
+++ b/slf4j/src/main/java/feign/slf4j/Slf4jLogger.java
@@ -68,6 +68,8 @@ public class Slf4jLogger extends feign.Logger {
   protected void log(String configKey, String format, Object... args) {
     // Not using SLF4J's support for parameterized messages (even though it would be more efficient) because it would
     // require the incoming message formats to be SLF4J-specific.
-    logger.debug(String.format(methodTag(configKey) + format, args));
+    if (logger.isDebugEnabled()) {
+      logger.debug(String.format(methodTag(configKey) + format, args));
+    }
   }
 }


adriancole:
no issue with guarding on debug, fine, but I'd benchmark before doing the
change to stringbuilder instead of concat. sometimes javac does this stuff
automatically the change to expose the stringbuilder makes the format
string code harder to understand.

http://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

schlosna:
I have a separate local branch I'll push and create a PR later tonight or tomorrow to hook up the benchmarks module and add a small JMH microbenchmark for the StringBuilder, but the main reason I added it was that I saw a bunch of string copies in some profiles since we almost always append an additional format string after generating the method tag format, and sometimes the new line placeholder. There is still another string copy for the String.format, though as already commented in the code we're not doing printf to slf4j message conversion (yet).

adriancole:
Since I asked for the benchmark, I'll share in the pain :) Here's a succinct way to compare (I think). Feel free to use or toss.

```java
@Benchmark
public void logBasic_concat() {
  logBasic(concatLogger);
}

@Benchmark
public void logBasic_builder() {
  logBasic(builderLogger);
}

static void logBasic(Logger logger) {
  logger.log("Github#contributors", "---> %s %s HTTP/1.1", "GET", "https://api.github.com/repos/openfeign/feign");
}

interface Logger {
  String toLog(String configKey, String format, Object... args);
}

static final Logger concatLogger = (configKey, format, args) -> {
 String methodTag = new StringBuilder().append('[').append(configKey.substring(0, configKey.indexOf('('))).append("] ").toString();
 return String.format(methodTag + format, args);
}

static final Logger builderLogger = (configKey, format, args) -> {
 StringBuilder methodTag = new StringBuilder().append('[').append(configKey.substring(0, configKey.indexOf('('))).append("] ").toString();
 return String.format(methodTag.append(format).toString(), args);
}
```

schlosna:
This PR is now just adding guard around logging.

Ran the JMH benchmark https://github.com/schlosna/feign/blob/feature/method-tag-benchmark/benchmark/src/main/java/feign/benchmark/MethodTagBenchmarks.java which shows the existing implementation as faster, so I've reverted that part of this change.

```
Benchmark                              Mode  Cnt       Score       Error  Units
MethodTagBenchmarks.logBasic_builder  thrpt   15  954826.745 ± 68500.736  ops/s
MethodTagBenchmarks.logBasic_concat   thrpt   15  979587.330 ± 48137.135  ops/s
```

adriancole:
JVM's a mysterious thing, right? :) thanks for the help and the diligence!

schlosna:
Thanks for merging. The results were a bit surprising, but good.

There have been a few discussions over the years about moving more of these optimizations into both javac and the JVM itself, see https://bugs.openjdk.java.net/browse/JDK-4059189 and http://markmail.org/message/mrhy6ystcirswhqf , and [Java 9's indy string concatenation](http://openjdk.java.net/jeps/280) will likely change the results as well.

PR #443, opened 2016-08-16
Adds FallbackFactory, allowing access to the cause of a Hystrix fallback

### LINKED ISSUE #431: Prefer method with Exception argument for fallbacks

adriancole:
The cause of the fallback is now logged by default to FINE level. You can programmatically inspect
the cause by making your own `FallbackFactory`. In many cases, the cause will be a `FeignException`,
which includes the http status.

Here's an example of using `FallbackFactory`:

```java
// This instance will be invoked if there are errors of any kind.
FallbackFactory<GitHub> fallbackFactory = cause -> (owner, repo) -> {
  if (cause instanceof FeignException && ((FeignException) cause).status() == 403) {
  } else {
    return Arrays.asList("yogi");
  }
};

GitHub github = HystrixFeign.builder()
                            ...
                            .target(GitHub.class, "https://api.github.com", fallbackFactory);
```

Fixes #431
See #432

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -38,15 +38,15 @@ final class HystrixInvocationHandler implements InvocationHandler {
 
   private final Target<?> target;
   private final Map<Method, MethodHandler> dispatch;
-  private final Object fallback; // Nullable
+  private final FallbackFactory<?> fallbackFactory; // Nullable
   private final Map<Method, Method> fallbackMethodMap;
   private final Map<Method, Setter> setterMethodMap;
 
   HystrixInvocationHandler(Target<?> target, Map<Method, MethodHandler> dispatch,
-                           SetterFactory setterFactory, Object fallback) {
+                           SetterFactory setterFactory, FallbackFactory<?> fallbackFactory) {
     this.target = checkNotNull(target, "target");
     this.dispatch = checkNotNull(dispatch, "dispatch");
-    this.fallback = fallback;
+    this.fallbackFactory = fallbackFactory;
     this.fallbackMethodMap = toFallbackMethod(dispatch);
     this.setterMethodMap = toSetters(setterFactory, target, dispatch.keySet());
   }
@@ -115,10 +115,11 @@ final class HystrixInvocationHandler implements InvocationHandler {
 
       @Override
       protected Object getFallback() {
-        if (fallback == null) {
+        if (fallbackFactory == null) {
           return super.getFallback();
         }
         try {
+          Object fallback = fallbackFactory.create(getFailedExecutionException());
           Object result = fallbackMethodMap.get(method).invoke(fallback, args);
           if (isReturnsHystrixCommand(method)) {
             return ((HystrixCommand) result).execute();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 9.3
+* Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
+
 ### Version 9.2
 * Adds Hystrix `SetterFactory` to customize group and command keys
 * Supports context path when using Ribbon `LoadBalancingTarget`
--- a/hystrix/README.md
+++ b/hystrix/README.md
@@ -106,3 +106,26 @@ GitHub github = HystrixFeign.builder()
                             ...
                             .target(GitHub.class, "https://api.github.com", fallback);
 ```
+
+#### Considering the cause
+
+The cause of the fallback is logged by default to FINE level. You can programmatically inspect
+the cause by making your own `FallbackFactory`. In many cases, the cause will be a `FeignException`,
+which includes the http status.
+
+Here's an example of using `FallbackFactory`:
+
+```java
+// This instance will be invoked if there are errors of any kind.
+FallbackFactory<GitHub> fallbackFactory = cause -> (owner, repo) -> {
+  if (cause instanceof FeignException && ((FeignException) cause).status() == 403) {
+    return Collections.emptyList();
+  } else {
+    return Arrays.asList("yogi");
+  }
+};
+
+GitHub github = HystrixFeign.builder()
+                            ...
+                            .target(GitHub.class, "https://api.github.com", fallbackFactory);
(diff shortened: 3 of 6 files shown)


adriancole:
cc @tjuchniewicz @spencergibb @ryanjbaxter @marcingrzejszczak

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -106,10 +104,11 @@ protected Object run() throws Exception {
 
       @Override
       protected Object getFallback() {
-        if (fallback == null) {
+        if (fallbackFactory == null) {
           return super.getFallback();
         }
         try {
+          Object fallback = fallbackFactory.create(getFailedExecutionException());
```
adriancole: note: if all we want to do is log the cause, we can delete the entire pull request and just log here!

```java
Throwable cause = getFailedExecutionException();
 if (logger.isLoggable(Level.FINE)) {
  logger.log(Level.FINE, "fallback due to: " + cause.getMessage(), cause);
}
```

spencergibb: No, I think folks want access to the actual exception object.

adriancole: okie dokie, then we don't need to delete this branch :)

adriancole:
PS for those who have existing fallback classes, here's the process for teaching them about the cause..

1. make a Throwable constructor parameter in your fallback class, and assign it to a field
2. make a FallbackFactory which invokes that constructor
3. For any of your methods that care about the cause, read the field you just added.

ryanjbaxter:
At first i was slightly confused by the name `FallbackFactory` but after thinking about it, I guess it makes sense since we are producing a dynamic fallback of sorts.  I think this looks good.

adriancole:
added a bunch of example tests, which aren't really about feign, more about a million ways to construct objects. One in particular could be helpful for those who want to retrofit an existing class and not declare a separate FallbackFactory. Note: if feign was java8 we wouldn't even need FallbackFactory!

```
  // retrofit so people don't have to track 2 classes
  static class FallbackApiRetro implements Api, FallbackFactory<FallbackApiRetro> {

    @Override public FallbackApiRetro create(Throwable cause) {
      return new FallbackApiRetro(cause);
    }

    final Throwable cause; // nullable

    public FallbackApiRetro() {
      this(null);
    }

    FallbackApiRetro(Throwable cause) {
      this.cause = cause;
    }

    // actual api methods can lazy implement any use of the cause.
    @Override public String invoke() {
      return cause != null ? cause.getMessage() : "foo";
    }
  }
```

adriancole:
ps if we aren't quite ready to land this PR, we should probably go ahead and do a release. If that's the case, I can add a logging line as an interim step. Lemme know.

bitsofinfo:
+1 I definitely have a need for this, right now when a fallback is invoked I have no idea about the underlying reason

bpicode:
+1 this feature is a must

adriancole:
OK will merge on green then cut feign 9.3

seetharamani:
Hi is this Feign FallbackFactory is working? I am using version 9.3.1. I am still not able to get the cause from fallback. I used the code example from retrofit test case.

spencergibb:
@seetharamani there are tests that demonstrate it working. If you think you have found a bug, open an issue with steps to reproduce or a sample project that demonstrates the problem. Asking if it works on the merge PR isn't real helpful.

seetharamani:
@spencergibb thank you for the response. Sure I can provide what i have tried. May be I am doing it in a wrong way. Our app is in java 1.6 right now. We are working on upgrading it to 8. But i would need this client call before that.

Here is my client interface.
`public interface MyServiceFeignClient {
	@RequestLine("POST /myService/check/order")
	@Headers("Content-Type: application/vnd.com.domain.dept.order+json;version=1.0")
	MyServiceResponse sendOrder(String content);
}`

Below is my Factory implementation where i actually instantiate the client.

`public class FeignMyServiceClientFactory {
	private static final Logger LOG = LoggerFactory.getLogger(FeignMyServiceClientFactory.class);
	private String serviceUrl;
	public FeignMyServiceClientFactory(final String serviceUrl) {
		this.serviceUrl = serviceUrl;
	}
	public MyServiceFeignClient newInstance() {
		return HystrixFeign.builder()
			.decoder(new GsonDecoder())
			.target(MyServiceFeignClient.class, serviceUrl, new MyServiceFallbackFactory());
	}
**// Below is my fallback implementation. Its inside the factory class**
	static class MyServiceFallbackFactory implements MyServiceFeignClient, FallbackFactory<MyServiceFallbackFactory> {
		final Throwable cause;
		public MyServiceFallbackFactory() {
			this(null);
		}
		MyServiceFallbackFactory(Throwable cause) {
			this.cause = cause;
		}
		@Override
		public MyServiceFallbackFactory create(Throwable cause) {
			if (cause != null) {
				String errMessage = StringUtils.isNotBlank(cause.getMessage()) ? cause.getMessage() : "unknown error occured";
**//I dont see this LOG statement at all. This method never gets called.**
				LOG.debug("service fallback called for the cause : {}", errMessage);
			}
			return new MyServiceFallbackFactory(cause);
		}

		@Override
		public MyServiceResponse sendOrder(String content) {
			LOG.debug("Payment risk service client api fallback called");
			MyServiceResponse response = new MyServiceResponse();
			String errMessage = (cause == null ? "service client api fallback called" : cause.getMessage());
**// here I am expecting to see the actual cause. But  cause is always null.**
			response.setErrorMessage(errMessage);
			response.setResultStatus("WARN";
			return response;
		}

	}
}`

Is the fallback implementation looks good? Please let me know if I am doing anything wrong here. Kindly bare any dumb mistakes as i am new to Feign as well as Hystrix.

spencergibb:
@seetharamani 

>  open an issue with steps to reproduce or a sample project that demonstrates the problem

seetharamani:
Opened the issue with sample project
https://github.com/OpenFeign/feign/issues/458

PR #444, opened 2016-08-16
Make Hystrix Command keys more unique

### LINKED ISSUE #434: Make Hystrix Command names more unique
By default Feign will create Hystrix Commands that use the method name as the key.  It is possible to have two Feign clients with the same method names that are completely separate Hystrix Commands.  However since the the method names are the same the underlying Hystrix Command cache will see these both as the same command.  We can do a little better at our default key naming convention by including the class name in the key we use as well.

This should be a simple change in the code here
https://github.com/OpenFeign/feign/blob/master/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java#L90

For more information see the discussion in [this issue](https://github.com/Netflix/Hystrix/issues/1293) in the Hystrix project.

ryanjbaxter:
Uses fallback method signature as the Hystrix command key.  Fixes #434.

### DIFF
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -87,7 +87,7 @@ final class HystrixInvocationHandler implements InvocationHandler {
     }
 
     String groupKey = this.target.name();
-    String commandKey = "feign#" + method.toString();
+    String commandKey = generateCommandKey(method);
     HystrixCommand.Setter setter = HystrixCommand.Setter
         .withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
         .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
@@ -153,13 +153,11 @@ final class HystrixInvocationHandler implements InvocationHandler {
   /**
    * Generates the key for the Hystrix command.  This key is used by
    * Hystrix in a cache, so we try and make the key as unique as possible.
-   * The default implementation uses the result of calling 
-   * {@link java.lang.Method.html#toString() toString} on the fallback Method.
    * @param method The fallback method.
    * @return A unique key to be used to cache the Hystrix command.
    */
-  protected String generateCommandKey(Method method) {
-	  return method.toString();
+  private String generateCommandKey(Method method) {
+	  return "feign#" + method.toString();
   }
 
   private boolean isReturnsCompletable(Method method) {
--- a/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
+++ b/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
@@ -87,7 +87,7 @@ final class HystrixInvocationHandler implements InvocationHandler {
     }
 
     String groupKey = this.target.name();
-    String commandKey = method.getName();
+    String commandKey = "feign#" + method.toString();
     HystrixCommand.Setter setter = HystrixCommand.Setter
         .withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
         .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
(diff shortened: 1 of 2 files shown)


hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
```diff
@@ -148,6 +149,18 @@ protected Object getFallback() {
...
+  
+  /**
+   * Generates the key for the Hystrix command.  This key is used by
+   * Hystrix in a cache, so we try and make the key as unique as possible.
+   * The default implementation uses the result of calling 
+   * {@link java.lang.Method.html#toString() toString} on the fallback Method.
+   * @param method The fallback method.
+   * @return A unique key to be used to cache the Hystrix command.
+   */
+  protected String generateCommandKey(Method method) {
```
adriancole: type is final so no need to talk about the difference between default impl and the actual impl (nor make it protected). nit.. uhh.. I think you forgot to call this method :)

ps LOVE the first two sentences of the docs..  the rest aren't needed as this will only be viewed by authors.

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
```diff
@@ -148,7 +148,7 @@ public void fallbacksApplyOnError() {
   @Test
   public void errorInFallbackHasExpectedBehavior() {
     thrown.expect(HystrixRuntimeException.class);
-    thrown.expectMessage("contributors failed and fallback failed.");
+    thrown.expectMessage("feign#public abstract java.util.List feign.hystrix.HystrixBuilderTest$GitHub.contributors(java.lang.String,java.lang.String) failed and fallback failed.");
```
adriancole: Feign.configKey key is prettier :P I suspect the command key ends up in the UI somewhere, right? Might want to check that this doesn't make the operator experience ugly

hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java
ryanjbaxter: ha!  and I also forgot to add `feign#` to the beginning of it.

hystrix/src/test/java/feign/hystrix/HystrixBuilderTest.java
ryanjbaxter: You mean in the hystrix dashboard?

spencergibb: Yeah, they do.

ryanjbaxter: So it is not ideal, but in my opinion this is a separate issue from the one we are trying to solve 
image

adriancole: general question is why we'd use something uglier and without any semantic
gain in uniqueness, and also not the convention of feign, when we have a
choice not to?

ryanjbaxter: Guess I am not sure what you mean by without any gain in uniqueness?  Right now the key is the method name now the key is the entire signature which makes it much more unique.  

What is the convention you are referring to?

adriancole: I think in related issues I've mentioned Feign.configKey is the
appropriate tool for the job.

The cheapest way is to replace your method.toString with:

Feign.configKey(method.getDeclaringClass(), method);

Later, I can help extract the Setter thing as I mentioned earlier, but
this would get us by. In fact, I'm not even sure we need the feign
prefix as it is highly unlikely anything else uses this formatting
(javadoc link format).

Feign.configKey keeps formatting consistent with other feign things
(like logs) and also has uniqueness including the simple name and the
method's parameters. If you don't think it looks better in the UI,
call me out!

adriancole:
closed via #447

PR #447, opened 2016-08-17
Adds Hystrix SetterFactory to customize group and command keys

### LINKED ISSUE #333: improve documentation about hystrix default's like commandKey, groupKey
There no/ not enough hints in documentation how defaults on hystrix for Feign are set.
E.g. what is used as commandKey and groupKey for hystrix. Following the code there is one. but not documented
https://github.com/Netflix/feign/blob/master/hystrix/src/main/java/feign/hystrix/HystrixInvocationHandler.java#L50 

It's related to https://github.com/Netflix/feign/issues/189#issuecomment-95688652

Missing in general is documentation how Hystrix can be used (anotations based) on a FeignClient to set custom hystrix configuration.

### LINKED ISSUE #334: allow anotation of hystrix commandKey and groupKey for interfaced feign methods

### LINKED ISSUE #434: Make Hystrix Command names more unique

adriancole:
By default, Hystrix group keys match the target name, and the target name is usually the base url. Hystrix command keys are now the same as logging keys, which are equivalent to javadoc references.

For example, for the canonical GitHub example...

* the group key would be "https://api.github.com" and
* the command key would be "GitHub#contributors(String,String)"

You can use `HystrixFeign.Builder#setterFactory(SetterFactory)` to customize this, for example, to read key mappings from configuration or annotations.

Ex.
```java
SetterFactory commandKeyIsRequestLine = (target, method) -> {
  String groupKey = target.name();
  String commandKey = method.getAnnotation(RequestLine.class).value();
  return HystrixCommand.Setter
      .withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
      .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
};
```

Fixes #334
Fixes #434
Fixes #333
Obviates #444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### Version 9.2
+* Adds Hystrix `SetterFactory` to customize group and command keys
 * Supports context path when using Ribbon `LoadBalancingTarget`
 * Adds builder methods for the Response object
 * Deprecates Response factory methods
--- a/hystrix/README.md
+++ b/hystrix/README.md
@@ -50,6 +50,34 @@ api.getYourType("a").execute();
 api.getYourTypeSynchronous("a");
 ```
 
+### Group and Command keys
+
+By default, Hystrix group keys match the target name, and the target name is usually the base url.
+Hystrix command keys are the same as logging keys, which are equivalent to javadoc references.
+
+For example, for the canonical GitHub example...
+
+* the group key would be "https://api.github.com" and
+* the command key would be "GitHub#contributors(String,String)"
+
+You can use `HystrixFeign.Builder#setterFactory(SetterFactory)` to customize this, for example, to
+read key mappings from configuration or annotations.
+
+Ex.
+```java
+SetterFactory commandKeyIsRequestLine = (target, method) -> {
+  String groupKey = target.name();
+  String commandKey = method.getAnnotation(RequestLine.class).value();
+  return HystrixCommand.Setter
+      .withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
+      .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
+};
+
+api = HystrixFeign.builder()
+                  .setterFactory(commandKeyIsRequestLine)
+                  ...
+```
+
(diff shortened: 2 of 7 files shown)


ryanjbaxter:
I like this approach.  Sorry for being dense about the confiKey approach, after reviewing the code, I see what you are saying now.  +1 for this approach

adriancole:
sweet. I can cut a release tomorrow.

ryanjbaxter:
Great thanks!

spencergibb:
Well done @adriancole!

adriancole:
thx for the kind words, folks!

jpuzzler:
Hi,
sorry for the bump up but I didn't figure out **where/when** do I have to customize the `HystrixFeign.Builder#setterFactory(SetterFactory)` ?

I added the following within my application created 

```java
    @Autowired
    private static Feign.Builder feignBuilder;

     // Customizes Feign HystrixCommand commandKey
     if (feignBuilder instanceof HystrixFeign.Builder) {
         HystrixFeign.Builder builder = (HystrixFeign.Builder) feignBuilder;
         builder.setterFactory(setterFactory());
     }
```

is null !

Thanks

PR #448, opened 2016-08-26
Introducing ResponseInterceptors and properties for RequestTemplate

ravihansa3000:
This is much similar to jax-rs ClientResponseFilter. It is useful for monitoring request/reponse flow, if you need to publish stats to an external monitoring service.

<string, object> HashMap attribute is introduced to RequestTemplate to allow additional information to be passed in a client request/response context. This can be used to correlate tracking/origin id.


adriancole:
It is possible to request correlation without adding a new map of stuff to feign's types. Generally, it is better to ask on gitter or stackoverflow or review old issues before proposing change to Feign, as this sort of thing has come up before. https://github.com/OpenFeign/feign/blob/master/HACKING.md

Issue #449, opened 2016-08-26
Re-add OSGi headers to all modules


raulk:
OSGi headers were added on #281 and were being generated as of 8.18.0. Unfortunately, with the migration from Netflix/feign to OpenFeign/feign and the switch of the build system, they were lost.

adriancole:
on the way out as 9.3.1

PR #450, opened 2016-08-26
#449 Re-add OSGi headers to all modules.

raulk:
Via the maven-bundle-plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
     <maven-javadoc-plugin.version>2.10.3</maven-javadoc-plugin.version>
     <maven-jar-plugin.version>2.6</maven-jar-plugin.version>
     <maven-release-plugin.version>2.5.3</maven-release-plugin.version>
+    <maven-bundle-plugin.version>3.2.0</maven-bundle-plugin.version>
     <centralsync-maven-plugin.version>0.1.0</centralsync-maven-plugin.version>
   </properties>
 
@@ -260,6 +261,11 @@
         <plugin>
           <artifactId>maven-jar-plugin</artifactId>
           <version>${maven-jar-plugin.version}</version>
+          <configuration>
+            <archive>
+              <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+            </archive>
+          </configuration>
         </plugin>
       </plugins>
     </pluginManagement>
@@ -333,6 +339,21 @@
           <repo>maven</repo>
           <packageName>feign</packageName>
         </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>${maven-bundle-plugin.version}</version>
+        <executions>
+          <execution>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <goals>
+              <goal>manifest</goal>
+            </goals>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
   </build>


spencergibb:
@raulk how do I verify they were added?

spencergibb:
I see the following in `META-INF/MANIFEST.MF`:
```
Manifest-Version: 1.0
Bundle-Description: Feign Core
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: io.github.openfeign.feign-core
Archiver-Version: Plexus Archiver
Built-By: sgibb
Bnd-LastModified: 1472228033106
Bundle-ManifestVersion: 2
Bundle-DocURL: https://github.com/openfeign
Bundle-Vendor: OpenFeign
Import-Package: feign,feign.codec,javax.net.ssl
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Tool: Bnd-3.2.0.201605172007
Export-Package: feign.auth;uses:=feign;version="9.3.1",feign;uses:="fe
 ign.codec,javax.net.ssl";version="9.3.1",feign.codec;uses:=feign;vers
 ion="9.3.1"
Bundle-Name: Feign Core
Bundle-Version: 9.3.1.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Build-Jdk: 1.8.0_91
```

From 8.17.0 I see
```
Manifest-Version: 1.0
Build-Date: 2016-06-05_00:09:03
Bundle-SymbolicName: com.netflix.feign.core
Build-Number: LOCAL
Built-By: travis
Bnd-LastModified: 1465085374000
Import-Package: javax.net.ssl
Branch: 1c1d2fa91ed0788cad7a5a9901feb22cbf3de8c3
Gradle-Version: 2.2.1
Export-Package: feign;version="8.17.0";uses:="feign.codec,javax.net.ss
 l",feign.auth;version="8.17.0";uses:=feign,feign.codec;version="8.17.
 0";uses:=feign
Built-OS: Linux
Bundle-Name: feign-core
Build-Host: testing-worker-linux-docker-f919d37f-3389-linux-2
Module-Email: talent@netflix.com
Implementation-Title: com.netflix.feign#feign-core;8.17.0
Module-Origin: https://github.com/Netflix/feign.git
Build-Id: LOCAL
Implementation-Version: 8.17.0
Module-Owner: talent@netflix.com
Bundle-ManifestVersion: 2
Change: 1c1d2fa
Module-Source: /core
Tool: Bnd-2.1.0.20130426-122213
Built-Status: integration
Build-Job: LOCAL
Bundle-Version: 8.17.0
X-Compile-Target-JDK: 1.6
X-Compile-Source-JDK: 1.6
Created-By: 1.8.0_31-b13 (Oracle Corporation)
Build-Java-Version: 1.8.0_31
```

I'd like a plus one from someone not me and not the author that the above is sufficient.

dsyer:
IIRC the only thing that really matters is the Export/Import-Package and that looks sane. If the author says it works I think that's good enough for me. But how are we generating this data (the versions must change for every release)?

spencergibb:
@dsyer `org.apache.felix:maven-bundle-plugin` maven plugin. I just ran a normal build against this PR and got the first `MANIFEST.MF`.

adriancole:
I don't understand the nuance of import export, but is it normal for feign
to export javax.net.ssl?

raulk:
OSGi has a select set of headers, out of which the important ones are
Bundle-SymbolicName, Import-Package, Export-Package and a few others.

Many of the manifest headers in 8.18.0 were not OSGi related. I suppose
they were added by the Nebula plugins as metadata.

We do not export javax.net.ssl. If you look closely, that package is part
of a "uses" clause of an export, which is merely used by OSGi to calculate
a consistent class space.

Version ranges are updated automatically by the maven-bundle-plugin as the
POM version evolves. There's nothing additional to take care of.

adriancole:
Ahh ok.

Any way to test this so that we dont break it again?

Ex. an invoker test in src/main/it or something?

raulk:
Hmmm... We could consider a Pax Exam test. I'll look into that separately.
Would appreciate if you can merge the PR for now without a regression test.

adriancole:
Yeah sure. I think i remember PAM used in jclouds and it did catch when
people like me screwed up things. Either that or you just let us know next
time someone messes it up!

adriancole:
LGTM

ravihansa3000:
This doesn't work with google-gson 2.7 in an actual OSGi environment, Eclipse P2 OSGi container to be precise. Reason being feign-gson module has imports to google-gson internal packages [1] which are not exported from google-gson. google-gson has been explicitly instructed not to export internal packages [2].

Problem is not with this PR, but we need to fix DoubleToIntMapTypeAdapter before anyone could make use of this.

[1] https://github.com/OpenFeign/feign/blob/master/gson/src/main/java/feign/gson/DoubleToIntMapTypeAdapter.java#L21-L22

[2] https://github.com/google/gson/blob/master/gson/pom.xml#L28

adriancole:
we are not required to make all things work in OSGi, particularly as it is
a minority of the ecosystem that care about it. We can just add a note to
the gson README saying that if you care about OSGi, then don't use the
default constructor for the gson adapters, as they use an internal class

GsonEncoder(Iterable<TypeAdapter<?>> adapters)

adriancole:
@raulk ping back if you need to make an adjustment in maven config to exclude the gson internal class on that module. I raised https://github.com/OpenFeign/feign/pull/451 to document this, as it has existed in this way since feign was created

ravihansa3000:
IMHO if we accept to make OpenFeign OSGi ready then it should be required to work in OSGi environment. Or else it would be better not to have it at all.
Do you think it's a good idea to have feign-gson as a fragment bundle of google-gson host bundle?

adriancole:
Opinion noted, and if strict OSGi compliance for all aspects of all code
becomes a dominant opinion we can revisit.

For now, you can either choose to use the workaround, choose to put in
effort to make the internal classes internal, or choose not to use the gson
module.

raulk:
My goal with this ticket was to fix a regression whereby newest versions of Feign stopped generating OSGi headers altogether.

We could discuss about OSGi in a broader context – and how to make things nicer for Gson, but I don't think it should prevent this PR from being merged.

raulk:
@ravihansa3000 – I wouldn't deploy feign-gson as an OSGi fragment of Gson, that makes little sense technically speaking, because your fragment would have to declare the `feign.gson` export which would result in the Gson bundle exporting `feign.gson`.

If anything, I would consider creating a little helper OSGi fragment that does nothing but attach to Gson and export the internal package you need, e.g. feign-gson-osgi-helper.

That or change feign-gson so that it doesn't rely on internal classes of Gson, if at all possible.

raulk:
@spencergibb – thanks for the merge 👍

ravihansa3000:
@adriancole thank you for suggesting a workaround. 

@raulk I agree. I'll try to initiate a discussion on this in Gitter.

PR #451, opened 2016-08-29
Adds note about gson map decoding to README

adriancole:
See https://github.com/OpenFeign/feign/pull/450#issuecomment-243010282

### DIFF
--- a/gson/README.md
+++ b/gson/README.md
@@ -11,3 +11,11 @@ GitHub github = Feign.builder()
                      .decoder(new GsonDecoder())
                      .target(GitHub.class, "https://api.github.com");
 ```
+
+### Map<String, Object> and Numbers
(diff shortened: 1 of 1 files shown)


adriancole:
while I don't care enough to do more than document this, since there's a workaround. If someone wants to shade the internal classes used in `DoubleToIntMapTypeAdapter`, or redo it portably without use of internal classes, knock yourself out! I'll help review

raulk:
There is a less invasive approach. If someone really does want to use an internal class of Gson which the bundle doesn't expose, they can deploy an [OSGi fragment](http://docs.spring.io/osgi/docs/current/reference/html/appendix-tips.html) that attaches to the Gson bundle and exports the additional (internal) package from it.

adriancole:
I went ahead and found a solution to the issue as I can't afford the time
to continually discuss the two lines involved

https://github.com/OpenFeign/feign/pull/452

PR #452, opened 2016-08-30
Removes use of internal gson class

adriancole:
Turns out an api was added a very long time ago to do what we do in gson
without use of internal apis.

See #449


adriancole:
cc @raulk @ravihansa3000

spencergibb:
:+1:

raulk:
Awesome, great find!

PR #453, opened 2016-09-02
Support request content-type with charset

stolyarchukav:
Hello,
This fix helps to resolve problem of using feign with spring cloud, when post request contains header like Content-Type = application/json;charset=utf-8

### DIFF
--- a/core/src/test/java/feign/client/AbstractClientTest.java
+++ b/core/src/test/java/feign/client/AbstractClientTest.java
@@ -15,6 +15,7 @@ import feign.Logger;
 import feign.Param;
 import feign.RequestLine;
 import feign.Response;
+import feign.Util;
 import feign.assertj.MockWebServerAssertions;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
@@ -218,6 +219,18 @@ public abstract class AbstractClientTest {
         assertEquals(expected, actual);
     }
 
+    @Test
+    public void testContentTypeWithCharset() throws Exception {
+        server.enqueue(new MockResponse()
+                .setBody("AAAAAAAA"));
+        TestInterface api = newBuilder()
+                .target(TestInterface.class, "http://localhost:" + server.getPort());
+
+        Response response = api.postContentTypeWithCharset("foo");
+        // Response length should not be null
+        assertEquals("AAAAAAAA", Util.toString(response.body().asReader()));
+    }
+
     public interface TestInterface {
 
         @RequestLine("POST /?foo=bar&foo=baz&qux=")
@@ -241,6 +254,10 @@ public abstract class AbstractClientTest {
 
         @RequestLine("PUT")
         String noPutBody();
+
+        @RequestLine("POST /?foo=bar&foo=baz&qux=")
+        @Headers({"Foo: Bar", "Foo: Baz", "Qux: ", "Content-Type: text/plain;charset=utf-8"})
+        Response postContentTypeWithCharset(String body);
     }
 
 }
--- a/core/src/test/java/feign/client/AbstractClientTest.java
+++ b/core/src/test/java/feign/client/AbstractClientTest.java
@@ -226,7 +226,19 @@ public abstract class AbstractClientTest {
         TestInterface api = newBuilder()
                 .target(TestInterface.class, "http://localhost:" + server.getPort());
 
-        Response response = api.postContentTypeWithCharset("foo");
+        Response response = api.postWithContentType("foo", "text/plain;charset=utf-8");
+        // Response length should not be null
+        assertEquals("AAAAAAAA", Util.toString(response.body().asReader()));
+    }
+
+    @Test
+    public void testContentTypeWithoutCharset() throws Exception {
+        server.enqueue(new MockResponse()
+                .setBody("AAAAAAAA"));
+        TestInterface api = newBuilder()
+                .target(TestInterface.class, "http://localhost:" + server.getPort());
+
+        Response response = api.postWithContentType("foo", "text/plain");
         // Response length should not be null
         assertEquals("AAAAAAAA", Util.toString(response.body().asReader()));
     }
@@ -256,8 +268,8 @@ public abstract class AbstractClientTest {
         String noPutBody();
 
         @RequestLine("POST /?foo=bar&foo=baz&qux=")
-        @Headers({"Foo: Bar", "Foo: Baz", "Qux: ", "Content-Type: text/plain;charset=utf-8"})
-        Response postContentTypeWithCharset(String body);
+        @Headers({"Foo: Bar", "Foo: Baz", "Qux: ", "Content-Type: {contentType}"})
+        Response postWithContentType(String body, @Param("contentType") String contentType);
     }
 
 }
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -157,11 +157,9 @@ public final class ApacheHttpClient implements Client {
         Collection<String> values = entry.getValue();
         if (values != null && !values.isEmpty()) {
           String mimeType = values.iterator().next();
-          try {
-            contentType = ContentType.create(mimeType, request.charset());
-          } catch (IllegalArgumentException e) {
-            // MIME type may not contain reserved characters, so try to parse content type
-            contentType = ContentType.parse(mimeType);
+          contentType = ContentType.parse(mimeType);
+          if (contentType.getCharset() == null) {
+            contentType = ContentType.create(contentType.getMimeType(), request.charset());
           }
           break;
         }
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -156,11 +156,7 @@ public final class ApacheHttpClient implements Client {
       if (entry.getKey().equalsIgnoreCase("Content-Type")) {
         Collection<String> values = entry.getValue();
         if (values != null && !values.isEmpty()) {
-          String mimeType = values.iterator().next();
-          contentType = ContentType.parse(mimeType);
-          if (contentType.getCharset() == null) {
-            contentType = ContentType.create(contentType.getMimeType(), request.charset());
-          }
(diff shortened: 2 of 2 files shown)


core/src/test/java/feign/client/AbstractClientTest.java
```diff
@@ -1,31 +1,22 @@
...
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
-
-import feign.Client;
+import feign.*;
```
spencergibb: don't do `*` imports

core/src/test/java/feign/client/AbstractClientTest.java
```diff
@@ -1,31 +1,22 @@
...
-import feign.Response;
 import feign.assertj.MockWebServerAssertions;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
-import static java.util.Arrays.asList;
+import java.io.ByteArrayInputStream;
```
spencergibb: don't rearrange imports

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -154,9 +154,14 @@ private ContentType getContentType(Request request) {
...
     if (entry.getKey().equalsIgnoreCase("Content-Type")) {
-      Collection values = entry.getValue();
+      Collection<String> values = entry.getValue();
       if (values != null && !values.isEmpty()) {
-        contentType = ContentType.create(entry.getValue().iterator().next(), request.charset());
+        String mimeType = values.iterator().next();
+        try {
+          contentType = ContentType.create(mimeType, request.charset());
+        } catch (IllegalArgumentException e) {
+          contentType = ContentType.parse(mimeType);
```
spencergibb: put a comment here explaining why your doing this.

stolyarchukav:
Thanks, changeset updated

stolyarchukav:
Please review

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -153,13 +153,19 @@ HttpUriRequest toHttpUriRequest(Request request, Request.Options options) throws
...
-        break;
+      if (entry.getKey().equalsIgnoreCase("Content-Type")) {
+        Collection<String> values = entry.getValue();
+        if (values != null && !values.isEmpty()) {
+          String mimeType = values.iterator().next();
+          try {
+            contentType = ContentType.create(mimeType, request.charset());
+          } catch (IllegalArgumentException e) {
+            // MIME type may not contain reserved characters, so try to parse content type
+            contentType = ContentType.parse(mimeType);
```
adriancole: since you are catching.. why not test this :)

core/src/test/java/feign/client/AbstractClientTest.java
```diff
@@ -241,6 +254,10 @@ public void testResponseLength() throws Exception {
 
         @RequestLine("PUT")
         String noPutBody();
+
+        @RequestLine("POST /?foo=bar&foo=baz&qux=")
+        @Headers({"Foo: Bar", "Foo: Baz", "Qux: ", "Content-Type: text/plain;charset=utf-8"})
```
adriancole: if you make Content-Type a header param, you can test it easier

adriancole:
nice progress!

core/src/test/java/feign/client/AbstractClientTest.java
stolyarchukav: Could you please help me to find an example?

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
stolyarchukav: Do you mean test this on correctness before trying to call create method?

core/src/test/java/feign/client/AbstractClientTest.java
adriancole: yep. FeignTestInterface.java has this:

@RequestLine("POST /")
@Headers({"Happy: sad", "Auth-Token: {authToken}"})
void headers(@Param("authToken") String token);


On Sun, Sep 4, 2016 at 1:37 AM, Andrey Stolyarchuk <notifications@github.com

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
adriancole: I meant that the comment "MIME type may not contain reserved characters, so
try to parse content type" has an if statement. this code shouldn't exist
unless it is tested, so I'd make a test that exercises that if statement

On Sun, Sep 4, 2016 at 1:43 AM, Andrey Stolyarchuk <notifications@github.com

stolyarchukav: Test added

core/src/test/java/feign/client/AbstractClientTest.java
stolyarchukav: Thanks, test was improved

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -153,13 +153,19 @@ HttpUriRequest toHttpUriRequest(Request request, Request.Options options) throws
...
-      if (values != null && !values.isEmpty()) {
-        contentType = ContentType.create(entry.getValue().iterator().next(), request.charset());
-        break;
+      if (entry.getKey().equalsIgnoreCase("Content-Type")) {
+        Collection<String> values = entry.getValue();
+        if (values != null && !values.isEmpty()) {
+          String mimeType = values.iterator().next();
+          try {
+            contentType = ContentType.create(mimeType, request.charset());
+          } catch (IllegalArgumentException e) {
```
adriancole: so the thing I'm curious about is that ContentType.create doesn't declare that it throws IllegalArgument, rather unsupported. https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/entity/ContentType.html#create(java.lang.String,%20java.lang.String)

Are you sure this branch of code is in use?

stolyarchukav: Yes, I'm sure it throws IllegalArgumentException in my use case and it broke our code, so currently i need to copy-paste this feign class in own code base.

adriancole: can you please make a test case that causes the
IllegalArgumentException? copy/pasting stuff works until the
implementation changes invalidating the assumption.

stolyarchukav: Test was written first, see testContentTypeWithCharset.
"copy/pasting stuff works until the
implementation changes invalidating the assumption." - That exactly why I requesting a pull

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -153,13 +153,19 @@ HttpUriRequest toHttpUriRequest(Request request, Request.Options options) throws
...
     for (Map.Entry<String, Collection<String>> entry : request.headers().entrySet())
-    if (entry.getKey().equalsIgnoreCase("Content-Type")) {
-      Collection values = entry.getValue();
-      if (values != null && !values.isEmpty()) {
-        contentType = ContentType.create(entry.getValue().iterator().next(), request.charset());
-        break;
+      if (entry.getKey().equalsIgnoreCase("Content-Type")) {
+        Collection<String> values = entry.getValue();
+        if (values != null && !values.isEmpty()) {
+          String mimeType = values.iterator().next();
```
adriancole: ok I figured out what's going wrong here. We are blindly throwing the entire content-type header into the mimeType field. The IllegalArgumentException is present because of the ';' which is invalid. AKA we have a bug in this code. catching the exception doesn't improve anything and it causes confusion.

The correct behavior is to parse the charset. Ex. behavior like this..

```java
String unparsed = values.iterator().next();
if (unparsed.indexOf(';') != -1) { // use charset from the response
  String mimeType = get part before ';'
  String charSet = get part after ';'
  contentType = ContentType.create(mimeType, charSet);
} else { // use charset from the request instead.
  contentType = ContentType.create(unparsed, request.charset());
}
```

adriancole: so for example "text/plain; charset=ISO-8859-1" mimeType="text/plain"  charSet="ISO-8859-1"

stolyarchukav: I agree, in this case you are doing the same as already done in apache http client code. Would not it be easer just to use ContentType.parse instead of create?

stolyarchukav:
Could you please have a look at the last change? Seems it is a proper way to reach the desired behaviour correctly.

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
adriancole: yeap if there's a feature already to do that, by all means.. use it!

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -153,13 +153,17 @@ HttpUriRequest toHttpUriRequest(Request request, Request.Options options) throws
...
+          contentType = ContentType.parse(mimeType);
+          if (contentType.getCharset() == null) {
+            contentType = ContentType.create(contentType.getMimeType(), request.charset());
```
adriancole: this was a bad recommendation of me. we shouldn't conflate the response charset with the request one.

Please simplify the code to the following vs pretending we know what the response charset is.. You'll notice all tests pass!

```
      if (values != null && !values.isEmpty()) {
        contentType = ContentType.parse(entry.getValue().iterator().next());
        break;
      }
```

stolyarchukav: Adrian, I removed using of request context here. Looks like it could be applied

adriancole:
thanks tons.. looks good.

PR #459, opened 2016-09-08
Remove overriding of retry handler.

JonathanO:
Setting the retry handler explicitly to the DEFAULT instance overrides
the retry configuration, making it impossible to specify a configuration.
It seems to be unnecessary since allowing LoadBalancerContext to
initialize it in initWithNiwsConfig appears to be sufficient, as if
there's no configuration the defaults from there should be appropriate.

### DIFF
--- a/ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
+++ b/ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
@@ -19,12 +19,16 @@ import static com.netflix.config.ConfigurationManager.getConfigInstance;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 
 import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
@@ -40,6 +44,8 @@ import feign.Feign;
 import feign.Param;
 import feign.Request;
 import feign.RequestLine;
+import feign.RetryableException;
+import feign.Retryer;
 import feign.client.TrustingSSLSocketFactory;
 
 public class RibbonClientTest {
@@ -50,6 +56,26 @@ public class RibbonClientTest {
   public final MockWebServer server1 = new MockWebServer();
   @Rule
   public final MockWebServer server2 = new MockWebServer();
+
+  private static String oldRetryConfig = null;
+
+  private static final String SUN_RETRY_PROPERTY = "sun.net.http.retryPost";
+
+  @BeforeClass
+  public static void disableSunRetry() throws Exception {
+    // The Sun HTTP Client retries all requests once on an IOException, which makes testing retry code harder than would
+    // be ideal. We can only disable it for post, so lets at least do that.
+    oldRetryConfig = System.setProperty(SUN_RETRY_PROPERTY, "false");
+  }
+
+  @AfterClass
+  public static void resetSunRetry() throws Exception {
+    if (oldRetryConfig == null) {
+      System.clearProperty(SUN_RETRY_PROPERTY);
+    } else {
+      System.setProperty(SUN_RETRY_PROPERTY, oldRetryConfig);
+    }
+  }
 
   static String hostAndPort(URL url) {
     // our build slaves have underscores in their hostnames which aren't permitted by ribbon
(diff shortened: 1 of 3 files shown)


spencergibb:
Wonder if you could come up with a test?

JonathanO:
OK, so that opened a can of worms ;-)
For starters I was somewhat surprised to discover that the underlying Sun http client implementation retries both POST and GET on some IOExceptions. Fortunately you can disable this behaviour for POST, though not GET. This led to more retries against my mock servers than I anticipated, and means neither Feign nor Ribbon actually have any control or visibility of this first retry! I've disabled these retries for the tests simply so that I can be certain we're actually testing the ribbon and feign retries.

Secondly, my change did modify the existing behaviour. While RetryHandler.DEFAULT sets retryNextServer to 0, very unfortunately the value of DefaultClientConfigImpl.DEFAULT_MAX_AUTO_RETRIES_NEXT_SERVER, which is used to populate the client config, is set to 1, leading to 1 retry being made. I've had to alter the LBClientFactory to pass a new default configuration which uses 0 as the default for that setting.

JonathanO:
Does this look sensible now? I could remove the disabling of sun.net.http.retryPost in the test and increase the number of expected retries, if that'd be preferable, though I don't know what the outcome in other JVMs would be.

ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
```diff
@@ -51,6 +57,26 @@
...
 
+  private static String oldRetryConfig = null;
+
+  private static final String SUN_RETRY_PROPERTY = "sun.net.http.retryPost";
+
+  @BeforeClass
+  public static void disableSunRetry() throws Exception {
+    // The Sun HTTP Client retries all requests once on an IOException, which makes testing retry code harder than would
+    // be ideal. We can only disable it for post, so lets at least do that.
+    oldRetryConfig = System.setProperty(SUN_RETRY_PROPERTY, "false");
```
adriancole: wow didn't know about this one

ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
```diff
@@ -98,6 +124,83 @@ public void ioExceptionRetry() throws IOException, InterruptedException {
...
+                    .target(TestInterface.class, "http://" + client());
+
+    try {
+      api.post();
+      fail("No exception thrown");
+    } catch (RetryableException ignored) {
+
+    }
+    assertEquals(1, server1.getRequestCount());
+    // TODO: verify ribbon stats match
```
adriancole: we should stop copy/pasting this TODO :) after however many years, I think we can give up on that.

adriancole:
@ryanjbaxter you cool with this?

ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
JonathanO: Neither did I until I wrote that test.

It's all about handling Keep-Alive connections. It'd be unfair on the application to propagate the IOException if it's caused by the reuse of a Keep-Alive connection that happened to be shutdown by the server (or a NAT/firewall timeout) at exactly the wrong moment, so the client will retry once when it looks like that might have happened (i.e. an IO error while writing a request or reading the response headers.)

This behaviour appears to be explicitly permitted, even for methods that aren't idempotent, by [RFC7230 section 6.3.2](https://tools.ietf.org/html/rfc7230#section-6.3.2).
Out of curiosity I had a quick dig through the Chromium source, and, assuming I'm reading it correctly, it too follows this behaviour. The Apache commons client, meanwhile, I think wont for a POST by default, which is why it'll sometimes throw inexplicable connection errors if keepalives are in use.

ryanjbaxter:
So if I am understanding this correctly, if we don't specify a `RetryHandler` than no retries will occur?

JonathanO:
Basically, yes, since I felt I should preserve the default behaviour prior to this PR. The RetryHandler will now be instantiated using the ClientConfig, which as an instance of DisableAutoRetriesByDefaultClientConfig defaults to the same behaviour as RetryHandler.DEFAULT (i.e. no retries.) This allows the Ribbon retry behaviour to be optionally configured via the ConfigurationManager instance, overriding the default.

ryanjbaxter:
👍 sounds good to me

ryanjbaxter:
Just to clarify something regarding which retry logic we are talking about here...

This change effect the Ribbon (ie. `RetryHandler`) retry logic and not the Feign retry logic (ie `Retryer`)

JonathanO:
Yes, this only affects the Ribbon retry logic.

adriancole:
I can merge this (or someone else can) tomorrow. Waiting a day in case
anyone changes their mind, as this is behavior affecting.

Issue #462, opened 2016-09-19
'HTTP 204 No Content' response results in java.io.EOFException in feign.slf4j.Slf4jLogger


maximdim:
REST DELETE call defined in Feign client as follows:

```
@RequestLine("DELETE /students/{id}")
void deleteStudent(@Param("id") String id);
```

Server returns "HTTP 204 No Content" which results in exception:

```
---> DELETE https://test.foo/students/f26f3612-3660-427c-a8b5-dc428de17df0
<--- HTTP/1.1 204 No Content (240ms)
cache-control: no-cache
content-type: application/json
date: Mon, 19 Sep 2016 17:14:26 GMT
expires: -1
okhttp-received-millis: 1474305275717
okhttp-sent-millis: 1474305275468
pragma: no-cache
request-id: 1836d724-f62b-4773-90d9-3c78c24f4adf
server: Microsoft-IIS/8.5
x-aspnet-version: 4.0.30319
x-powered-by: ASP.NET

<--- ERROR EOFException: null (240ms)
java.io.EOFException
    at okio.RealBufferedSource.require(RealBufferedSource.java:64)
    at okio.GzipSource.consumeHeader(GzipSource.java:114)
    at okio.GzipSource.read(GzipSource.java:73)
    at okio.RealBufferedSource$1.read(RealBufferedSource.java:371)
    at java.io.InputStream.read(InputStream.java:101)
    at feign.Util.copy(Util.java:307)
    at feign.Util.toByteArray(Util.java:290)
    at feign.Logger.logAndRebufferResponse(Logger.java:96)
    at feign.slf4j.Slf4jLogger.logAndRebufferResponse(Slf4jLogger.java:62)
    at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:112)
    at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:76)
    at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:103)
    at com.sun.proxy.$Proxy13.deleteStudent(Unknown Source)
    at Client.deleteStudent(IsusClient2.java:318)
```

maximdim:
Feign version 9.3.1

PR #463, opened 2016-09-20
Fix for https://github.com/OpenFeign/feign/issues/462

maximdim:


core/src/main/java/feign/Logger.java
```diff
@@ -89,7 +90,9 @@ protected Response logAndRebufferResponse(String configKey, Level logLevel, Resp
       }
 
       int bodyLength = 0;
-      if (response.body() != null) {
+      if (response.body() != null && !(status == 204 || status == 205)) {
```
adriancole: According to the rules below, it seems erroneous if there a body when status like this is sent? what http client are you using? could you add a MockWebServer test to show this? Why don't we fix the client instead?

maximdim: I'm using OkHttpClient. There was similar issue in retrofit and from what I can tell they decided that this is not okhttp bug: https://github.com/square/retrofit/issues/735

adriancole:
We can make different decisions than retrofit (we have in the past). please
add tests to the base http client. you may find this only affects okhttp,
or not.

On Tue, Sep 20, 2016 at 9:41 AM, Dmitri Maximovich <notifications@github.com

adriancole:
since this only affects the logger, I'm not really bothered by this. we should add a test that it works. Can you do this?

PR #465, opened 2016-09-26
Implements possibility to disable @Param url encoding

parkee:
Now it is possible with @Param(encoded = true) to disable hardcoded call of urlEncode method on every @Param
This helps to implement matrix params or send unencoded path variables or query parameters which is required by some APIs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### Version 9.3
 * Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
-* Adds support for encoded parameters in @Param via @Param(encoded = true)
+* Adds support for encoded parameters via @Param(encoded = true)
 
 ### Version 9.2
 * Adds Hystrix `SetterFactory` to customize group and command keys
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### Version 9.3
 * Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
-* Adds support for encoded parameters via @Param(encoded = true)
+* Adds support for encoded parameters via `@Param(encoded = true)`
 
 ### Version 9.2
 * Adds Hystrix `SetterFactory` to customize group and command keys
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Version 9.3
 * Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
+* Adds support for encoded parameters in @Param via @Param(encoded = true)
 
 ### Version 9.2
 * Adds Hystrix `SetterFactory` to customize group and command keys
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -249,7 +249,7 @@ public interface Contract {
       for (Annotation annotation : annotations) {
         Class<? extends Annotation> annotationType = annotation.annotationType();
         if (annotationType == Param.class) {
-          final Param paramAnnotation = (Param) annotation;
+          Param paramAnnotation = (Param) annotation;
           String name = paramAnnotation.value();
           checkState(emptyToNull(name) != null, "Param annotation was empty on param %s.", paramIndex);
           nameParam(data, name, paramIndex);
@@ -288,7 +288,7 @@ public interface Contract {
 
       for (Collection<String> entry : values) {
         for (String value : entry) {
-          if (value.indexOf(search) != -1) {
+          if (value.contains(search)) {
             return true;
           }
         }
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -249,19 +249,18 @@ public interface Contract {
       for (Annotation annotation : annotations) {
         Class<? extends Annotation> annotationType = annotation.annotationType();
         if (annotationType == Param.class) {
-          String name = ((Param) annotation).value();
-          checkState(emptyToNull(name) != null, "Param annotation was empty on param %s.",
-              paramIndex);
+          final Param paramAnnotation = (Param) annotation;
+          String name = paramAnnotation.value();
+          checkState(emptyToNull(name) != null, "Param annotation was empty on param %s.", paramIndex);
           nameParam(data, name, paramIndex);
-          if (annotationType == Param.class) {
-            Class<? extends Param.Expander> expander = ((Param) annotation).expander();
-            if (expander != Param.ToStringExpander.class) {
-              data.indexToExpanderClass().put(paramIndex, expander);
-            }
+          Class<? extends Param.Expander> expander = paramAnnotation.expander();
+          if (expander != Param.ToStringExpander.class) {
+            data.indexToExpanderClass().put(paramIndex, expander);
           }
+          data.indexToEncoded().put(paramIndex, paramAnnotation.encoded());
           isHttpAnnotation = true;
           String varName = '{' + name + '}';
-          if (data.template().url().indexOf(varName) == -1 &&
+          if (!data.template().url().contains(varName) &&
               !searchMapValuesContainsSubstring(data.template().queries(), varName) &&
               !searchMapValuesContainsSubstring(data.template().headers(), varName)) {
             data.formParams().add(name);
--- a/core/src/main/java/feign/Param.java
+++ b/core/src/main/java/feign/Param.java
@@ -40,7 +40,9 @@ public @interface Param {
 
   /**
    * Specifies whether argument is already encoded
-   * Value be ignored for headers (headers are never encoded)
+   * The value is ignored for headers (headers are never encoded)
+   *
+   * @see QueryMap#encoded
    */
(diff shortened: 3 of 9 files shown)


core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -291,7 +291,14 @@ private RequestTemplate addQueryMapQueryParameters(Object[] argv, RequestTemplat
 
     protected RequestTemplate resolve(Object[] argv, RequestTemplate mutable,
                                       Map<String, Object> variables) {
-      return mutable.resolve(variables);
+      final HashMap<String, Boolean> variableToEncoded = new HashMap<String, Boolean>();
```
adriancole: LinkedHashMap pls

parkee:
Added one test to exemplify this behaviour and changed HashMap to LinkedHashMap
We are using matrix params in one of out internal API
Also wheel-size api will not accept encoded parameter: i.e. trim should be "5.2FSi+", not "5.2FSi%2b":
So this request won't work 
https://api.wheel-size.com/v1/vehicles/?user_key=<user_key>&make=audi&model=r8&year=2015&trim=5.2FSi%2b
But not encoded will:
https://api.wheel-size.com/v1/vehicles/?user_key=<user_key>&make=audi&model=r8&year=2015&trim=5.2FSi+
In some cases it would be path variable, not query parameter

core/src/main/java/feign/Contract.java
```diff
@@ -249,19 +249,18 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
       for (Annotation annotation : annotations) {
         Class<? extends Annotation> annotationType = annotation.annotationType();
         if (annotationType == Param.class) {
-          String name = ((Param) annotation).value();
-          checkState(emptyToNull(name) != null, "Param annotation was empty on param %s.",
-              paramIndex);
+          final Param paramAnnotation = (Param) annotation;
```
adriancole: nit we don't usually mark locals final unless doing so is notable (ex used in an anonymous type)

core/src/main/java/feign/Contract.java
```diff
@@ -249,19 +249,18 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
...
+          String name = paramAnnotation.value();
+          checkState(emptyToNull(name) != null, "Param annotation was empty on param %s.", paramIndex);
           nameParam(data, name, paramIndex);
-          if (annotationType == Param.class) {
```
adriancole: weird.. we actually checked twice :P

core/src/main/java/feign/Contract.java
```diff
@@ -249,19 +249,18 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
...
-              data.indexToExpanderClass().put(paramIndex, expander);
-            }
+          Class<? extends Param.Expander> expander = paramAnnotation.expander();
+          if (expander != Param.ToStringExpander.class) {
+            data.indexToExpanderClass().put(paramIndex, expander);
           }
+          data.indexToEncoded().put(paramIndex, paramAnnotation.encoded());
           isHttpAnnotation = true;
           String varName = '{' + name + '}';
-          if (data.template().url().indexOf(varName) == -1 &&
```
adriancole: fair enough!

core/src/main/java/feign/Param.java
```diff
@@ -38,6 +38,12 @@
    */
   Class<? extends Expander> expander() default ToStringExpander.class;
 
+  /**
+   * Specifies whether argument is already encoded
+   * Value be ignored for headers (headers are never encoded)
```
adriancole: good clarification

adriancole: probably should also put `@see QueryMap#encoded` as well putting a see link there back to here

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -291,7 +291,14 @@ private RequestTemplate addQueryMapQueryParameters(Object[] argv, RequestTemplat
 
     protected RequestTemplate resolve(Object[] argv, RequestTemplate mutable,
                                       Map<String, Object> variables) {
-      return mutable.resolve(variables);
+      final Map<String, Boolean> variableToEncoded = new LinkedHashMap<String, Boolean>();
```
adriancole: add a comment of what's going on here. (also pls unfinal the locals)

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -200,21 +200,28 @@ private static void putKV(String stringToParse, Map<String, Collection<String>>
     map.put(key, values);
   }
 
+  public RequestTemplate resolve(Map<String, ?> unencoded) {
```
adriancole: add see link

ex.
```java
/** like {@link #resolve(Map, Map)}, which assumes no parameter is encoded */
```

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -593,7 +609,7 @@ public String toString() {
    * Replaces query values which are templated with corresponding values from the {@code unencoded}
    * map. Any unresolved queries are removed.
    */
-  public void replaceQueryValues(Map<String, ?> unencoded) {
```
adriancole: does this break api? if so, please add an overload

core/src/test/java/feign/FeignTest.java
```diff
@@ -704,6 +716,9 @@ void form(
     @RequestLine("GET /?name={name}")
     void queryMapWithQueryParams(@Param("name") String name, @QueryMap Map<String, Object> queryMap);
 
+    @RequestLine("GET /?trim={trim}")
+    void encodedQueryParam(@Param(value = "trim", encoded = true) String trim);
```
adriancole: nice

core/src/main/java/feign/RequestTemplate.java
adriancole: please overload with the prior signature so that we don't break other people's code.

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -593,7 +609,7 @@ public String toString() {
...
+  public void replaceQueryValues(Map<String, ?> unencoded, Map<String, Boolean> alreadyEncoded) {
```
adriancole: does this need to be public (ex are you using this directly?)

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### Version 9.3
 * Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
+* Adds support for encoded parameters in @Param via @Param(encoded = true)
```
adriancole: simplify and format? Ex.

* Adds support for encoded parameters via `@Param(encoded = true)`

core/src/main/java/feign/RequestTemplate.java
parkee: ? It's already done.

core/src/main/java/feign/RequestTemplate.java
parkee: I see no reason in original method being public either, so I decided to leave overloaded one public too to be consistent.
Should I make new method private instead of overloading public method?

CHANGELOG.md
parkee: ok

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### Version 9.3
 * Adds `FallbackFactory`, allowing access to the cause of a Hystrix fallback
+* Adds support for encoded parameters via @Param(encoded = true)
```
adriancole: nit, if you get used to doing backticks, not only will the format look nicer, but you won't accidentally mention someone named Param :P

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -200,21 +200,29 @@ private static void putKV(String stringToParse, Map<String, Collection<String>>
...
+
   /**
    * Resolves any template parameters in the requests path, query, or headers against the supplied
    * unencoded arguments. <br> <br><br><b>relationship to JAXRS 2.0</b><br> <br> This call is
    * similar to {@code javax.ws.rs.client.WebTarget.resolveTemplates(templateValues, true)} , except
    * that the template values apply to any part of the request, not just the URL
    */
-  public RequestTemplate resolve(Map<String, ?> unencoded) {
-    replaceQueryValues(unencoded);
+  public RequestTemplate resolve(Map<String, ?> unencoded, Map<String, Boolean> alreadyEncoded) {
```
adriancole: for this and below, package private is fine (as opposed to private). these are new methods, so we can hide them until someone asks. package private is better as it can be tested directly (plus the bytecode is smaller)

core/src/main/java/feign/RequestTemplate.java
adriancole: diff problem no worries

core/src/main/java/feign/RequestTemplate.java
adriancole: package private por favor

Issue #468, opened 2016-09-27
Provide ability to control how IOExceptions and retries are logged


rfalke:
Currently it is not possible to override feign.Logger#logIOException() and feign.Logger#logRetry() in a custom logger since the methods have package access level. There seems no reason reason for this. Can these methods be made at least protected?

adriancole:
fair enough!

PR #469, opened 2016-09-27
Make logRetry() and logIOException() protected

rfalke:


rfalke:
Fix for #468

adriancole:
I see your point. we usually don't open things until they are asked for, and in this case I don't see any harm doing so.

EXPEkmajewski:
I am also interested in consuming this fix. Any idea when it will be released? Thanks.

rfalke:
Well it is not in 9.3.1. No idea when the next version will be created.
@adriancole ?

adriancole:
9.4 on the way now! thanks for the nag

PR #481, opened 2016-10-25
Allow ErrorDecoder.decode(...) to return null which signals that no ...

rfalke:
... exception should be thrown. This allows to react to non-200 statuses not only with an
exception but also to return it in the return object.

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -107,7 +107,7 @@ public abstract class Logger {
     return response;
   }
 
-  protected IOException logIOException(String configKey, Level logLevel, IOException ioe, long elapsedTime) {
+  protected void logIOException(String configKey, Level logLevel, IOException ioe, long elapsedTime) {
     log(configKey, "<--- ERROR %s: %s (%sms)", ioe.getClass().getSimpleName(), ioe.getMessage(),
         elapsedTime);
     if (logLevel.ordinal() >= Level.FULL.ordinal()) {
@@ -116,7 +116,6 @@ public abstract class Logger {
       log(configKey, sw.toString());
       log(configKey, "<--- END ERROR");
     }
-    return ioe;
   }
 
   /**
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -79,7 +79,6 @@ final class SynchronousMethodHandler implements MethodHandler {
         if (logLevel != Logger.Level.NONE) {
           logger.logRetry(metadata.configKey(), logLevel);
         }
-        continue;
       }
     }
   }
@@ -105,14 +104,25 @@ final class SynchronousMethodHandler implements MethodHandler {
     }
     long elapsedTime = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
 
-    boolean shouldClose = true;
     try {
       if (logLevel != Logger.Level.NONE) {
         response =
-            logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
+                logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
         // ensure the request is set. TODO: remove in Feign 10
         response.toBuilder().request(request).build();
       }
+      return decodeResponse(response);
+    } catch (IOException e) {
+      if (logLevel != Logger.Level.NONE) {
+        logger.logIOException(metadata.configKey(), logLevel, e, elapsedTime);
+      }
+      throw errorReading(request, e);
+    }
+  }
+
+  private Object decodeResponse(Response response) throws Throwable {
+    boolean shouldClose = true;
+    try {
       if (Response.class == metadata.returnType()) {
         if (response.body() == null) {
           return response;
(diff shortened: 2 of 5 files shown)


core/src/main/java/feign/Logger.java
```diff
@@ -107,7 +107,7 @@ protected Response logAndRebufferResponse(String configKey, Level logLevel, Resp
     return response;
   }
 
-  protected IOException logIOException(String configKey, Level logLevel, IOException ioe, long elapsedTime) {
+  protected void logIOException(String configKey, Level logLevel, IOException ioe, long elapsedTime) {
```
spencergibb: Why this change? It breaks an API?

core/src/main/java/feign/codec/ErrorDecoder.java
```diff
@@ -76,13 +73,15 @@
...
-   * application-specific exception decoded by the implementation. If the throwable is retryable, it
-   * should be wrapped, or a subtype of {@link RetryableException}
+   * @return Exception An Exception or <code>null</code>. IOException, if there was a network error
+   * reading the response or an application-specific exception decoded by the implementation. If
+   * the throwable is retryable, it should be wrapped, or a subtype of {@link RetryableException}.
+   * If <code>null</code> is returned no exception is thrown and the normal {@link Decoder} will
+   * be asked to decode the response.
    */
-  public Exception decode(String methodKey, Response response);
+  Exception decode(String methodKey, Response response);
```
spencergibb: Why change visibility?

core/src/main/java/feign/codec/ErrorDecoder.java
```diff
@@ -76,13 +73,15 @@
...
 
-  public static class Default implements ErrorDecoder {
+  class Default implements ErrorDecoder {
```
spencergibb: Why?

core/src/main/java/feign/codec/ErrorDecoder.java
```diff
@@ -108,7 +107,7 @@ public Exception decode(String methodKey, Response response) {
    * Decodes a {@link feign.Util#RETRY_AFTER} header into an absolute date, if possible. <br> See <a
    * href="https://tools.ietf.org/html/rfc2616#section-14.37">Retry-After format</a>
    */
-  static class RetryAfterDecoder {
+  class RetryAfterDecoder {
```
spencergibb: Why?

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -129,26 +139,31 @@ Object executeAndDecode(RequestTemplate template) throws Throwable {
...
-      throw errorReading(request, response, e);
+      return decodeWithDecoder(response);
     } finally {
       if (shouldClose) {
         ensureClosed(response.body());
       }
     }
   }
 
+  Object decodeWithDecoder(Response response) throws Throwable {
```
spencergibb: Why move this?

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -105,14 +104,25 @@ Object executeAndDecode(RequestTemplate template) throws Throwable {
...
-    boolean shouldClose = true;
     try {
       if (logLevel != Logger.Level.NONE) {
         response =
-            logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
+                logger.logAndRebufferResponse(metadata.configKey(), logLevel, response, elapsedTime);
         // ensure the request is set. TODO: remove in Feign 10
         response.toBuilder().request(request).build();
       }
+      return decodeResponse(response);
```
spencergibb: Moving this to another method just makes it harder to see what you are actually changing.

core/src/main/java/feign/Logger.java
rfalke: 1. The return value is unused. 
1. The method is neither a command nor a query.

core/src/main/java/feign/codec/ErrorDecoder.java
rfalke: Interface methods are public by default. The 'public' is redundant here. The IDE flagged this.

I will re-add it if the code style guide requires it.

core/src/main/java/feign/codec/ErrorDecoder.java
rfalke: Same.

core/src/main/java/feign/SynchronousMethodHandler.java
rfalke: The old method was already big. Too big for my taste. Even the new decodeResponse() is on the upper limit for my taste.

core/src/main/java/feign/SynchronousMethodHandler.java
rfalke: So that is nearer to the one-and-only caller.

rfalke:
Just tell me which changes should be reverted and I will do so.

core/src/main/java/feign/Logger.java
adriancole: not good enough reasons, as this would break people who are extending this regardless.

a good enough reason is that this method was exposed protected, but not yet released! (unless I'm mistaken) This means that no one would have been able to extend it yet, so we wouldn't be breaking anyone.

core/src/main/java/feign/codec/ErrorDecoder.java
adriancole: it is more about making distracting change in the middle of a request, not about some style guide ruling. please be considerate to the time it takes to figure out what is actually needed or not.

core/src/test/java/feign/FeignTest.java
```diff
@@ -548,7 +549,7 @@ public Object decode(Response response, Type type) throws IOException {
   @Test
   public void decoderCanThrowUnwrappedExceptionInDecode404Mode() throws Exception {
     server.enqueue(new MockResponse().setResponseCode(404));
-    thrown.expect(NoSuchElementException.class);
+    thrown.expectCause(isA(NoSuchElementException.class));
```
adriancole: was this needed?

rfalke:
Is superseded by #487 and #488.

PR #486, opened 2016-11-04
Remove setting of properties since it broke tests in some environments

JonathanO:
In some environments (not sure exactly what it is about them that causes the problem) the setting of the "sun.net.http.retryPost" property in the RibbonClientTest fails to have any effect, causing the tests to fail.

This PR removes use of that property, and instead just accounts for the extra retries caused by URLConnection.


spencergibb:
I'd like to know what environments rather than trying to mask some symptoms.

adriancole:
closing until interest resumes

PR #487, opened 2016-11-07
Also wrap the exception of Decoder.decode() in case of a 404.

rfalke:


core/src/test/java/feign/FeignTest.java
```diff
@@ -546,9 +547,10 @@ public Object decode(Response response, Type type) throws IOException {
   }
 
   @Test
-  public void decoderCanThrowUnwrappedExceptionInDecode404Mode() throws Exception {
```
adriancole: so this is making the behavior the opposite of the test case. I'm ok with merging this, but if it gets kick-back we'd revert in a patch.

PR #488, opened 2016-11-07
Allow ErrorDecoder.decode(...) to return null.

rfalke:
Null signals that no exception should be thrown. This allows to
react to non-200 statuses not only with an exception but also
to return such statuses encoded somehow in the return object.


adriancole:
generally, I don't like this signal as it can lead to NPE or other cases where exception handling no longer applies. I'd like a few others to say they like it before merging

adriancole:
didn't get any support in several months. closing until interest resumes

PR #492, opened 2016-11-10
Allow configuration of status codes that cause Ribbon retries.

JonathanO:
This reintroduces a behaviour present in the original Ribbon RestClient which allowed a 503 status code to be treated as a connection failure by Ribbon's retry mechanisms. Further, it makes it configurable, allowing you to specify which status codes are considered to be server throttled.

By default it preserves the current Feign ribbon-client behaviour.

I've added this feature since in our environment there's often a gateway in front of an application server which will return 502 or 503 if the application server instance is unavailable. These should be treated as retryable (but only on another server) as per the original RestClient behaviour.


ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -41,21 +44,34 @@
...
 
   public static LBClient create(ILoadBalancer lb, IClientConfig clientConfig) {
     return new LBClient(lb, clientConfig);
   }
 
+  private Set<Integer> parseStatusCodes(String statusCodesString) {
+    if (statusCodesString == null || statusCodesString.isEmpty()) {
+      return Collections.emptySet();
+    }
+    Set<Integer> codes = new HashSet<Integer>();
```
adriancole: nit always linkedhashset (prevents distraction in tests etc)

ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -41,21 +44,34 @@
   private final int connectTimeout;
   private final int readTimeout;
   private final IClientConfig clientConfig;
+  private final Set<Integer> serverThrottledStatusCodes;
 
   public static LBClient create(ILoadBalancer lb, IClientConfig clientConfig) {
     return new LBClient(lb, clientConfig);
   }
 
+  private Set<Integer> parseStatusCodes(String statusCodesString) {
```
adriancole: static? also package protected makes this testable

ribbon/src/main/java/feign/ribbon/LBClientFactory.java
```diff
@@ -12,7 +14,7 @@
   /**
    * Uses {@link ClientFactory} static factories from ribbon to create an LBClient.
    */
-  public static final class Default implements LBClientFactory {
```
adriancole: this is a noop as defaults to public static as member type of an interface?

ribbon/src/main/java/feign/ribbon/LBClientFactory.java
```diff
@@ -21,10 +23,18 @@ public LBClient create(String clientName) {
     }
   }
 
+  IClientConfigKey<String> ServerThrottledStatusCodes = new CommonClientConfigKey<String>("ServerThrottledStatusCodes") {};
```
adriancole: strange casing for constant..

JonathanO: Yes, though consistent with the similar statics in CommonClientConfigKey itself, so I thought their might be some weird convention about CommonClientConfigKey instances. If not, I'll change it to something more conventional.

ryanjbaxter:
Yes I do.  I was actually looking at an issue yesterday in Spring Cloud Netflix that was similar, in Cloud Foundry the router will return a `404` when an instance is unavailable.  Typically this is not considered an error by the HTTP client so no exception is thrown and no retry happens.  Maybe it is best it make it generic and have something like `retryableStatusCodes` and not just throttled status codes?

adriancole:
anything on @ryanjbaxter's last comment? in favor or against switching ServerThrottledStatusCodes to retryable ones?

ryanjbaxter:
FYI here is the change we make in Spring Cloud Netflix.  https://github.com/spring-cloud/spring-cloud-netflix/pull/1836

adriancole:
note: release in a couple days, so here's a chance!

JonathanO:
Hi, I've done the rename.

PR #493, opened 2016-11-10
Make Util.decodeOrDefault(...) public.

rfalke:
The method is used for Loggers which may not live in the feign package.


adriancole:
I don't really like feign's Util being used by other code because we can't change it whenever we like when that's the case. Ex we don't want to be guava. On the other hand, I've noticed we haven't ever marked this as internal, so in a way we are messy anyway. Moreover, this is the only non-public method, it seems. So, I guess it is ok.

Issue #498, opened 2016-12-04
Align Ribbon/Hystrix/Archaius versions


johnament:
Right now, Feign points to:

- Hystrix 1.4.26
- Ribbon 2.1.1

The problem is that Hystrix brings in Archaius 0.4.1 and Ribbon brings in 0.6.6.  Depending on the order you declare these dependencies in your dependency management tool, you may get one version of archaius or the other.  If you get the older one, Ribbon will not work as it requires access to some package scoped constructors.

To fix this, I'd propose manually bumping Archaius to 0.6.6 in the Hystrix pom.

kdavisk6:
This looks like has been taken care of, we should close this.

PR #499, opened 2016-12-04
Fixing Archaius version

johnament:
Fix for #498 Forcibly set the Archaius version to allow Ribbon/Hystrix to work in tandem.

### DIFF
--- a/hystrix/pom.xml
+++ b/hystrix/pom.xml
@@ -22,6 +22,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.netflix.archaius</groupId>
+      <artifactId>archaius-core</artifactId>
+      <version>0.6.6</version>
+    </dependency>
+
+    <dependency>
       <groupId>com.netflix.hystrix</groupId>
       <artifactId>hystrix-core</artifactId>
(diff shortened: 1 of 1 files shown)


adriancole:
maybe add a comment about what we should be careful about when considering the two versions? I'm not quite sure how to not break this moving forward, basically.

adriancole:
basically, make a disclaimer in the pom similar to the former issue you raise?

(ps issue links are better in the description than the subject line as they are more obv. also when raising and fixing the issue at the same time, just PR is fine.. oh and.. Thanks for your help!)

johnament:
@adriancole you're right, i knew I messed up something in the PR.  Fixed.

TBH, I'm not sure there's an easy way to fix this in general (e.g. never make it happen).

I can add a comment and maybe move the dependency up to the root pom to make sure its called out better.  The flipside, users of only hystrix wouldn't have even noticed the issue.  Or users may be less stubborn than I am and have it overridden locally already.

adriancole:
ps I'm ok with merging if @ryanjbaxter is. I've not been following the hystrix/ribbon world, lately

PR #502, opened 2016-12-05
Fix readme method name to produce client based in Target<T>

marceldiass:
Fix readme method name to produce client based in Target<T>

From 
```java
Feign feign = Feign.builder().build();
CloudDNS cloudDNS = feign.target(new CloudIdentityTarget<CloudDNS>(user, apiKey));
``` 
To 
```java
CloudDNS cloudDNS = Feign.builder().target(new CloudIdentityTarget<CloudDNS>(user, apiKey));
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,7 @@ Feign can produce multiple api interfaces.  These are defined as `Target<T>` (de
 For example, the following pattern might decorate each request with the current url and auth token from the identity service.
 
 ```java
-Feign feign = Feign.builder().build();
-CloudDNS cloudDNS = feign.newInstance(new CloudIdentityTarget<CloudDNS>(user, apiKey));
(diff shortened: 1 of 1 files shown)


spencergibb:
Why? That's a major rename that I'm not keen on doing

adriancole:
this is actually a doc bug. it is subtle because few use the newInstance method. Maybe it would be better to switch the reference to builder (as opposed to .build(); then .newInstance)

https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Feign.java#L92

marceldiass:
@spencergibb I've modified that because `target` method is not accessible from a Feign instance.

As @adriancole suggested I just the doc to `builder().target()` approach.

adriancole:
thanks @marceldiass!

Issue #510, opened 2016-12-27
Query params are sent as body instead of empty body due to wrong usage of Apache HttpClient


Infeligo:
I define my client like this:

```java
    @RequestLine("POST /test?query={query}")
    Foo test(@Param("query") String query);
```

I expect it to send empty body and "query" named parameter as part of query string. It does so when I use Feign's default client. However, it behaves differently when Apache HttpClient is used. The reason is in [these lines](https://github.com/OpenFeign/feign/blob/master/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java#L137) of ApacheHttpClient:

```java
if (request.body() != null) {
      ...
      requestBuilder.setEntity(entity);
}
```

There is no "else" part. Feign only sets entity when body is not null. However, Apache HttpClient's RequestBuilder [documentation ](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/methods/RequestBuilder.html)states the following:

> Please note that this class treats parameters differently depending on composition of the request: if the request has a content entity explicitly set with setEntity(org.apache.http.HttpEntity) or it is not an entity enclosing method (such as POST or PUT), parameters will be added to the query component of the request URI. Otherwise, parameters will be added as a URL encoded entity.

See also RequestBuilder's [source](https://apache.googlesource.com/httpclient/+/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/RequestBuilder.java#459) for exact implementation of this description.

spencergibb:
What would the `else` part be?

Infeligo:
I created a PR with a possible fix.

PR #511, opened 2016-12-28
Set empty HttpEntity if request body is null.

Infeligo:
This fixes the issue I raised in #510 where Feign did not account for Apache HttpClient behaviour difference in case `HttpEntity` was not set at all (e.g. due to null body in this case). This resulted in `RequestBuilder` creating `HttpEntity` from query parameters.

### DIFF
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -145,6 +145,8 @@ public final class ApacheHttpClient implements Client {
       }
 
       requestBuilder.setEntity(entity);
+    } else {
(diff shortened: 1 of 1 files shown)


spencergibb:
restarting build

adriancole:
test?

adriancole:
closing until interest resumes (specifically adding a test)

jonfreedman:
I've created a failing test, I introduced a test scope dependency on feign-jaxrs from feign-httpclient as that's specifically how we're using this, you may be able to re-write the test to simplify it.

adriancole:
thanks! closed via 431b3282378104029ebf5035d9316b6affff9526

PR #531, opened 2017-02-23
Allow delegate client to pass in RequestConfig params

safreiberg:


safreiberg:
Hey Feign team - 

For our use case we need to pass in a custom request configuration to allow things like circular redirects. When using the ApacheHttpClient, the request config on the delegate is replaced. This change would allow passing through the delegate's request config and augmenting it with the normal feign parameters.

Thanks!
Stephen

adriancole:
how does this relate to #554

kdavisk6:
This change was made by another PR.  Closing as a duplicate

PR #533, opened 2017-03-04
AnnotationBased Error decoder

saintf:
AnnotationBased Error decoder - includes tests/etc and updated documentation.

Allows for more interesting building of exceptions via annotations on the exception classes themselves.


saintf:
Build succeeded without the ribbon tests:
https://travis-ci.org/OpenFeign/feign/builds/207774848

Have enabled again but concerned they'll cause the whole thing to fail again...

annotation-error-decoder/README.md
```diff
@@ -0,0 +1,142 @@
...
+```java
+GitHub github = Feign.builder()
+                     .errorDecoder(
+                         AnnotationErrorDecoder.builderFor(GitHub.class).build()
+                     )
+                     .target(GitHub.class, "https://api.github.com");
+```
+
+##Leveraging the annotations and priority order
+For annotation decoding to work, the class must be annotated with @ErrorHandling` tags.
```
spencergibb: should be `@ErrorHandling` (missing opening backtick).

annotation-error-decoder/build.gradle
```diff
@@ -0,0 +1,11 @@
+apply plugin: 'java'
```
spencergibb: Why a `build.gradle`?

adriancole:
hmm only 2 in favor, and looks like it has no technical need to be in the core project. should we open a repo for this instead?

annotation-error-decoder/README.md
saintf: Done

annotation-error-decoder/build.gradle
saintf: I saw in multiple other modules that there were build.gradle files. I assumed (perhaps a mistake) that there was potentially a migration underway, and added the file to help out. I'm absolutely fine to delete.

saintf: And done :)

saintf:
@adriancole It's worth the question. I'd be keen on letting others contribute to it, and given the tight coupling with Feign itself, it felt like the right place was right next to the core modules. However, if there's a better spot that it can be next to Feign (and as Feign upgrades it ideally continues to upgrade with it) then I'm happy to put it there.

saintf:
@adriancole So looked over other repos under the OpenFeign project. Would you be ok to have it under a feign-annotation-error-decoder (or perhaps something smaller in terms of name) repo but still within the OpenFeign organisation?

adriancole:
Yep as long as you are ok championing and releasing it.

saintf:
@adriancole sign me up! As long as I continue to get code reviews (don't want to become a one-man show for it... and want to make sure I stay tightly in line with Feign!)

adriancole:
looking at this, I do think it is big enough to be on its own, and there seem to be a few interested.. I'll open a repo.

adriancole:
feign-annotation-error-decoder is long, but I can't think a better name.. will open that in the next day or two unless someone thinks of something better

adriancole:
ok I opened the repo. you can follow the leader on https://github.com/OpenFeign/feign-opentracing for things to do

PR #534, opened 2017-03-11
Interceptor decoder

odin-delrio:
Hi all!

I have an use case in which I have to interact with some HTTP API that is currently responding ciphered JSONs (I can't change this), so I need to first decipher that JSON before giving it to the Decoder (JacksonDecoder in my case).

To do this in a generic way I implemented an **InterceptorDecoder** that allows you to intercept the feign response before the desired decoder is called.

This also could be used to rewrite the response headers.
For example, okhttp allow to do that in a similar way: https://github.com/square/okhttp/wiki/Interceptors#rewriting-responses


spencergibb:
Why don't we wait to see if more people want something like this in core.

odin-delrio:
@spencergibb Cool! I just updated the description to reference headers rewriting ;)

adriancole:
tons of votes on this one (relatively speaking)

core/src/main/java/feign/codec/InterceptorDecoder.java
```diff
@@ -0,0 +1,26 @@
+package feign.codec;
+
+import feign.Response;
+
+import java.io.IOException;
+import java.lang.reflect.Type;
+
+public class InterceptorDecoder implements Decoder {
```
adriancole: ResponseMappingDecoder? Add some javadoc since this is a public type, and tighten up with finals etc?

README.md
```diff
@@ -187,6 +187,20 @@ GitHub github = Feign.builder()
                      .target(GitHub.class, "https://api.github.com");
 ```
 
+If you need to pre-process the response before give it to the Decoder, you can use the [InterceptorDecoder](/feign/codec/InterceptorDecoder.java).
+
+```java
+GitHub github = Feign.builder()
+                     .decoder(
+                         new InterceptorDecoder(
```
adriancole: tempted to make this stuff private and only expose the builder..

Ex. mapAndDecode((response, type) -> myTransformingLogic(response, type), new GsonDecoder())

thoughts?

adriancole:
can those voting mention what you want to use this for?

core/src/main/java/feign/codec/InterceptorDecoder.java
odin-delrio: ResponseMappingDecoder seems a good name for me, I can change it!
Also you're right about the final modifier to the properties and some javadoc, I'm on it ;)

README.md
odin-delrio: It is ok for me! But for make this Decoder private (at least package private) I have to move it from the ```feign.codec```package to the ```feign```one... is that ok for you?

adriancole: it could be a static package protected member of feign

adriancole: fyi we'll be doing a release in a couple days

odin-delrio: thanks for be patient!
I'm not sure that I understand what do you want to say with "static package protected member of feign"...
make the `ResponseMappingDecoder` an inner class insied the `feign.Feign` abstract class?

adriancole: yep that's what I meant. The users don't need to know about this type.

core/src/main/java/feign/Feign.java
```diff
@@ -219,4 +228,27 @@ public Feign build() {
...
+    @Override
+    public Object decode(Response response, Type type) throws IOException {
+      return delegate.decode(mapper.map(response, type), type);
+    }
+  }
+
+  /**
+   * Map function to apply to the response before decoding it.
+   */
+  public interface ResponseMapper {
```
adriancole: make this part a top-level type.

Also, please add an example in the javadoc.

ex. 
```
<pre>{@code
// your compelling example here:
}</pre>
```

README.md
```diff
@@ -187,6 +187,14 @@ GitHub github = Feign.builder()
                      .target(GitHub.class, "https://api.github.com");
 ```
 
+If you need to pre-process the response before give it to the Decoder, you can use the `mapAndDecode` builder method:
+
+```java
+GitHub github = Feign.builder()
+                     .mapAndDecode((response, type) -> myTransformingLogic(response, type), new GsonDecoder())
```
adriancole: myTransformingLogic is coolio, but there's probably a better hint. Ex what made you want this in the first place? is there a concrete, but easy to understand example? Notably, the map happens *before* the decoding, so something like unwrapping a nested json envelope doesn't seem to be realistic.

Let's make this easy to understand!

ps.. leave out the github thing as their api is already nice :)

odin-delrio: I had 2 use cases for this, ciphered JSONS and JSONP responses...
I explained the second one because I think that could be more usual.

Issue #470, opened 2016-09-30
Issue on QueryMap when the value start with '{'


seno5979:
I'm facing issue when using queryMap when value starting with '{' like below example :

```
publiic interface TestService {
    @RequestLine("GET /test")
    Country getCountry(@QueryMap Map queryMap);
}
```

The encodeIfNotVariable logic made the encode step be skipped because the query map value starting with '{'
https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/RequestTemplate.java#L365

spencergibb:
`{` is how variables are distinguished. Are you using that without variables?

seno5979:
Hi Spencer,

I have a case to call solr service which one of the value is in json format. 

Currently as a workaround, I need to encode the value by myself before put to the queryMap and disable the auto encoding by @QueryMap(encoding=true)

adriancole:
personally, I think this is a reasonable workaround as json is an odd thing to put in the request line.

note there's also this option which is another way to accomplish the same #465

seno5979:
Hi Adrian,

By your comment above, can I say that when using feign we are not allowed to have value start with '{' ? (which default @QueryMap encoded false)

In my case, I'm creating a feign client to call a SOLR service. Below is the documentation which is in some cases, the query value need to send a **curly bracket**

[https://cwiki.apache.org/confluence/display/solr/The+Standard+Query+Parser](https://cwiki.apache.org/confluence/display/solr/The+Standard+Query+Parser)

Below is the exception I received before disable the auto encoded in QueryMap(@QueryMap(encoded=true)) :

```
Caused by: java.net.URISyntaxException: Illegal character in query at index 106: http://solrServiceClientProductSearchCloudRestCF/solr/SYS_4741549800/select?q=swei&df=keywordSearch_en&fq={!tag=categoryName}(categoryName:"Awesome I")&sort=score+desc&start=0&rows=24&fl=productLocaleKey%2Cscore&preferLocalShards=true&spellcheck=true&spellcheck.collate=true&facet=true&facet.field={!ex=categoryName}categoryName&f.categoryName.facet.sort=alpha&f.categoryName.facet.mincount=1&f.categoryName.facet.limit=-1
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3111)
[...]

PR #540, opened 2017-03-23
fix gh-470, support the values of QueryMap starting with '{'

lowzj:
see https://github.com/OpenFeign/feign/issues/470#issuecomment-286992065

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -83,7 +83,7 @@ public final class RequestTemplate implements Serializable {
     }
   }
 
-  private static String urlEncode(Object arg) {
+  static String urlEncode(Object arg) {
     try {
       return URLEncoder.encode(String.valueOf(arg), UTF_8.name());
     } catch (UnsupportedEncodingException e) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -75,15 +75,15 @@ public RequestTemplate(RequestTemplate toCopy) {
     this.decodeSlash = toCopy.decodeSlash;
   }
 
-  private static String urlDecode(String arg) {
+  public static String urlDecode(String arg) {
```
adriancole: these are in the same package, so package private would do

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -75,15 +75,15 @@ public RequestTemplate(RequestTemplate toCopy) {
...
     try {
       return URLDecoder.decode(arg, UTF_8.name());
     } catch (UnsupportedEncodingException e) {
       throw new RuntimeException(e);
     }
   }
 
-  private static String urlEncode(Object arg) {
+  public static String urlEncode(Object arg) {
```
adriancole: same here

adriancole: actually this can remain private as it isn't used in this PR

adriancole: make package private

core/src/main/java/feign/RequestTemplate.java
adriancole: revert as this code isn't used in this PR

Issue #542, opened 2017-03-27
Support Multiple Collection Formats


bbdouglas:
I recently ran into a compatibility issue of representing a query parameter with multiple values. The server I was talking to expected a comma-separated list of values, while feign produced repeated parameters.

    /?foo=bar&foo=baz  // This is what was sent
    /?foo=bar,baz      // This is what was expected

It seems there is no single standard for which form to use. I believe JAX-RS uses repeated parameters, while [Swagger](http://swagger.io/specification/#fixed-fields-45) defaults to csv. This issue has come up before with the [Spring Data REST Endpoint](https://github.com/spring-cloud/spring-cloud-netflix/issues/1634), which had a workaround of adding a RequestInterceptor.

One possible change that would clarify and expand the behavior would be to allow setting the collection format type in the [@RequestLine](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/RequestLine.java) annotation. It could look something like this:

    public @interface RequestLine {
      String value();
      boolean decodeSlash() default true;
      // new parameter for declaring how collections should be encoded
      String collectionFormat() default "multi";  // or "csv", "ssv", "tsv", "pipes" a la Swagger
    }

I believe this is a cleaner solution than custom RequestInterceptor code as it is a short declarative statement as opposed to boilerplate code that would need to be copied for each new service added. It also documents the default behavior, which has been unclear to users in the past (see #398). The name of the new parameter and the various options in this case are inspired by [Swagger](http://swagger.io/specification/#fixed-fields-45). The default value would be the current behavior, so backwards compatibility would be maintained. The implementation should be easy and I can submit a pull request if there interest in this approach.

Please let me know how you feel about this idea.

spencergibb:
I think the issue with an option on request line is extensibility. How would you add formats?

bbdouglas:
I understand your concern with extensibility. How about this for an alternative: extend the Param.Expander class to cover concatenation of elements in a collection?

The current Param.Expander class already allows you to format individual elements in a collection, but can't be used to customize how they are concatenated. What if another method was added to Param.Expander called `expandCollection` that accepted a `Collection<?>` instead of an `Object`? Implementing the current behavior (repeating the parameter name for each element) becomes a little trickier since the expander needs access to the parameter name. I could imagine a few ways around this, such as returning `null` for the default behavior or additionally passing in the parameter name.

Kindrat:
[...]

PR #543, opened 2017-03-27
Allows different collection encodings

bbdouglas:
In the case where a parameter represents a collection of values, there are
conflicting ways of encoding that collection. Common ways are repeating the
parameter name (foo=bar&foo=baz) and using comma separated values (foo=bar,baz).
The current behavior repeats the parameter name. This change introduces an
additional RequestLine parameter that explicitly specifies the encoding type,
one of CSV, TSV, space-delimited, pipe-delimited, and repeating the parameter
name. The default value for this option is repeating the parameter name, so
backwards compatibility is maintained.

Possible implementation for #542.

### DIFF
--- a/core/src/main/java/feign/CollectionFormat.java
+++ b/core/src/main/java/feign/CollectionFormat.java
@@ -1,20 +1,85 @@
+/**
+ * Copyright 2012-2018 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package feign;
+
+import java.util.Collection;
 
 /**
  * Various ways to encode collections in URL parameters.
  *
- * These specific cases are inspired by the
- * <a href="http://swagger.io/specification/">OpenAPI specification</a>.
+ * <p>These specific cases are inspired by the
+ * <a href="http://swagger.io/specification/">OpenAPI specification</a>.</p>
  */
 public enum CollectionFormat {
   /** Comma separated values, eg foo=bar,baz */
-  CSV,
+  CSV(","),
   /** Space separated values, eg foo=bar baz */
-  SSV,
+  SSV(" "),
   /** Tab separated values, eg foo=bar[tab]baz */
-  TSV,
+  TSV("\t"),
   /** Values separated with the pipe (|) character, eg foo=bar|baz */
-  PIPES,
+  PIPES("|"),
   /** Parameter name repeated for each value, eg foo=bar&foo=baz */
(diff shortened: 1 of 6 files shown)


adriancole:
thanks for raising. per the HACKING file, we don't tend to add features ahead of multiple people requesting them. If we did, an enum might be a bit brittle or "guessy" as in we are guessing which formats people might want, and locking them into the old approach if not. Currently, we have Param.Extractor which allows people to implement their favorite toString'er for a parameter (memory fails me, but it might be possible to use this in lieu of this change). At any rate, something like that is more likely to reduce the amount of similar change requests.

If you want to proceed with something like this, review the issues and see if you can find similar ones. We're looking for at least a few users as code like this usually lingers past the interest of the author (no offense)

bbdouglas:
Hi Adrian,

Thank you for reviewing my issue. I understand about your concern about
feature-creep.

I was able to find one other user who ran into the same issue as me (
https://github.com/spring-cloud/spring-cloud-netflix/issues/1634). He was
able to use a RequestInterceptor as a workaround. I could do that, too, but
for my use case (auto-generating client stubs from Swagger specs) the
annotation route is cleaner.

In any case, let's see if anyone else shows interest in that topic.

Thanks,
Ben

rage-shadowman:
How many votes does it take? And what do you consider votes? I'm assuming the thumbs-up icon is a vote for yes?

core/src/main/java/feign/CollectionFormat.java
```diff
@@ -0,0 +1,20 @@
...
+  /** Comma separated values, eg foo=bar,baz */
+  CSV,
+  /** Space separated values, eg foo=bar baz */
+  SSV,
+  /** Tab separated values, eg foo=bar[tab]baz */
+  TSV,
+  /** Values separated with the pipe (|) character, eg foo=bar|baz */
+  PIPES,
+  /** Parameter name repeated for each value, eg foo=bar&foo=baz */
+  MULTI,
```
rage-shadowman: I would either add a package-local method in here for `abstract String getSeparator(String field);` rather than a `switch` statement in `RequestTemplate`, or I'd drop the enum in favor of a `ParamJoiner` interface (which again has the `String getSeparator(String field)` method).

velo:
I think after conflict being solved and the switch being removed, we can merge this.

core/src/main/java/feign/CollectionFormat.java
kdavisk6: I can use cases for CSV, SSV, PIPES and MULTI.  I am concerned with TAB.  Is the expectation that when that is used a TAB character is present on the query string?  Is that considered safe?

kdavisk6: Awesome.  Thanks for the info!  Good to know.

bbdouglas: @rage-shadowman @kdavisk6  I went ahead and refactored this a bit according to your suggestions, and I think it's a bit cleaner now. Instead of the switch statement, I added a `join` method to the enum that contains all the logic for joining together collections of values with the given field name. Now the joining logic in `RequestTemplate` is minimal.

I also renamed MULTI to EXPLODED, which is the term used in the newest edition of OpenAPI. Users will generally not see this value anyway, since that's the default.

bbdouglas:
@velo I rebased to master and fixed the merge conflict. I also removed the switch statement in favor of moving the joining logic into the enum class.

PR #547, opened 2017-04-02
More compact version of Slf4jLogger

lanwen:
Writes logs in format:

```
DEBUG feign.Logger - http	req-id=[701d8882-e43b-4af3-b5e5-7dc4b4cf4de7]   call=[someMethod()]	method=[GET]	uri=[http://api.example.com]
DEBUG feign.Logger - http	req-id=[701d8882-e43b-4af3-b5e5-7dc4b4cf4de7]	status=[200]	reason=[OK] elapsed-ms=[273]	length=[0]

```

Then you can grep such logs with `| cut -f 3` (with right column number)

Also this logger adds unique `req-id` string to merge request and response.
All retries and error log lines have such string too.
This string is changed each request.

### DIFF
--- a/slf4j/README.md
+++ b/slf4j/README.md
@@ -10,3 +10,28 @@ GitHub github = Feign.builder()
                      .logger(new Slf4jLogger())
                      .target(GitHub.class, "https://api.github.com");
 ```
+
+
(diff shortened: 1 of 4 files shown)


lanwen:
Will make it as separate library

Issue #294, opened 2015-11-03
OptionalAwareDecoder in feign core?


uschi2000:
Hej again,
How are the chances of getting a "feign-guava-optional" module (or similar) into feign that knows about 404-decoding Guava Optional? Such a module would have a Guava dependency and only make sense where decode404 is enabled.
  R.

```
/**
 * Decorates a Feign {@link Decoder} such that it returns {@link Optional#absent} when observing an HTTP 404 error code
 * for a method with {@link Type} {@link Optional}. Propagates the exception returned by the given {@link ErrorDecoder}
 * in case the response is 404 for a non-Optional method.
 */
public final class OptionalAwareDecoder implements Decoder {

    private final Decoder delegate;
    private final ErrorDecoder errorDecoder;

    public OptionalAwareDecoder(Decoder delegate, ErrorDecoder errorDecoder) {
        this.delegate = delegate;
        this.errorDecoder = errorDecoder;
    }

    @Override
    public Object decode(Response response, Type type) throws IOException, FeignException {
        if (response.status() == 404) {
            if (Types.getRawType(type).equals(Optional.class)) {
                return Optional.absent();
            } else {
                throw Throwables.propagate(errorDecoder.decode(null, response));
            }
        } else {
            return delegate.decode(response, type);
        }
    }
}
```

adriancole:
Thanks for the idea. First start would be finding those who'd also use
this. Let's see if this issue gets +1s. In the mean time, thanks for
pasting the solution.

greyfairer:
+1 for me, allthough I need support for Java 8 Optional, too.

ilaborie:
+1
but it's require a JDK 8

spencergibb:
@adriancole this would be good as a separate project when we move feign to an org.

adriancole:
this is probably doable with some trickery on the 8.x branch since Optional is a type, not a language feature cc @oillio

amb-it-ion:
+1
I had enough NullPointerExceptions for this life.

edio:
We have similar thing in our project too

adriancole:
Boy I wish it actually made NPEs disappear! nothing more vexing than a
null Option pointer. :D

anyway...

OptionalAwareDecoder can be written as a decorator, which can wrap
something like jackson, and handle the 404 case. It can be done for
java 8, guava you name it in very few lines of code.

If optional means escalating into the decoders themselves, that would
be more work.

[...]

PR #548, opened 2017-04-02
introduce OptionalDecoder

arielsegura:
Addresses #294


feign-optional-decoder/src/main/java/feign/optionals/OptionalDecoder.java
```diff
@@ -0,0 +1,38 @@
...
+import feign.Response;
+import feign.Util;
+import feign.codec.Decoder;
+
+import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Optional;
+
+public class OptionalDecoder implements Decoder {
```
adriancole: final

feign-optional-decoder/src/main/java/feign/optionals/OptionalDecoder.java
```diff
@@ -0,0 +1,38 @@
...
+    final Decoder delegate;
+
+    public OptionalDecoder(Decoder delegate) {
+        this.delegate = delegate;
```
adriancole: check null

arielsegura:
@adriancole done!

spencergibb:
I like `feign-java8` as well

arielsegura:
Naming it as feign-java8 will imply that further features around java8 will have to be shipped in this artefact, so for instance if one just wants to use Optional stuff, will bring the whole java8 support.

arielsegura:
Fair enough 👍

adriancole:
if you can rejig in the next couple days, will go in the next release. Don't forget CHANGES file
cc @jfuerth

arielsegura:
Done!

adriancole:
mind renaming the dir to java8 (pom artifact is good, just this is more conventional with others)

arielsegura:
@adriancole renamed

PR #549, opened 2017-04-03
Avoid `Decoder.decode` on 404 when void response type

### LINKED ISSUE #518: `Decoder.decode` being called for `void` response type for `decode404`
404 errors will be silently swallowed if the method has a return type of `void` and `decode404` is set.
PR #72 stops `Decoder` being called for void types for 2xx response codes but there should be similar protection against calling `Decoder` for 404 errors.

In short, this line [1] should read `} else if (decode404 && response.status() == 404 && void.class != metadata.returnType()) {`.

[1] https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/SynchronousMethodHandler.java#L135

wjam:
Avoid calling `decode` when receiving a 404 error and `decode404` is set and the response type is `void`.

This PR fixes #518.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 9.4.1
+* 404 responses are no longer swallowed for `void` return types.
+
 ### Version 9.4
 * Adds Builder class to JAXBDecoder for disabling namespace-awareness (defaults to true).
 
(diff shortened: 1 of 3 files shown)


adriancole:
This might break people, so add something to CHANGES like..

404 response is no longer swallowed for void return types

wjam:
Update `CHANGELOG.md`

adriancole:
thanks @wjam

Issue #553, opened 2017-04-20
ApacheHttpClient hardcodes RequestConfig


andysworkshop:
Please can you not [hardcode the RequestConfig](https://github.com/OpenFeign/feign/blob/master/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java#L96) if the ApacheHttpClient has been supplied externally in the constructor?

The RequestConfig contains some critical stuff, such as the HTTP proxy configuration, that you are explicitly denying access to by not allowing us to customize the request config.

It would be sufficient to not specify a RequestConfig at all since you can assume that if we have taken control by specifying an apache http client in the constructor then we have set up appropriate request configuration of our own.

adriancole:
Do you have a suggestion (preferably as a pull request) that does what you
feel is better but doesnt impact existing users?

andysworkshop:
Hi Adrian, yes I do. I'll try to put together a PR this weekend. Can't do it from work unfortunately.

kdavisk6:
This is something that we should really add to the documentation.  Closing as we have a few newer issues related to the same issue #781

PR #554, opened 2017-04-23
Permit use of default RequestConfig

andysworkshop:
This feature adds the ability for a user who is supplying their own custom `HttpClient` to have their default `RequestConfig` honored instead of being overriden. This permits the user to set up important stuff like the HTTP proxy configuration for an enterprise environment. See issue #553.

I hope you don't mind the addition of the mockito dependency (and its requirement for hamcrest-core) at test scope so I can properly test the enhancement.

Also made the class non-final so that users can subclass and override where further customisation is necessary and also so the unit tests can subclass and override the creation of the `RequestBuilder` to insert the mock.

### DIFF
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -100,7 +100,7 @@ public class ApacheHttpClientTest extends AbstractClientTest {
 
 
   @Test
-  public void shouldUseDefaultRequestConfigDefaultClient() throws Exception {
+  public void shouldUseDefaultRequestConfigUserClient() throws Exception {
 
     this.apacheHttpClient = new ApacheHttpClient(this.httpClient, true) {
       @Override
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -13,22 +13,19 @@
  */
 package feign.httpclient;
 
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.HashMap;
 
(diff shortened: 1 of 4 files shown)


adriancole:
I think the key change here is that we are choosing to not honor Request options provided by feign. Maybe we can make that property more explicit (either by documentation or otherwise).

I'm not a fan of making this not final as it wasn't designed for extension and most of our other stuff is sealed. When we support extension, it would likely imply making some methods not-final, else we will be opening up for bugs or lock-in to apis that are accidentally protected.

I think there's another way to test, which is via side effects. Either you could test that request options are ignored, or some property removed when we override config is not overridden.

andysworkshop:
PR updated based on feedback.

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -1,31 +1,99 @@
 /*
  * Copyright 2015 Netflix, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
```
spencergibb: Please don't reformat when it's not part of the actual changes.

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -1,31 +1,99 @@
...
+
+
+  @Test
+  public void shouldUseEmbeddedRequestConfigDefaultClient() throws Exception {
+
+    ApacheHttpClient apacheHttpClient = new ApacheHttpClient();
+    HttpRequestBase httpRequest =
+        (HttpRequestBase) apacheHttpClient.toHttpUriRequest(this.request, this.options);
+
+    assertNotNull(httpRequest.getConfig());
```
spencergibb: Shouldn't the tests do more than a null check?

adriancole: yeah I think this is a bit confusing, particularly as we are unit testing a very specific method.

let's test side effects? ex show that if you pass true (as the new option) that request.options are ignored.
then show that if you don't, some realistic user config that exist is honored.

adriancole:
how does this relate to #531?

adriancole:
ping

kdavisk6:
ping timeout

PR #555, opened 2017-04-25
Fix the issue encountered when the value of queries starting with '{'

Echo19890615:
When using RibbonClient, the starting `%7B` in the value of queries would be replaced with `{`. For example, a request URL like `http://example.com/q?p=%7Baaa%7D` would become `http://example.com/q?p={aaa}` after passing `toRequest()` in feign.ribbon.LBClient.

This issue could make feign not able to play with newer versions of Tomcat after fixing CVE-2016-6816.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -15,14 +15,6 @@
  */
 package feign;
 
-import static feign.Util.CONTENT_LENGTH;
-import static feign.Util.UTF_8;
-import static feign.Util.checkArgument;
-import static feign.Util.checkNotNull;
-import static feign.Util.emptyToNull;
-import static feign.Util.toArray;
-import static feign.Util.valuesOrEmpty;
-
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
@@ -37,6 +29,14 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+
+import static feign.Util.CONTENT_LENGTH;
+import static feign.Util.UTF_8;
+import static feign.Util.checkArgument;
+import static feign.Util.checkNotNull;
+import static feign.Util.emptyToNull;
+import static feign.Util.toArray;
+import static feign.Util.valuesOrEmpty;
 
 /**
  * Builds a request to an http target. Not thread safe. <br> <br><br><b>relationship to JAXRS
@@ -308,17 +308,6 @@ public final class RequestTemplate implements Serializable {
     return this;
   }
 
-  /**
-   * Append the specified {@code value} to {@code url} without decoding or encoding. Append all 
-   * the queries into {@code queries} without decoding or encoding if {@code url} already has queries. 
-   * <br/> <b>Note:</b> The difference between {@link #append(CharSequence value)} and 
-   * {@link appendDirectly(CharSequence value)} is: {@link #append(CharSequence value)} decodes 
-   * the {@code value} then encodes it(encode can see here {@link #encodeIfNotVariable(String)}).
-   * {@link appendDirectly(CharSequence value)} processes the {@code value} directly, which means 
-   * it does not decode or encode at all. <br/>{@link RibbonRequest.toRequest()} should use this method.
-   * @param value the value to append to url/queries
-   * @see #url()
-   */
   public RequestTemplate appendDirectly(CharSequence value) {
 	url.append(value);
 
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -15,6 +15,14 @@
  */
 package feign;
 
+import static feign.Util.CONTENT_LENGTH;
+import static feign.Util.UTF_8;
+import static feign.Util.checkArgument;
+import static feign.Util.checkNotNull;
+import static feign.Util.emptyToNull;
+import static feign.Util.toArray;
+import static feign.Util.valuesOrEmpty;
+
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
@@ -29,14 +37,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-
-import static feign.Util.CONTENT_LENGTH;
-import static feign.Util.UTF_8;
-import static feign.Util.checkArgument;
-import static feign.Util.checkNotNull;
-import static feign.Util.emptyToNull;
-import static feign.Util.toArray;
-import static feign.Util.valuesOrEmpty;
 
 /**
  * Builds a request to an http target. Not thread safe. <br> <br><br><b>relationship to JAXRS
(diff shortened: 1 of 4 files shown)


adriancole:
I feel like this is yuk. Yuk that it requires so much care, but this is
sort of par for the course.

Since adding a public api, probably good to javadoc when you need this.
Also note in comment where used. If there is a way to make ribbon break
without. Test there would be nice too

fengchj:
i am puzzled, is this a bug?

Echo19890615:
Hi, adriancole

IMHO this is a serious issue. That makes a request with queries whose value is a json string have chance to get a response with code 400 since the request line contains "unwise" characters referring to RFC 2396. We want to use Ribbon in our project & are really appreciate that Feign can reduce our work. So we do this to make Feign suitable for wider use.

`RequestTemplate.appendDirectly()` is added just for RibbonClient & there is no other place to call, therefore there is no side effect to be seen.

adriancole:
just an aside, I think there's other types of things that will probably break similarly.. iirc there was some outstanding work due to elasticsearch's query api as well.

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -291,6 +308,33 @@ public RequestTemplate append(CharSequence value) {
     return this;
   }
 
+  /**
+   * Append the specified {@code value} to {@code url} without decoding or encoding. Append all
```
adriancole: lots of doc. but no example. forward referencing to ribbon probably won't help people understand.

Try to shorten this to maybe two succinct sentences and an example.

Ex. 

```
Like {@link #foo} except ..

<p> <-- note p not br and the newline break makes it easy to read, too..
ex if your input is X unless you call this method, it would become Y
```

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -30,14 +38,6 @@
 import java.util.Map;
 import java.util.Map.Entry;
 
-import static feign.Util.CONTENT_LENGTH;
```
adriancole: revert the re-order

core/src/test/java/feign/RequestTemplateTest.java
```diff
@@ -362,4 +362,18 @@ public void encodedQueryWithUnsafeCharactersMixedWithUnencoded() throws Exceptio
     assertThat(template.queries()).doesNotContain(entry("params[]", asList("not encoded")));
     assertThat(template.queries()).contains(entry("params[]", asList("encoded")));
   }
+  
+  @Test
+  public void appendDirectlytest() throws Exception {
+    String url = "http://test.feign.com/query?params=%7b%22int%22%3a1%2c%22string%22%3a%22str%22%7d";
```
adriancole: comment why we care. ex people won't necessarily be able to understand that this is escaped json (if it is)

ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -96,7 +96,7 @@ public RequestSpecificRetryHandler getRequestSpecificRetryHandler(
     Request toRequest() {
       return new RequestTemplate()
           .method(request.method())
-          .append(getUri().toASCIIString())
+          .appendDirectly(getUri().toASCIIString())
```
adriancole: since this is only added for ribbon, there should be a test here.

adriancole: wait.. couldn't this be much easier by just making a request?

adriancole: like this..

```java
    Request toRequest() {
      return Request.create(
          request.method(),
          getUri().toASCIIString(),
          request.headers(),
          request.body(), request.charset()
      );
    }
```

Echo19890615: the junit for RibbonRequest.toRequest() has been added, please review

Echo19890615: Using Request.create() is exactly easier to make an orignal request, thanks for reminding :) . But the old code(use append) has the logic that add the header "Content-Length" to the request. Base on keeping the old logic and code reuse, should we continue to use appendDirectly()?

adriancole:
I think based on how confusing the alternative is, and that it requires a
new public api to implement it.. stick with Request.create until that
becomes a problem. That way, this can be a patch release

Echo19890615:
1. revert all the codes of RequestTemplate.appendDirectly() by using 'git revert'
2. keep the junit test for RibbonRequest.toRequest()
3. fix the issue by using Request.create(), and keeping the old logic(add the header 'Content-Length' manually).

ribbon/src/main/java/feign/ribbon/LBClient.java
```diff
@@ -94,12 +97,14 @@ public RequestSpecificRetryHandler getRequestSpecificRetryHandler(
...
-          .method(request.method())
-          .append(getUri().toASCIIString())
-          .headers(request.headers())
-          .body(request.body(), request.charset())
-          .request();
+      // add header "Content-Length" according to the request body
+      final byte[] body = request.body();
+      final int bodyLength = body != null ? body.length : 0;
+      // create a new Map to avoid side effect, not to change the old headers
+      Map<String, Collection<String>> headers = new HashMap<String, Collection<String>>();
```
adriancole: please use LinkedHashMap as this ensures tests don't flake on iteration order

ribbon/src/test/java/feign/ribbon/LBClientTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    // test the same attributes
+    assertEquals(requestOrign.method(), requestGetFormToRequest.method());
+    assertEquals(requestOrign.url(), requestGetFormToRequest.url());
+    assertEquals(urlWithEncodedJson, requestGetFormToRequest.url()); // the url in Request should not been encoded or decoded
+    assertEquals(requestOrign.body(), requestGetFormToRequest.body());
+    assertEquals(requestOrign.charset(), requestGetFormToRequest.charset());
+    assertNotNull(requestOrign.headers());
+    assertEquals(0, requestOrign.headers().size()); // double check the baseline Request
+    assertNotNull(requestGetFormToRequest.headers()); // double check the baseline Request, header is not null, but empty
```
adriancole: avoid checks that are duplicated from constructor  (I think it isn't possible that this is null)

ribbon/src/test/java/feign/ribbon/LBClientTest.java
```diff
@@ -0,0 +1,56 @@
...
+    Map<String, Collection<String>> headers = new HashMap<String, Collection<String>>();
+    // create a Request as baseline for recreate and assert
+    Request requestOrign = Request.create("GET", uri.toASCIIString(), headers, null, Charset.forName("utf-8")); 
+    RibbonRequest ribbonRequest = new RibbonRequest(null, requestOrign, uri);
+
+    // use toRequest() recreate a Request
+    Request requestGetFormToRequest = ribbonRequest.toRequest();
+
+    // test the same attributes
+    assertEquals(requestOrign.method(), requestGetFormToRequest.method());
```
adriancole: this test is kindof hard to read. Can you look at other unit tests? you'll notice we use assertj which is helpful

adriancole:
LGTM let me know if you are ready for merge

Echo19890615:
Of course! Feel free to merge it.

PR #556, opened 2017-05-01
Don't leak OkHttp response when response.body() is null

jfuerth:
This fixes an issue where OkHttp3 logs a message like

```
A connection to https://your-api-here/ was leaked. Did you forget to close a response body?
To see where this was allocated, set the OkHttpClient logger level to FINE:
Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```

when the underlying OkHttp request returns a zero-length body.

### DIFF
--- a/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
+++ b/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
@@ -25,7 +25,6 @@ import okhttp3.ResponseBody;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -106,6 +105,9 @@ public final class OkHttpClient implements Client {
 
   private static feign.Response.Body toBody(final ResponseBody input) throws IOException {
     if (input == null || input.contentLength() == 0) {
+      if (input != null) {
+        input.close();
+      }
       return null;
     }
     final Integer length = input.contentLength() >= 0 && input.contentLength() <= Integer.MAX_VALUE ?


okhttp/src/main/java/feign/okhttp/OkHttpClient.java
```diff
@@ -106,6 +105,9 @@ static Request toOkHttpRequest(feign.Request input) {
 
   private static feign.Response.Body toBody(final ResponseBody input) throws IOException {
     if (input == null || input.contentLength() == 0) {
+      if (input != null) {
+        input.close();
```
spencergibb: Seems odd. `input.close()` is in `close()` a few lines below. Why isn't that being called? What happens if `close()` is called before the input stream is read?

jfuerth: Hi Spencer!

It's apparently always okay to call close() on the body, even if you haven't read it. In this case, it should be doubly okay since we already checked that the content-length is 0 so there's nothing to read anyway.

We inserted the code here because on the very next line the toBody() method returns null, so callers have no way to access the unclosed ResponseBody object. OkHttp3 requires that every ResponseBody object is closed, regardless of whether or not the body has any contents.

spencergibb: Thanks for the update. LGTM

adriancole:
nice

jfuerth:
Sorry to be a bother, but when do you expect the next point release will happen? We're trying to decide if we should publish our fork to our maven repo (and then have to update all our services twice) or wait for the next OpenFeign release (and only sweep through and do the update once).

Thanks!

adriancole:
will cut one this weekend!

jfuerth:
Amazing, thanks so much!

PR #567, opened 2017-06-16
Perform Map validation of @QueryMap and @HeaderMap lazily

EthanLozano:
This allows an InvocationHandlerFactory to convert a parameter to a
Map<String, Object> after the method is invoked. This is useful for
parameter expansion.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -117,16 +117,6 @@ public interface Contract {
         }
       }
 
-      if (data.headerMapIndex() != null) {
-        checkState(Map.class.isAssignableFrom(parameterTypes[data.headerMapIndex()]),
-                "HeaderMap parameter must be a Map: %s", parameterTypes[data.headerMapIndex()]);
-      }
-
-      if (data.queryMapIndex() != null) {
-        checkState(Map.class.isAssignableFrom(parameterTypes[data.queryMapIndex()]),
-                "QueryMap parameter must be a Map: %s", parameterTypes[data.queryMapIndex()]);
-      }
-
       return data;
     }
 
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -214,11 +214,11 @@ public class ReflectiveFeign extends Feign {
       if (metadata.queryMapIndex() != null) {
         // add query map parameters after initial resolve so that they take
         // precedence over any predefined values
-        template = addQueryMapQueryParameters(argv, template);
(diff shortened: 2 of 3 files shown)


EthanLozano:
Here is an example where this is useful: https://github.com/Qualys/feign-jaxrs/issues/2#issuecomment-308914191
In that example, I perform parameter expansion of a `@BeanParam` object by using a `InvocationHandlerFactory`. So, the object is initially a POJO, but the `InvocationHandlerFactory` converts the POJO to a `Map<String, Object>`

adriancole:
Hmm point of contract is to do as much as possible up front, so that there
are less runtime exceptions. Is there no other way to accomplish what you
want?

adriancole:
Theres also a performance reason to not do later (on each request) what can
be done once. Maybe think about how to customize the contract vs changing
the default?

adriancole:
looking at BeanParam.. it is wider than query parameters, so doing a
hack here would result in other small breaks, right? BeanParam can
hold a lot of things, not just queries.. it more similar to a body
param in feign

http://docs.oracle.com/javaee/7/api/javax/ws/rs/BeanParam.html

Bottom-line is try to focus this in a way that existing code isn't
made loose or has more runtime checks as a result. a custom contract
might be the way out on this.

EthanLozano:
Yeah, I'm not sold on this change either: it is nice to fail fast in a Contract. At the very least, it unifies the type validation into a single location.

My goal is to expand a `@BeanParam` POJO into multiple query params (while also having a body object).

So, I want to support the following method signature:
`ResultObject run(@BeanParam BeanObject bean, BodyObject body);`

Where `BeanObject` might look like this:

    class BeanObject {
        @QueryParam("one") int one;
        @QueryParam("two") String two;
    }

As far as I can tell, `MethodMetadata`'s only way to convert a single index to multiple elements is through headerMapIndex or queryMapIndex.

The other option I considered was to write a Contract that fudges the index of each element by flattening the POJO. So, in the example above I could have:
"one" = 0
"two" = 1 (split out into a different index)
body = 2 (normally would be 1)

Then, I would also have to write a corresponding `InvocationHandlerFactory` that flattened POJOs into an array. But, this approach doesn't let me re-use Contract.BaseContract, which relies heavily on paramIndex.

adriancole:
Looking at this, I think the only way to support BeanParam neatly is to
have index-specific encoders..

adriancole:
of course it is complicated because in beanparam they can collaborate on forms..

EthanLozano:
I'm closing this since I agree that more validation should be done in the Contract. Also, I think the appropriate fix is to have index-specific encoders

PR #569, opened 2017-06-23
Add the Content-Type if specified, unless we are setting the body

### LINKED ISSUE #391: Feign OkhttpClient doesn't add Content-Type as header when body is null for GET request
I am facing issue with feign.OkHttpClient. The problem is if the request is of GET type and request body is null, then Content-Type header is not added in the requestBody and also not in headers.
<strike>This creates a problem where server accepts only spectific type of Content.</strike>
This creates a problem when Server consumes and/or accepts specific type of content.

class is feign.okhttp.OkHttpClient.java line number 66 and 89

DavidTanner:
Fixes #391 
If we specify the Content-Type we should send it, even without a body. 

The Apache client does the same.
https://github.com/OpenFeign/feign/blob/release-9.5.0/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java#L127-L129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 9.5.1
+* Update Okhttp client so that if specified, the content-type is included even without a body.
+
 ### Version 9.5
 * Introduces `feign-java8` with support for `java.util.Optional`
 * Adds `Feign.Builder.mapAndDecode()` to allow response preprocessing before decoding it.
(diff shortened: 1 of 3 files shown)


DavidTanner:
@adriancole Is there any chance this might be approved?

adriancole:
SGTM Add to CHANGES file for the next minor version?

DavidTanner:
@adriancole when are you planning the next release?

DavidTanner:
@adriancole Is there something I need to do to get this released?

Issue #571, opened 2017-07-01
animalsniffer dependency should be optional


brenuart:
`feing-core` declares `org.jvnet:animal-sniffer-annotation` as runtime dependency.
However, animal sniffer is used only during compilation.
It should ideally be declared _optional_ and/or _provided_.

adriancole:
agreed. optional is usually used I think. wanna raise a PR?

brenuart:
PR is merged... 
Can we close the issue ?

PR #572, opened 2017-07-02
make `animal-sniffer-annotation` optional (gh571)

brenuart:
make `animal-sniffer-annotation` optional as it is required only for compilation.
See https://github.com/OpenFeign/feign/issues/571


adriancole:
thanks!

PR #573, opened 2017-07-05
Perform type validation for QueryMap and HeaderMap in Contract

EthanLozano:


EthanLozano:
https://github.com/OpenFeign/feign/pull/567 concluded that more checks should be done upfront in the Contract. This PR aims to perform both the Map validation AND the Map's String key validation in the contract. This reduces the number of checks that happen at "runtime" within the ReflectiveFeign class

sunruh:
This change breaks clients which use a method parameter as HeaderMap which is non-generic (like Spring's HttpHeaders).

The following used to work before the PR (when using spring-cloud-openfeign):
```java
@RequestMapping(method = RequestMethod.POST, path = "/path")
void method(@RequestHeader HttpHeaders headers);
```

It now fails with `java.lang.ClassCastException: java.lang.Class cannot be cast to java.lang.reflect.ParameterizedType` as `HttpHeaders` is non-generic but implements `MultiValueMap<String, String>` which in turn implements `Map<String, List<String>>`.

As a workaround `HttpHeaders` can be changed to `MultiValueMap<String, List<String>>`.

kdavisk6:
@sunruh this bug was corrected in Feign 9.7 via #689

sunruh:
I must have missed that PR looking for fixes. Thanks for the quick feedback.

PR #583, opened 2017-08-17
Add a method metadata property, in order to control whether need to c…

dyc87112:
hello, this pull request is to make the request Body to be an optional properties. 

Current version, if the body is null, feign will return error message “Body parameter %s was null”. 

If this PR can be accept. Then I can fix another issue in Spring Cloud Feign：https://github.com/spring-cloud/spring-cloud-netflix/issues/1047

Thanks.

### DIFF
--- a/core/src/main/java/feign/MethodMetadata.java
+++ b/core/src/main/java/feign/MethodMetadata.java
@@ -44,6 +44,7 @@ public final class MethodMetadata implements Serializable {
       new LinkedHashMap<Integer, Class<? extends Expander>>();
   private Map<Integer, Boolean> indexToEncoded = new LinkedHashMap<Integer, Boolean>();
   private transient Map<Integer, Expander> indexToExpander;
+  private boolean required = true;
 
   MethodMetadata() {
   }
@@ -166,5 +167,13 @@ public final class MethodMetadata implements Serializable {
    */
   public Map<Integer, Expander> indexToExpander() {
     return indexToExpander;
+  }
+
+  public boolean isRequired() {
+    return required;
+  }
+
+  public void setRequired(boolean required) {
+    this.required = required;
   }
 }
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -340,7 +340,9 @@ public class ReflectiveFeign extends Feign {
     protected RequestTemplate resolve(Object[] argv, RequestTemplate mutable,
                                       Map<String, Object> variables) {
       Object body = argv[metadata.bodyIndex()];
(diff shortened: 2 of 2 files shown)


core/src/main/java/feign/MethodMetadata.java
```diff
@@ -44,6 +44,7 @@
       new LinkedHashMap<Integer, Class<? extends Expander>>();
   private Map<Integer, Boolean> indexToEncoded = new LinkedHashMap<Integer, Boolean>();
   private transient Map<Integer, Expander> indexToExpander;
+  private boolean required = true;
```
spencergibb: Should be more descriptive, like bodyRequired.

dyc87112: Good idea 👍

velo:
Hey, why don't you overload the method, creating a new one without the body.

That would accomplish it without passing nulls forward.

I see a point on supporting `Optional<Body>`.

Lemme know what you think.

kdavisk6:
@dyc87112 There are no test cases included in this PR.  Could you please add some to ensure that we have coverage?

kdavisk6:
ping?

How would you intend for the `bodyRequired` flag to be set?

cezar-tech:
@velo @kdavisk6 
If you are interested perhaps I can help but I understand few of how Feign works internally.
If you could explain me how to do it I can give a try

velo:
Well, I'm not sure what was the goal on this pr.

You would need the author back on the discussion.

OlgaMaciaszek:
Hi, @dyc87112 @velo @kdavisk6 I would like to reopen the discussion of this PR.  This kind of functionality would be useful for us to fix [this issue](https://github.com/spring-cloud/spring-cloud-openfeign/issues/126), also discussed [here](https://github.com/spring-cloud/spring-cloud-openfeign/pull/263#discussion_r408804710). I see the PR was closed due to lack of tests and feedback. I can take it over and add tests if you are fine with adding this feature.

velo:
Lemme see if I got it correctly.

Spring has an annotation `@RequestBody` that can be used to "ignore" a parameter.

So, you looking for a mechanism to tell feign to not process a parameter?!

Saw many comments on the ticket and didn't read them all.

kdavisk6:
@OlgaMaciaszek 

Can please create a new issue referencing your issue?  I'd like to move the discussion there so it does not get lost.

SribniyOleg:
@kdavisk6 
Hi - i can refactor this pr or just create a new one with this case. 
In my case I'm facing with the same issue with the author - just need to enable Feign to support RequestBody(required = false ) cuz I need to the optional body for my logic

velo:
We need to differentiate NULL body, from absent

Having another boolean flag is not healthy they will explode out of control someday

Issue #514, opened 2017-01-04
Close stream automatically (SynchronousMethodHandler)


SirWayne:
Hello everybody,

i would like to return an `ResponseEntity<InputStream>` at my interface. Therefore i have an own feign decoder fo an octet strea,. 
```
       if (contentType.equals(OCTET_STREAM)) {
            return new ResponseEntity<InputStream>(response.body().asInputStream(),
                    HttpStatus.valueOf(response.status()));
        }
```

This works really good, but i cannot read from this input stream, because the stream is closed by SynchronousMethodHandler.
The SynchronousMethodHandler close the response body at `executeAndDecode`
```
finally {
      if (shouldClose) {
        ensureClosed(response.body());
      }
    }
```
Why is the stream closed automatically?

spencergibb:
IIRC, Feign isn't designed for streaming. Is that right @adriancole?

adriancole:
>
> IIRC, Feign isn't designed for streaming. Is that right @adriancole
> <https://github.com/adriancole>?
>
> actually, the flag mentioned (shouldClose) is conditional on the body
being a specific size. If within a threshold, it is rebuffered. Otherwise,
it is left as a stream. There's probably a test case showing this somewhere.

phymbert:
Hello, I have submitted StreamDecoder sub interface to let decoder to close the input stream itself in [#586](https://github.com/OpenFeign/feign/pull/586)
Please vote if it solves your cases; also interface name could be discussed.

Issue #220, opened 2015-04-13
Streaming support


spencergibb:
Does feign support streaming as opposed to `byte[]` in memory?  [Saw a former co-worker ask on SO](http://stackoverflow.com/questions/27391234/passing-along-a-data-stream-using-netflix-feign).

adriancole:
pasting what I said on SO:

Feign was designed for control plane apis, which often don't benefit by streaming upwards. Streaming downwards is supported, though.

I've no concern with being more efficient with regards to how buffering works (ex. alternative to byte array). Just bear in mind that most of feign's design revolves around templating forms (json or xml) and reusing these as much as possible (ex. on retransmit, buffered + fixed length is easy and predictable).

I think I'd be most happy about a "streaming" design if it were coupled to the http client. IOTW, a subtype that addresses streaming in a way that makes sense in the transport. For example, InputStream for regular java, OkIo buffer for OkHttp, Netty Buffer for Netty, etc.

adriancole:
ps thanks.. I wasn't watching the `netflix-feign` tag on SO!

adriancole:
FWIW, what I'm thinking is that the body param would address streaming somehow.
Ex. when okhttp is installed, simply passing an okio buffer as a parameter would magically work.

```java
    @RequestLine("POST /")
    void body(okio.BufferedSource source);

    @RequestLine("POST /")
    void body(okio.Buffer source);
```

The tricky part is implicitly getting a correct length or determining chunking. This is where things get complex, and an area I've hoped to avoid as dealing with control apis is complicated enough :)

adriancole:
from an extension POV, a feign Client could register the types it handles natively. These would override the encoder types and skip processing somehow.

ex.
```java
List<Type> transportTypes();
```

`Client.Default` might return `byte[]`, `InputStream`.
`OkHttpClient` might return `com.squareup.okhttp.Request`, `okio.Buffer`, `okio.ByteString` etc.

just thoughts..

spencergibb:
@adriancole thanks for the response.  I've ended up pushing feign beyond control plane apis :-)  I like the idea of customizing the Client.  Maybe the `Type` mapped to some kind of `Handler` to actually the processing.

adriancole:
[...]

PR #586, opened 2017-08-21
Support of Java8 Stream and Jackson Parser

phymbert:


phymbert:
This first PR commit allows to stream **only** the response body in java8 way.
I have also tried to stream request body based on InputStream as suggested by @spencergibb on [#220](https://github.com/OpenFeign/feign/issues/220 ).  But it requires additional changes on the core and maybe not appropriate implementation.
You can have a look on this [fork](https://github.com/phymbert/feign/commit/c75932f2b4aecf8b4e78f9ee7eadc1c051bad090#diff-c3e8e537451608cbdf2705e04803e787R52).
@adriancole Please tell if it is eligible to another PR or should be included in this one or cancelled.

Fixes #514.

phymbert:
Apologize once again... Messed up with rebase/merging.

PR #602, opened 2017-11-15
Add an option to not follow redirects (302) and add a unit test for that

Gui13:
This change allows the Options() constructor to specify whether the HttpConnection should follow redirects or not. The default is yes (as has been from the beginning) but you can now specify that the connection should return the 3xx response if received.

This might fix the #249 issue. At least it solves it in my case: I wanted to receive the 302 response of the API I'm interrogating and act upon that.

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -88,7 +88,7 @@ public interface Client {
       connection.setConnectTimeout(options.connectTimeoutMillis());
       connection.setReadTimeout(options.readTimeoutMillis());
       connection.setAllowUserInteraction(false);
-      connection.setInstanceFollowRedirects(true);
+      connection.setInstanceFollowRedirects(options.isFollowRedirects());
       connection.setRequestMethod(request.method());
 
       Collection<String> contentEncodingValues = request.headers().get(CONTENT_ENCODING);
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -13,6 +13,7 @@
  */
 package feign;
 
+import java.net.HttpURLConnection;
 import java.nio.charset.Charset;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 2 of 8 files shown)


Gui13:
@kdavisk6 I have rebased on master and reworked the imports to keep it clean (my IntelliJ optimizes imports on the fly).

It should now be acceptable?

core/src/main/java/feign/Client.java
```diff
@@ -88,7 +88,7 @@ HttpURLConnection convertAndSend(Request request, Options options) throws IOExce
       connection.setConnectTimeout(options.connectTimeoutMillis());
       connection.setReadTimeout(options.readTimeoutMillis());
       connection.setAllowUserInteraction(false);
-      connection.setInstanceFollowRedirects(true);
+      connection.setInstanceFollowRedirects(options.isFollowRedirects());
```
velo: Won't this impact other clients too?

spencergibb: The other clients would need to be updated to use the option.

Gui13: I made the option default to true, so the previous clients will keep the previous behavior.

You have to explicitely set the option to false to have the new behavior.

velo: This change need to be included to the okhttp and ribbon
https://github.com/OpenFeign/feign/blob/master/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
https://github.com/OpenFeign/feign/blob/master/ribbon/src/main/java/feign/ribbon/RibbonClient.java

Gui13: @velo I have added implementations for Ribbon, OkHttp and LbClient as well (since unit tests of Ribbon use it too).

velo:
One of the tests failed. Could you please take a look?

Gui13:
I did and corrected. I just forgot to run the entire testing suite after commiting the new options.

Issue #604, opened 2017-11-16
Enable OptionalDecoder to parse 204 responses into Optional.empty


androidfred:
GIVEN
An API that returns

```
{"userId": "1", "friends": [{"name": "John"},{"name":"Jane"}]}
```

on GET /users/1/friends, and returns 204 no body on GET /users/2/friends (I would prefer people didn't design their APIs like this, but it's not uncommon)

STEPS TO REPRODUCE
```
class FriendsResponse { String userId; List<Friends> friends; }
```

```
interface SomeAPI { //wired up with OptionalDecoder
  @RequestLine("GET /users/{userId}/friends")
  Optional<FriendsResponse> getFriendsFor(@Param("userId") String userId);
}
```

```
someAPI.getFriendsFor("2");
```

EXPECTED RESULT
Returns Optional.empty

ACTUAL RESULT
Doesn't return Optional.empty

kdavisk6:
I wonder if we can combine this with #618 as a way to add more fine tuned control to decoding responses.  

In any case, `204` does mean `NO CONTENT`, so I think your approach is sound.  We'll look at the PR as part of a future release.

PR #605, opened 2017-11-16
Enable OptionalDecoder to parse 204 responses into Optional.empty

androidfred:
Addresses https://github.com/OpenFeign/feign/issues/604

### DIFF
--- a/java8/src/main/java/feign/optionals/OptionalDecoder.java
+++ b/java8/src/main/java/feign/optionals/OptionalDecoder.java
@@ -22,8 +22,7 @@ public final class OptionalDecoder implements Decoder {
         if(!isOptional(type)) {
             return delegate.decode(response, type);
         }
-
-        if(response.status() == 404) {
+        if(response.status() == 404 || response.status() == 204) {
             return Optional.empty();
         }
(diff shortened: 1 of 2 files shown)


java8/src/main/java/feign/optionals/OptionalDecoder.java
```diff
@@ -22,8 +22,7 @@ public OptionalDecoder(Decoder delegate) {
         if(!isOptional(type)) {
             return delegate.decode(response, type);
         }
-
-        if(response.status() == 404) {
+        if(response.status() == 404 || response.status() == 204) {
```
androidfred: Potentially even more statuses could be added, in which case it should be rewritten to something like 

```
if (Stream.of(404, 204).anyMatch(status -> status == response.status()) { ... }
```

kdavisk6: At the moment, we are keeping `feign` core at Java 6+.  Java 8 Specific features are available in the `feign-java8` module.

velo:
# 204 No Content
The server successfully processed the request and is not returning any content.[14]

With that in mind, this behavior makes sense for me

Issue #503, opened 2016-12-07
Bug on POST request with body


DmitryRyzhikov:
Hi Adrian, 

Encountered the situation when Feign is used to send POST request with body to server. The only way to do it described in  [(documentation)](https://github.com/OpenFeign/feign#body-templates) - is to accept necessary parameters as method arguments annotated with @Param and place it into template annotated with @Body.

Used feign version:
`<feign.version>8.18.0</feign.version>`

Looks I found a bug in implementation. This is how my interface method looks on client side:

```
    @RequestLine("POST api/v1/{channel}/finalize")
    @Headers("Content-Type: application/json")
    @Body("%7B\n" +
            "  \"nickname\": \"{nickname}\",\n" +
            "  \"seanceId\": {seanceId},\n" +
            "%7D")
    Result finalize(@Param("channel") String channel,
                                     @Param("nickname") String nickname,
                                     @Param("seanceId") Integer seanceId);
```

If I'm using it with all parameters passed (without NULLs), everything is OK.
`Result result = service.finalize("test", "player", 12345);`

But if I'm calling this method with null instead of any arguments, like this
`Result result = service.finalize("test", "player", null);`

then on service side I'm catching POST request with this body:
```
{
  "nickname": "player",
  "seanceId": {seanceId}
}
```

As you can see seanceId placeholder is not replaced with null. Instead it came with unresolved placeholder and I got an exception on attempt of de-serialization accepted JSON to object that is expected in controller as BODY.

I found similar bug solved in this PR, but it doesn't help in situation described above:
https://github.com/OpenFeign/feign/pull/47/files

vladislav-sidorovich:
As fix of this issue we can just remove `if` statement at line `feign/ReflectiveFeign.java:203`. 

Are exist some reasons for this statement?

britztopher:
any update on this issue? As I still am experiencing this in `spring-cloud-starter-feign 1.2.4` release

sakhno:
have same problem with "text/plain" and null string as body, any movements with this issue? feign version 8.18.0

adriancole:
Did someone raise a PR with a fix and test yet?

sakhno:
This [pull request](https://github.com/OpenFeign/feign/pull/607) should fix the problem.

kdavisk6:
[...]

PR #607, opened 2017-11-25
Null values handling in templates

sakhno:
Fixes [#503](https://github.com/OpenFeign/feign/issues/503). Replace null values in template with "null" if it is json object(begins from "%7B" and ends "%7D") and ignores them in all other conditions. Although, it removes double quotes around null values in case it is json object.


kdavisk6:
Addressed via #778.  Unresolved variables in Body Templates are left alone.

PR #614, opened 2017-12-11
Using proper formatting pattern for logIOException

### LINKED ISSUE #613: UnknownFormatConversionException when FULL logging some IOExceptions
Some IOExceptions actually contain a '%' sign which trips the String.format function used by Logger.
logIOException logs the exception stacktrace as a format string, rather than an argument to the "%s" format.

For example:
```
<--- ERROR HttpHostConnectException: Connect to localhost:8080 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1, localhost/fe80:0:0:0:0:0:0:1%1] failed: Connection refused (Connection refused) (2174ms)

java.util.UnknownFormatConversionException: Conversion = '1'

	at java.util.Formatter.checkText(Formatter.java:2579)
	at java.util.Formatter.parse(Formatter.java:2565)
	at java.util.Formatter.format(Formatter.java:2501)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2940)
	at feign.slf4j.Slf4jLogger.log(Slf4jLogger.java:72)
	at feign.Logger.logIOException(Logger.java:119)
```

jvanderhoek:
Logger.logIOException is supplying the stacktrace in the format pattern of the logger, rather than as an argument. This causes unexpected exceptions when the exception contains a '%' character (like in UnknownHostException on Mac OSX).

Fixes #613

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -116,7 +116,7 @@ public abstract class Logger {
     if (logLevel.ordinal() >= Level.FULL.ordinal()) {
       StringWriter sw = new StringWriter();
       ioe.printStackTrace(new PrintWriter(sw));
-      log(configKey, sw.toString());
+      log(configKey, "%s", sw.toString());
       log(configKey, "<--- END ERROR");
     }
     return ioe;
--- a/core/src/test/java/feign/LoggerTest.java
+++ b/core/src/test/java/feign/LoggerTest.java
@@ -23,6 +23,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.runners.Enclosed;
 import org.junit.rules.ExpectedException;
+import org.junit.rules.RuleChain;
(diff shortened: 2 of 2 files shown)


kdavisk6:
This looks good to me.  I don't see any tests included however.  Can you add one?

jvanderhoek:
I've added the FormatCharacterTest to LoggerTest to test the original fix.

On adding this test I discovered that the RecordingLogger Rule was not actually being applied, because the FeignException gets thrown first, skips the recording logger and then the exception gets handled by the expected exception rule. So I've added a RuleChain to enforce the expected exception handling first and then the recording logging rule. This (of course) revealed that some of the tests were actually broken, so I've fixed those.  I hope it's acceptable.

kdavisk6:
@jvanderhoek This looks very good.  Great job on catching that error.  I think this is ready to be merged.

Issue #520, opened 2017-01-10
How to pass request parameters as object ?


anton-arsentyev:
Hello, is there a way to pass object as a parameter to interface and get all fields as query params ?

Example:

Interface:

```
@RequestLine("POST /api/doSomething")
void doSomething(@QueryObject Parameters parameters);
```

Model:

```
class Parameters {
  String name;
  String secondName;
}
```

And result request would be:

`{host}/api/doSomething?name=name123&secondName=sn222`

imtiaz-m4:
I have found a solution here 
[Using request objects with Feign](http://source.coveo.com/2016/07/21/feign-request-object) by Jonathan Rochette and it looks pretty neat. Those extra annotation to correctly identify which parameter should go where.
But it will need extra coding in your application. Looks like Feign does not yet provide support for this.

I also need this solution built in Feign.

DanielYWoo:
We want it! Let's say you have two mandatory parameters and an optional parameters, you will have to implement two methods: foo(m1, m2) and foo(m1, m2, o1). If you have more optional parameters you will have to implement a lot of overloaded methods, otherwise you can only implement a method with a lot of parameters and when it is called, you have to pass in a bunch of null values like foo(m1, m2, null, null, null, null, o5, null, null, o8) which is very ugly.

If we have object parameter we can do this: first, use a constructor to pass in all mandatory parameters, then use setters to set the optional parameters like this:

```
q = new FindUsersRequest(category, type);
q.setRowStart(10);
q.setGender(M);
feign.findUsers(q);

```

smougenot:
Proposed solution is basic but fis the need in this PR (I also required it, so I share what I tried)
Hope it'll help

kdavisk6:
This has been addressed by #667.  If there are additional use cases or use cases where this may not apply, please open a new issue.

Issue #601, opened 2017-11-13
Provide RequestTemplate.getParams() to access template variables (@Param) within an interceptor


d0x:
It would be nice having a method to access @Param annotated parameters within an Interceptor.
Currently @Param template variables are only applied to header, requestline or the body.

As example this kind of **RequestTemplate.getParams()** would really help us.

```
public class MyServce {
   private final MyClient client;

   public void doSmth() {
      // Set the codec to some value. e.g. MP3. The key thing is that only the class "MyService"
      // knows which codec to choose.
      client.getSomething("MP3");
   }
}

public interface MyClient {
    // define the template variable "codec".
    // codec is not used in RequestLine, Body or Header
    @RequestLine("GET /")
    Object getSomething(@Param("codec") String codec);
}

public class CodecInterceptor implements RequestInterceptor {
    @Override
    public void apply(RequestTemplate requestTemplate) {
        // use params.get("codec") to get the codec and generate the headers
        Map<String, Object> params = requestTemplate.getParams();
        // ... 
    }
}
```

Since the business logic to detect which "codec" is required, is and should be placed in `MyService`, there is currently no way to pass this information (in a thread safe manner) to the interceptor. If there would be a `RequestTemplate.getParams()`, the business logic to detect the codec could stay in `MyService` and the logic to translate the codec into headers could be done within the interceptor.

That would help to accomplish separation of concerns (Service=BusinessLogic, Interceptor=HttpMagic).

d0x:
Another example would be that MyService detected the current `UserId` and then passes it to the interceptor which will use it to generate a valid token.

rage-shadowman:
This could be solved using `@CustomParam` annotation if #667 gets pulled in.

kdavisk6:
@d0x 

In this situation, you may be able to use a custom Parameter Expander to resolve the value.  For example:

```java
public interface MyClient {
   
    @RequestLine("GET /")
    @Headers("Codec-Header: {codec}")
    Object getSomething(@Param("codec", expander=CodecExpander.class) String codec);
}
```
The `CodecExpander` can process the `codec` value and return the expanded version of the parameter.  This would allow you to completely change what `codec` resolved do on-demand.  Now, this only works if you need to resolve a single template expression, but it may be enough to help you, without requiring a new feature.

nickmafra:
I need 2 clients:
- a client that uses a dynamic basic auth (change username and password at runtime), but I haven't found a feature for that.
[...]

PR #620, opened 2017-12-29
Support Object Query with GET request

DanielYWoo:
Let's say you have two mandatory parameters and an optional parameters, you will have to implement it this way:
```
foo(m1, m2, o1)
```

If you have more optional parameters you will have to implement a lot of overloaded methods, otherwise you can only implement a method with a lot of parameters and when it is called, you have to pass in a bunch of null values like this:
```
foo(m1, m2, null, null, null, null, o5, null, null, o8) 
```

This is very ugly, if we have object parameter we can do this: first, use a constructor to pass in all mandatory parameters, then use setters to set the optional parameters like this:

```
q = new FindUsersRequest(category, type);
q.setRowStart(10);
q.setGender(M);
feign.findUsers(q);

```
Relaated to #520

### DIFF
--- a/core/src/main/java/feign/QueryObject.java
+++ b/core/src/main/java/feign/QueryObject.java
@@ -1,0 +1,23 @@
+package feign;
+
+import java.lang.annotation.Retention;
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * This is a tag annotation to support Object Query. By default, Feign does not support object query with a GET request,
+ * You can use use QueryObjectEncoder to support this.
+ */
+@Retention(RUNTIME)
+@java.lang.annotation.Target(TYPE)
+public @interface QueryObject {
+
+    @Retention(RUNTIME)
+    @java.lang.annotation.Target(TYPE)
+    @interface Param {
+        String value();
+    }
+
+}
+
--- a/core/src/main/java/feign/QueryObject.java
+++ b/core/src/main/java/feign/QueryObject.java
@@ -2,6 +2,7 @@ package feign;
 
 import java.lang.annotation.Retention;
 
+import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
@@ -14,7 +15,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 public @interface QueryObject {
 
     @Retention(RUNTIME)
-    @java.lang.annotation.Target(TYPE)
+    @java.lang.annotation.Target(METHOD)
     @interface Param {
         String value();
     }
--- a/core/src/main/java/feign/codec/QueryObjectEncoder.java
+++ b/core/src/main/java/feign/codec/QueryObjectEncoder.java
@@ -1,7 +1,6 @@
 package feign.codec;
 
 
-import feign.Param;
 import feign.QueryObject;
 import feign.RequestTemplate;
 
@@ -9,9 +8,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 public class QueryObjectEncoder implements Encoder {
 
@@ -24,7 +21,7 @@ public class QueryObjectEncoder implements Encoder {
     @Override
     public void encode(Object parametersObject, Type bodyType, RequestTemplate template) throws EncodeException {
         if (parametersObject.getClass().getAnnotation(QueryObject.class) != null) {
-            Map<String, Object> params = new HashMap<>();
+            Map<String, Object> params = new HashMap<String, Object>();
             try {
                 for (Method method : parametersObject.getClass().getMethods()) {
                     QueryObject.Param param = method.getAnnotation(QueryObject.Param.class);
@@ -32,18 +29,16 @@ public class QueryObjectEncoder implements Encoder {
                         String key = param.value();
                         Object value = method.invoke(parametersObject);
                         if (value != null) {
-                            if (List.class.isAssignableFrom(method.getReturnType())) {
-                                value = ((List<?>) value).stream().map(String::valueOf).collect(Collectors.toList());
-                            } else {
-                                value = String.valueOf(value);
-                            }
-                            params.put(key, value);
-                            template.query(key, keyToTemplate(key));
+                            value = String.valueOf(value);
                         }
+                        params.put(key, value);
+                        template.query(key, keyToTemplate(key));
                     }
                 }
                 template.resolve(params);
-            } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+            } catch (IllegalAccessException e) {
+                throw new EncodeException("Could not encode object query correctly", e);
+            } catch (InvocationTargetException e) {
                 throw new EncodeException("Could not encode object query correctly", e);
             }
         } else {
--- a/core/src/main/java/feign/codec/QueryObjectEncoder.java
+++ b/core/src/main/java/feign/codec/QueryObjectEncoder.java
@@ -10,6 +10,26 @@ import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * You can create an QueryObject like this
+ * <pre>
+ * {@code
+ * @QueryObject
+ * class FindUsersQuery {
+ *    .....
+ *    @QueryObject.Param("userId")
+ *    long getUserId() {...}
+ *    @QueryObject.Param("userName")
+ *    long getUserName() {...}
+ * }
+ * </pre>
+ * Then you can declare the client like this:
+ * <pre>
+ * {@code
+ * @RequestLine("GET /users?userId={userId}&userName={userName}")
+ * List<User> getUsers(FindUsersQuery query);
+ * </pre>
+ */
 public class QueryObjectEncoder implements Encoder {
 
     private Encoder fallbackEncoder;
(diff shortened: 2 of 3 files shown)


Gui13:
I'm not affiliated to Feign but it would be nice to have some unit tests to test this feature?

DanielYWoo:
yep, I will add some unit tests to the pull request.

Thanks,
Daniel
________________________________

DanielYWoo:
Too busy recently, will add unit tests later, sorry

DanielYWoo:
unit test added, please review it, thanks!

yangtao309:
good job ！ +1 @DanielYWoo

benmanbs:
On the whole I think this looks like a pretty cool bit of functionality. I have one main concern - I don't think this is documented well at all. I would want to see some very clear documentation that indicates some of the assumptions that this code makes:
- the `@QueryObject.Param` annotation goes on the getter, not the field
- the valid entries for query params are only simple objects, things like lists or arrays are not supported
- this does not support boolean getters with the format `isFoo` instead of `getFoo` (your code directly checks for the presence of `get` at the beginning of the method)

I am unfamiliar with the format `foo={bar}`, just out of curiosity does this work on the other side? Like if you implement the feign interface with your actual controller, will the object also unpack from the request? (A common pattern at my company is to package a feign interface in the same repo as the controller itself, and have the controller implement the interface). 

Would love to see some unit tests with edge cases for the ObjectEncoder in addition to the integration style tests you've got on the Feign class. 

Finally, I would recommend that you read https://github.com/OpenFeign/feign/blob/master/CONTRIBUTING.md and format your PR accordingly. For example, I don't see any licensing information on new files.

kdavisk6:
With regards to the `isFoo` vs `getFoo`, you may want to use the Java Beans api to determine the properties.  It would allow for more robust handling and interoperability.

rage-shadowman:
Why not also allow the annotations on member variables? (kinda like the jackson library)

rage-shadowman:
This is an excellent change. I would personally like to see it included, and even expanded upon (maybe allowing path param annotations as well, and definitely allowing annotations on member variables rather than just getter methods).

kdavisk6:
If you want that, why not use `Spring`.  

https://cloud.spring.io/spring-cloud-netflix/multi/multi_spring-cloud-feign.html

This adds support for the Spring MVC annotations include Path, Request, and other variables.

DanielYWoo:
@benmanbs I will change accordingly later, I am too busy recently, sorry about the late response.

DanielYWoo:
@kdavisk6 Yes, you spotted a bug about the java bean convention. I will fix it and add more unit tests. And I probably will make the annotation only on member variables instead of getters, to be friendly to lombok. And I will do some benchmark to see how much performance can be improved with javassist by avoiding reflections on each call. If the performance improvements is worthy of it, I probably make javassist as an *optional* dependency.

kdavisk6:
@DanielYWoo with regards to Spring, I was making a reference to something that has already been implemented showing what you all wanted.  If your desire is to add this to Feign proper, I think a more elegant solution is to add an additional `Contract` or modify the existing default `Contract` implementation.  Have you looked at that Interface?

rage-shadowman:
@kdavisk6 oohhh... That's nice. I hadn't noticed that before. The jax-rs contract seems like a good place to start: https://github.com/OpenFeign/feign/blob/master/jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java

Thanks for the tip.

As regards Spring, I agree with @DanielYWoo that it isn't for libraries. Spring is way too much overkill for a simple client library that might be used in any context. We do have some spring services here, so it must be usable in such a context, but not everything we have needs or wants all of that complexity, so it shouldn't pull in spring as a dependency just to use (what should be) a simple client library.

rage-shadowman:
@kdavisk6 looking at the `Contract` interface, itonly tags the individual passed-in parameters, but does not have access to the actual objects themselves. It seems to expect all arguments tagged to be passed via the method parameters individually, to be handle the actual encoding/expanding of them elsewhere.

This doesn't look like it would work. Am I missing something? Do you have any examples that you could point us to?

kdavisk6:
Again, I'll go back to Spring

https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java

In their code, they created `AnnotatedParameterProcessor` that is used within the `Contract`, along with a `ConversionService`.  The parameter processor is used to interpret the `@RequestParam`, `@PathParam`, and `@RequestMapping` annotations on the `Feign` interface and translate the arguments for use later by the `Client`.  

The `ConversionService` is used convert items from Objects to the appropriate types.  For example, Spring registers a `MappingJacksonHttpMessageConverter`, with the `DefaultConversionService` to handle objects that should be mapped to JSON output.  The `SpringEncoder` recognizes this based on the information provided by the `SpringMvcContract` and will encode the objects in the Request accordingly.

Using Spring as an example, one way to make this work in core Feign would be expand the `Contract` to expand on what @DanielYWoo has already done in this PR.  @benmanbs has already made some very good suggestions in that regards.  

I'm learning too as I review this PR.  The more I look into it, the more @DanielYWoo's approach is pretty close.  I still think moving away from reflection is a better idea, and @DanielYWoo has already agreed.  This is a great start.

rage-shadowman:
@kdavisk6 that Spring class doesn't do what this pull request is trying to do, it can actually generate everything it needs with the method signature, so it can use a custom `Contract`.

I think the only way to do what this pull request is trying to do, is via reflection. And I believe @benmanbs was talking about using reflection and caching the results, instead of using reflection with each call (as opposed to moving away from reflection, or not using it at all).

kdavisk6:
Understood.  In that case, I still recommend that `javaassist` or the bean api be used instead of pure reflection.  Otherwise I think we should be good to go.

rage-shadowman:
I think that the cleaner solution would be to modify the default `Contract` to accept a new annotation and the default `InvocationHandlerFactory` to handle the new annotation.

rage-shadowman:
I attempted the "cleaner solution" myself in #667. Mine is the 3rd attempt to solve this particular problem (#636 and this one are the first 2) that I've seen here. I hope one of these gets traction and actually gets into the feign code base.

rage-shadowman:
@DanielYWoo Would #667 work for you?

It should allow you to create your QueryObject annotation and QueryObjectEncoder class locally and include it via:
```java
@CustomParam(encoder = QueryObjectEncoder.class) MyQueryObject myQueryObject
```

That PR allows your custom encoder to modify query params (as your class is doing) or headers (we needed to use param context in generating our auth headers as well) or whatever you need (as with custom encoders and interceptors, use at your own risk since you can totally mangle everything in the template if you screw something up).

I'd like to find something that would work for you, me, #636 and anyone else who needs something similar (this might even solve #601 if it doesn't need to be run again on every retry).

DanielYWoo:
@rage-shadowman I am fine with #667, it's more flexible, but I hope you can include QueryParams into your change, because this annotation will solve 95% of our real world problems so it should just work out of box.
And we probably need javassist for better performance.

rage-shadowman:
@DanielYWoo what do you mean "QueryParams"? I updated the PR based on comments to just use the existing `QueryMap` annotation (`@QueryMap MyPojo myPojo`), and it now just looks up all fields on your POJO (rather than getter methods) and uses their names and values to internally generate a `Map` that should work the same as if you just passed in a `Map` as in: `@QueryMap Map<String, Object> myOptionalQueryParams`

kdavisk6:
With support added for `QueryMap` in 9.7, should we consider that approach sufficient or is this issue still sufficiently different that we should continue this discussion?

DanielYWoo:
@kdavisk6 Does QueryMap in 9.7 support POJO other than Map<Object, Object>?
What we want is POJO because
1). all mandatory parameters in its constructor, with only getters
2). all optional parameters with setters/getters

rage-shadowman:
Yes, you can provide your own `QueryMapEncoder` or you can use the default one. It's documented in the [README.md](../blob/master/README.md) file (near the end, under the section "**Dynamic Query Parameters**").

DanielYWoo:
@rage-shadowman It works but you will have to write a custom encoder for each query object class. This is not what I want. What I want is a single generic encoder to use the annotation info in your query object class.

kdavisk6:
@DanielYWoo, I think you may have misunderstood the documentation.  `@QueryMap` can now take a POJO in addition to a `Map<>`.  The default encoder will transform all properties of the POJO into `Map<>` internally and expand them onto the URL.  For example:

```java
public class FormBean {
   private String name;
   private String favoriteIceCream;
}
```
``` java
public interface IceCreamService {
   @RequestLine("POST /favorite")
   public void favoriteIceCream(@QueryMap FormBean favoriteForm)
}
```
will expand to
` /favorite?name={name}&favoriteIceCream={favoriteIceCream}`

If you need to customize this processing, you can provide an optional `QueryMapEncoder` on the `@QueryMap` annotation.  At this time, we decided to keep the parsing simple.  If we find that more users want more control over the mapping, we can consider expanding the feature.  I suggest giving 9.7 a try.  If afterwards you still feel that it's not sufficient, please open an issue so we can gauge interest.

rage-shadowman:
@DanielYWoo You can use some of the code from the `QueryObjectEncoder` that you have here, but instead of manipulating the template directly, your encoder would just build and return a query map (`Map<String, Object>`) based on your object.

The default one will build the map for you, using the variable names as keys, and the variable values as values. The default encoder is extremely simple, no getter parsing, just looks at the local POJO member variables to build the map (they can even be private and final if you like).

kdavisk6:
@DanielYWoo what are your thoughts on this?  Are there any additional use cases you have where a `@QueryMap` with a `QueryMapEncoder` will not suffice?

DanielYWoo:
@kdavisk6 I think @rage-shadowman 's solution is more flexible but a little overkill, and you will have to write a whole lot of decoders, the QueryMap annotation is just fine for me. I think we can close this ticket now. Thank you!

Issue #630, opened 2018-01-25
Allow empty @Path annotation in JAXRSContract?


iamdanfox:
Currently, I'm unable to connect to a server using the following interface:

```java
@Path("") // empty string causes 🔥🔥🔥 
public interface NoSlashService {
    @GET
    @Path("/foo")
    @Produces(MediaType.APPLICATION_JSON)
    Map<String, Integer> foo();
}
```

Given that jetty/jersey seem to be able to serve this happily, would it be possible to relax the validation error from `JAXRSContract`, possibly turning it into a warning?

```
java.lang.IllegalStateException: Path.value() was empty on type com.something.demo.NoSlashService
	at feign.Util.checkState(Util.java:128)
	at feign.jaxrs.JAXRSContract.processAnnotationOnClass(JAXRSContract.java:58)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:91)
	at feign.jaxrs.JAXRSContract.parseAndValidateMetadata(JAXRSContract.java:50)
	at feign.Contract$BaseContract.parseAndValidatateMetadata(Contract.java:64)
```

I know there's an easy workaround, just move the slash to the top (below).  Just curious whether feign is unnecessarily banning a technically valid configuration!

```java
@Path("/")
public interface NoSlashService {
    @GET
    @Path("foo")
    @Produces(MediaType.APPLICATION_JSON)
    Map<String, Integer> foo();
}
```

kdavisk6:
Addressed by #631

PR #631, opened 2018-01-25
@Path("") annotations are equivalent to @Path("/")

cakofony:
The documentation suggests that all paths are relative:
https://docs.oracle.com/javaee/6/api/javax/ws/rs/Path.html

### DIFF
--- a/jaxrs/README.md
+++ b/jaxrs/README.md
@@ -11,8 +11,6 @@ server interface behavior.
 ## Currently Supported Annotation Processing
 Feign only supports processing java interfaces (not abstract or concrete classes).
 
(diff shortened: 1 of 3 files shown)


cakofony:
This change resolves #630

kdavisk6:
This looks good to me.  I suggest adding some changes to the documentation as well.  Right now the README calls this situation out explicitly as something that will throw an `IllegalStateException`.  See [JAXRS Readme](https://github.com/OpenFeign/feign/blob/master/jaxrs/README.md)

cakofony:
Great catch @kdavisk6, thanks! Fixed.

kdavisk6:
Approved from my point of view.  @velo any additional feedback?

velo:
Nop, good to go

PR #636, opened 2018-02-07
Add the ability to pass request parameters as object

### LINKED ISSUE #520: How to pass request parameters as object ?

smougenot:
* Adds an encoder using a parameter object to add query parameters to the request from its non null properties
  * encoder introspects the object to find getters
  * only add parameters to the query if the value is non null
  * uses the property name as the query parameter name
* Does not require to add any annotation in the parameter object class (sometime you don't have this possibility; as with external lib)

Fixes #520

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,27 @@ LoginClient client = Feign.builder()
 client.login(new Credentials("denominator", "secret"));
 ```
 
+Sending `Get` request using an object containing request parameters can be achieved using `ObjectToQueryParamEncoder`.
+
+This encoder uses an Object as source for query parameters. Add non null values of properties as query parameters using the property name as the query parameter name.
+
+```Java
+
+interface LoginClient {
+  @RequestLine("GET /")
+  void login(Credentials creds);
+}
+...
+LoginClient client = Feign.builder()
+                          .encoder(new ObjectToQueryParamEncoder())
+                          .logLevel(Logger.Level.FULL)
+                          .target(LoginClient.class, "https://foo.com");
+
+client.login(new Credentials("denominator", "secret"));
+```
+
+This will generate a get request `https://foo.com?user_name=denominator&password=secret`
+
 ### @Body templates
 The `@Body` annotation indicates a template to expand using parameters annotated with `@Param`. You will likely need to add a `Content-Type` header.
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -41,6 +41,12 @@
       <version>4.2.5.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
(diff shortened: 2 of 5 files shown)


smougenot:
❓ where should I amend the documentation about this new `Encoder`, not sure if it's possible for me

spencergibb:
https://github.com/OpenFeign/feign/blob/master/README.md#encoders

smougenot:
Documentation added

rage-shadowman:
I'm new to feign, so I may be missing something obvious, but...

Since this is expanding query parameters objects, rather than request body objects, shouldn't this be an `Expander` instead of an `Encoder`? As an `Encoder`, won't this take your POST request body object and try to throw it on the URL as a set of request parameters?

rage-shadowman:
Here is another attempt at a solution by someone else that uses a fallback encoder:
https://github.com/OpenFeign/feign/pull/620/files#diff-98ef494385d00576dd615054594e5033

It would be nice if encoders were typed and had a `canEncode` method so you could chain them easier or have an ordered that would be run through until one is found that can encode your object (maybe checking things like object annotations or "Content-Type" and "Accept" headers to know).

rage-shadowman:
@smougenot Would #667 work for you?

It should allow you to create your ObjectToQueryParamEncoder class locally and include it via:
```java
@CustomParam(encoder = ObjectToQueryParamEncoder.class) MyCustomObject myCustomObject
```

That PR allows your custom encoder to modify query params (as your class is doing) or headers (we needed to use param context in generating our auth headers as well) or whatever you need (as with custom encoders and interceptors, use at your own risk since you can totally mangle everything in the template if you screw something up).

I'd like to find something that would work for you, me, #620 and anyone else who needs something similar (this might even solve #601 if it doesn't need to be run again on every retry).

rage-shadowman:
@smougenot, #667 has been changed to work with POJOs by using field names as parameter names and field values as parameter values. The new usage is just:

```java
@QueryMap MyCustomObject myCustomObject
```

...and your custom object will have to use fields (`private final` is fine) rather than getter methods.
Will this work for your case?

smougenot:
Looks fine (I didn't gave it a try)
And most certainly worth having in Feign

But you still rely on injecting annotation.
Whereas I only tuned the feign client (code external to the POJO and Interface).

PR #638, opened 2018-02-08
Add support for `CompletableFuture` for method return types

wjam:
Implements support for `CompletableFuture` on method return types by converting through RxJava `Observable`


kdavisk6:
There are few things that will help make this PR ready for review.  Please see [CONTRIBUTING](https://github.com/OpenFeign/feign/blob/master/CONTRIBUTING.md) for more information.

Also, this does appear to push Java version up.  In the past we have tried not to do this, for various reasons.  Take a look at [HACKING](https://github.com/OpenFeign/feign/blob/master/HACKING.md) for more tips.  If this feature gets enough support, it may be better of in a new `feign-hystrix-java8` module.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# top-most EditorConfig file
```
kdavisk6: I'm not sure we should be including editor configuration files this way.

hystrix/src/main/java/feign/hystrix/ObservableCompletableFuture.java
```diff
@@ -0,0 +1,35 @@
+package feign.hystrix;
+
+import com.netflix.hystrix.HystrixCommand;
+import org.jvnet.animal_sniffer.IgnoreJRERequirement;
+import rx.Subscription;
+import rx.functions.Action1;
+
+import java.util.concurrent.CompletableFuture;
+
+@IgnoreJRERequirement
```
kdavisk6: Personally, I don't like this kind of thing.  If this code requires JDK8, then it should be in the `jdk8` module.

velo: Is there any strong reason for our current java 6 constraint?

kdavisk6: It is my understanding, from reading previous comments and documentation from @adriancole, it is the opinion of this project that core `feign` should remain JDK 6+ and that all JDK 8 features are in the `feign-java8` module.  This PR will require core `feign` to move to JDK 8, which has been called out in other PR and issues before as something we should try to avoid.

Personally, I do not have a strong opinion on this.  My questions and thoughts are an attempt to stay consistent.  I am open to having a discussion with the other contributors to see if we should revisit this decision.

spencergibb: I change to jdk 8 baseline should require a major revision (ie 10.x)

velo: > change to jdk 8 baseline should require a major revision (ie 10.x)

LGTM =)

kdavisk6:
@wjam This PR limits support to `HystrixFeign` only.  Would it be possible to update this PR to add general support for `CompleteableFuture` outside of `Hystrix`?

PR #641, opened 2018-02-12
fix default method gets wrapped twice

swirekadam:
There is a bug on wrapping interface default methods on Hystrix Targetter:

`
interface test {
 default Observable<String> observable() {
return Observable.just("one", "two");
 }
}
`

becomes: 

`Observable<Observable<String>>`


kdavisk6:
This seems straightforward enough.  No comments from me.

velo:
LGTM too

PR #649, opened 2018-03-05
add doNotCloseAfterDecode flag to Feign builder

### LINKED ISSUE #514: Close stream automatically (SynchronousMethodHandler)

benmanbs:
This commit adds the `doNotCloseAfterDecode` flag to the Feign builder object. This allows you to
lazily evaluate the response in your Decoder, in order to support Iterators or Java 8 Streams.

This is a pretty light weight change, to support a do-it-yourself approach to lazy instantiation.

Fixes #514.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 9.6
+* Feign builder now supports flag `doNotCloseAfterDecode` to support lazy iteration of responses.
+
 ### Version 9.5.1
 * When specified, Content-Type header is now included on OkHttp requests lacking a body.
 * Sets empty HttpEntity if apache request body is null.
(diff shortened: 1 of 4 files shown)


adriancole:
brownie points for someone who does `./mvnw versions:set -DnewVersion=9.6.0-SNAPSHOT -DgenerateBackupPoms=false` and raises a PR to update the versions since we are going for a new one now.

cc @velo

benmanbs:
I'm happy to do that

PR #651, opened 2018-03-06
Support of Java 8 Stream Decoder and Jackson Iterator Decoder

phymbert:
Introduces 2 new decoders:

**JacksonIteratorDecoder** in feign-jackson provides a closeable iterator for lazy processing json arrays. This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled. Built-in iterator auto closes the Response when it reached json array end or failed to parse stream. If this iterator is not fetched till the end, it has to be casted to Closeable and explicity `Closeable.close()` by the consumer.

**StreamDecoder** in feign-java8 provides a `java.util.stream.Stream` for lazy processing response based on a wrapped decoder that is responsible to transform the Response to an Iterator. This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled.

Added decoders benchmark to compare performance of 3 differents implementations to process json arrays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,6 @@
 ### Version 9.6
 * Feign builder now supports flag `doNotCloseAfterDecode` to support lazy iteration of responses.
-* Introduces `JacksonIteratorDecoder` in `feign-jackson` which provides a closeable iterator for lazy processing json arrays. This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled. Built-in iterator auto closes the `Response` when it reached json array end or failed to parse stream.
-  If this iterator is not fetched till the end, it has to be casted to `Closeable` and explicity `Closeable.close()` by the consumer.
-* Introduces `StreamDecoder` in `feign-java8` which provides a `java.util.stream.Stream` for lazy processing response based on a wrapped decoder that is responsible to transform the `Response` to an `Iterator`.
- This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled.
+* Adds `JacksonIteratorDecoder` and `StreamDecoder` to decode responses as `java.util.Iterator` or `java.util.stream.Stream`.
 
 ### Version 9.5.1
 * When specified, Content-Type header is now included on OkHttp requests lacking a body.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ### Version 9.6
 * Feign builder now supports flag `doNotCloseAfterDecode` to support lazy iteration of responses.
+* Introduces `JacksonIteratorDecoder` in `feign-jackson` which provides a closeable iterator for lazy processing json arrays. This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled. Built-in iterator auto closes the `Response` when it reached json array end or failed to parse stream.
+  If this iterator is not fetched till the end, it has to be casted to `Closeable` and explicity `Closeable.close()` by the consumer.
+* Introduces `StreamDecoder` in `feign-java8` which provides a `java.util.stream.Stream` for lazy processing response based on a wrapped decoder that is responsible to transform the `Response` to an `Iterator`.
+ This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled.
 
 ### Version 9.5.1
 * When specified, Content-Type header is now included on OkHttp requests lacking a body.
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -24,26 +24,38 @@
     <version>9</version>
   </parent>
 
-  <!-- TODO: change group id when 9.0 is released -->
-  <groupId>com.netflix.feign</groupId>
   <artifactId>feign-benchmark</artifactId>
-  <packaging>jar</packaging>
-  <version>8.1.0-SNAPSHOT</version>
+  <version>9.6.0-SNAPSHOT</version>
   <name>Feign Benchmark (JMH)</name>
 
   <properties>
-    <jmh.version>1.11.2</jmh.version>
+    <jmh.version>1.20</jmh.version>
+    <!-- override default bytecode version for src/main from parent pom -->
+    <main.java.version>1.8</main.java.version>
+    <main.signature.artifact>java18</main.signature.artifact>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
   </properties>
 
   <dependencies>
     <dependency>
-      <groupId>com.netflix.feign</groupId>
+      <groupId>io.github.openfeign</groupId>
       <artifactId>feign-core</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.netflix.feign</groupId>
+      <groupId>io.github.openfeign</groupId>
       <artifactId>feign-okhttp</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.github.openfeign</groupId>
+      <artifactId>feign-jackson</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.github.openfeign</groupId>
+      <artifactId>feign-java8</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -24,6 +24,7 @@
     <version>9</version>
   </parent>
 
+  <groupId>io.github.openfeign</groupId>
   <artifactId>feign-benchmark</artifactId>
   <version>9.6.0-SNAPSHOT</version>
   <name>Feign Benchmark (JMH)</name>
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -40,22 +40,22 @@
 
   <dependencies>
     <dependency>
-      <groupId>io.github.openfeign</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>feign-core</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>io.github.openfeign</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>feign-okhttp</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>io.github.openfeign</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>feign-jackson</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>io.github.openfeign</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>feign-java8</artifactId>
       <version>${project.version}</version>
     </dependency>
--- a/jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
+++ b/jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
@@ -1,0 +1,168 @@
+/**
+ * Copyright 2012-2018 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package feign.jackson;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.*;
+import feign.Response;
+import feign.Util;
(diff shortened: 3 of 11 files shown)


phymbert:
I dont get why build failed, because license is there in DecoderIteratorsBenchmark.java. @velo Any idea ?

> [WARNING] Missing header in: /home/travis/build/OpenFeign/feign/benchmark/src/main/java/feign/benchmark/DecoderIteratorsBenchmark.java

phymbert:
Well looks like simple EOL issue with license headers. Fixed now.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ### Version 9.6
 * Feign builder now supports flag `doNotCloseAfterDecode` to support lazy iteration of responses.
+* Introduces `JacksonIteratorDecoder` in `feign-jackson` which provides a closeable iterator for lazy processing json arrays. This decoder has to be used with `Feign.Builder.doNotCloseAfterDecode()` option disabled. Built-in iterator auto closes the `Response` when it reached json array end or failed to parse stream.
```
adriancole: wording tip. Changelog is highlight not how-to. Your first sentence is fine assuming javadoc or readme elsewhere says the how-to. Same thing below

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,168 @@
...
+      return new JacksonIteratorDecoder(new ObjectMapper()
+          .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+          .registerModules(modules));
+    }
+    public static JacksonIteratorDecoder create(ObjectMapper objectMapper) {
+      return new JacksonIteratorDecoder(objectMapper);
+    }
+  }
+
+  private final class JacksonIterator<T> implements Iterator<T>, Closeable {
```
adriancole: can't tell by eyeballing, but seems this type can be static

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,168 @@
...
+    private final Response response;
+    private final JsonParser parser;
+    private final ObjectReader objectReader;
+
+    private T current;
+
+    private JacksonIterator(Type type, ObjectMapper mapper, Response response, Reader reader)
```
adriancole: one thing you'll notice in other feign code is that we infrequently use private access as the modifiers bloat bytecode and sharing packages isn't supported outside feign anyway. if you feel strongly about hiding things, making the type private but default modifiers on the ctor and symbols accomplishes the same.

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,168 @@
...
+      throw new IllegalArgumentException("Not supported type " + type.toString());
+    }
+    ParameterizedType parameterizedType = (ParameterizedType) type;
+    if (!Iterator.class.equals(parameterizedType.getRawType())) {
+      throw new IllegalArgumentException("Not an iterator type " + parameterizedType.getRawType().toString());
+    }
+    return ((ParameterizedType) type).getActualTypeArguments()[0];
+  }
+
+  public static final class Factory {
```
adriancole: I think you'll want to flatten the factory methods into the enclosing type. There's no need for a nested type here

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,168 @@
...
+ * Feign.builder()
+ *   .decoder(JacksonIteratorDecoder.Factory.create())
+ *   .doNotCloseAfterDecode() // Required to fetch the iterator after the response is processed, need to be close
+ *   .target(GitHub.class, "https://api.github.com");
+ * interface GitHub {
+ *  {@literal @}RequestLine("GET /repos/{owner}/{repo}/contributors")
+ *   Iterator<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
+ * }</code></pre>
+ */
+public class JacksonIteratorDecoder implements Decoder {
```
adriancole: final unless you designed this for subclassing

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,168 @@
...
+      return new JacksonIterator<Object>(actualIteratorTypeArgument(type), mapper, response, reader);
+    } catch (RuntimeJsonMappingException e) {
+      if (e.getCause() != null && e.getCause() instanceof IOException) {
+        throw IOException.class.cast(e.getCause());
+      }
+      throw e;
+    }
+  }
+
+  private Type actualIteratorTypeArgument(Type type) {
```
adriancole: looks like this can be static

java8/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -0,0 +1,110 @@
...
+ * Feign.builder()
+ *   .decoder(StreamDecoder.Factory.create(JacksonIteratorDecoder.Factory.create()))
+ *   .doNotCloseAfterDecode() // Required for streaming
+ *   .target(GitHub.class, "https://api.github.com");
+ * interface GitHub {
+ *  {@literal @}RequestLine("GET /repos/{owner}/{repo}/contributors")
+ *   Stream<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
+ * }</code></pre>
+ */
+public class StreamDecoder implements Decoder {
```
adriancole: same general feedback on this type as the other

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
adriancole: there's a fair amount of logic here, so a unit test class will help. you can look at https://github.com/Netflix/denominator/blob/master/model/src/test/java/denominator/common/PeekingIteratorTest.java as an example (just of generically testing an iterator) but the test cases will be a bit specific. Once your iterator type is made package protected, you can increase coverage significantly by using it in isolation. This might involve less work in the constructor (ex doing things before you call the ctor vs here).

Often, I look at the main project themselves to see if they have an similar test. For example, jackson might have something similar in nature to borrow ideas from.

phymbert:
Updated code with review, thank you again and again @adriancole for your time and advices.
Build failed, but license header is present. It failed for obscure reason on the CI server. It works fine on my env both linux and windows. Can you please help @velo ?

velo:
Thanks for spotting it @phymbert 
https://github.com/OpenFeign/feign/pull/653

kdavisk6:
Forgive my ignorance here, but is it appropriate to add this to the core `feign` library when there is a `feign-java8` module?   Does that feel like a better fit?

kdavisk6:
Apologies, ignore my comment.  I see it is in the `java8` module.  Carry on :)

benchmark/pom.xml
```diff
@@ -24,28 +24,40 @@
     <version>9</version>
   </parent>
 
-  <!-- TODO: change group id when 9.0 is released -->
-  <groupId>com.netflix.feign</groupId>
```
adriancole: we still should be under the io.github.openfeign group ID, not the sonatype one. you'll want to re-add this I think

jackson/src/test/java/feign/jackson/JacksonIteratorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        .reason("OK")
+        .headers(Collections.<String, Collection<String>>emptyMap())
+        .body(inputStream, jsonBytes.length)
+        .build();
+
+    try {
+      thrown.expect(DecodeException.class);
+      assertThat(iterator(Boolean.class, response)).hasSize(1);
+    } finally {
+      assertThat(closed.get()).isTrue();
```
adriancole: nice one

java8/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+  @Override
+  public Object decode(Response response, Type type)
+      throws IOException, FeignException {
+    if (!(type instanceof ParameterizedType)) {
+      throw new IllegalArgumentException(
+          "StreamDecoder supports only stream: unknown " + type);
+    }
+    ParameterizedType streamType = (ParameterizedType) type;
+    if (!Stream.class.equals(streamType.getRawType())) {
+      throw new IllegalArgumentException(
```
adriancole: nit https://github.com/square/java-code-styles (which I think we use unless I'm mistaken)  has 100char line wraps. It is easier to read when we don't wrap so often

java8/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+          "StreamDecoder supports only stream: unknown " + type);
+    }
+    Iterator<?> iterator = (Iterator) iteratorDecoder.decode(response,
+        new IteratorParameterizedType(streamType));
+
+    return StreamSupport.stream(
+        Spliterators.spliteratorUnknownSize(iterator,
+            Spliterator.DISTINCT | Spliterator.NONNULL), false)
```
adriancole: NONNULL makes sense, but DISTINCT? I'm not sure we want this (as it implies buffering to ensure the distinctness). however if we do, we should test it and document that not only is this a stream, but a distinct element one

java8/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+        .onClose(() -> {
+          if (iterator instanceof Closeable) {
+            ensureClosed((Closeable) iterator);
```
adriancole: you might need mock to test this. otherwise it is eyeballing (which may be ok)

phymbert:
Updated following final review. Please confirm

velo:
Taking a look on this

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,164 @@
...
+    return new JacksonIteratorDecoder(new ObjectMapper()
+        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+        .registerModules(modules));
+  }
+
+  public static JacksonIteratorDecoder create(ObjectMapper objectMapper) {
+    return new JacksonIteratorDecoder(objectMapper);
+  }
+
+  static final class JacksonIterator<T> implements Iterator<T>, Closeable {
```
velo: I think we should make this class `public` and we should suggest people to use `JacksonIterator` (which I would also rename to `CloseableIterator`, optional)

Anyway, on the javadoc we should recommend using the `JacksonIterator`, compiler will warn you about closing it.

In the end, I think the `Iterator` should be dealt with similar to `Stream`  https://github.com/OpenFeign/feign/pull/651/files#diff-ee67453098d0b7bbe061ae7c7826ea96R74

jackson/src/main/java/feign/jackson/JacksonIteratorDecoder.java
```diff
@@ -0,0 +1,164 @@
...
+ * <p>
+ * <p>Example: <br>
+ * <pre><code>
+ * Feign.builder()
+ *   .decoder(JacksonIteratorDecoder.create())
+ *   .doNotCloseAfterDecode() // Required to fetch the iterator after the response is processed, need to be close
+ *   .target(GitHub.class, "https://api.github.com");
+ * interface GitHub {
+ *  {@literal @}RequestLine("GET /repos/{owner}/{repo}/contributors")
+ *   Iterator<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
```
velo: IMHO, we must tell people to use `JacksonIterator` and manage it.  By manage I mean close is when done.

My concern is someone partially reading the `Iterator` and leaving it open. So we recommend people using the `JacksonIterator`, even tough you can use simple `Iterator` at your own risk.

phymbert:
Thank you @velo for approving these changes.

From your review, I don't clearly understand what **feign** user should finally declare, could you please detail:
```java
  <T> FeignCloseableIterator<T> feignIt2();
  <T, I extends Closeable & Iterator<T>> I feignIt3();
  <T, I2 extends AutoCloseable & Iterator<T>> I2 feignIt4(); // java7 only ?
  <T> Iterator<T> feignIt5();
```
IMHO feignIt5 is more readeable but sure you have to read feign doc.

velo:
I agreed that `feignIt5` is more readable. But, you need to be sure of what you are doing (no partial read)
That is why I think we should recommend `feignIt2`. 

Maybe, include all  `feignIt2` and `feignIt5` with a disclaimer that `feignIt5` may leave the connection open on partial reads.

Lemme know what you think

adriancole:
Just a note though rationale is ok as you will be merging this.

Usually I dont make anything public unless there are multiple users asking
for it for reasons including effective java least visibility and also
maintenance.

Notably encouraging use of a codec specific iterator is kindof meh because
these are business interfaces. It is locking those interfaces into a very
specific impl. I wouldnt encourage it, as the cure is worse than the
disease. Plus there are other ways like error-prone annotations. In other
words through docs or other mechanisms you can probably find a way to not
further extend feign into user result types.

adriancole:
Think of it this way. There is exactly the same concern in moshi, gson or
any other streaming json not to mention xml but to the point is.. if
closing an iterator is a concern to take up in feign even if I think it is
early just consider that business interfaces who currently have no idea
what library is used for json.. they will ;) and they will be compile bound
to it.

Knowing all of this if you still want to make public go for it, just
mentioning some things that might not be obvious

velo:
Make sense, will merge as it is later tonight

PR #653, opened 2018-03-06
Master build was failing due to attempt to build benchmark

velo:


velo:
I will merge this since master is broken ATM

adriancole:
good save. note there are other things in that file that mention benchmark. they won't fail until publish :D

PR #658, opened 2018-03-17
Enhance JAXBContextFactory to accept a list of binding classes.

wolfch:

of binding classes.

Add concurrent unit test to verify integrity of pojo list-to-context map.

### DIFF
--- a/jaxb/pom.xml
+++ b/jaxb/pom.xml
@@ -27,5 +27,10 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.vmlens</groupId>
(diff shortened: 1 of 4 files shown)


wolfch:
Gonna close & re-open to try to trigger a build without a dummy push.

wolfch:
I keep getting the following build error:
```
    [ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project feign-jaxb:
    Some files do not have the expected license header -> [Help 1]
```
Even  though all the files I added have that standard Netflix copyright comment header.

wolfch:
Will try again.

PR #659, opened 2018-03-17
Enhance JAXBContextFactory to accept a list of binding classes.

wolfch:

of binding classes.

Add concurrent unit test to verify integrity of pojo list-to-context map.

Update to new license text.


wolfch:
*** _DO NOT MERGE YET_  ***.  I may have improperly trimmed merge commit logs, I will verify and re-open.

PR #660, opened 2018-03-17
Enhancement to JAXBContextFactory to accept a list of binding classes.

wolfch:

of binding classes.

### DIFF
--- a/jaxb/pom.xml
+++ b/jaxb/pom.xml
@@ -42,5 +42,10 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.vmlens</groupId>
+      <artifactId>concurrent-junit</artifactId>
+      <version>1.0.2</version>
+    </dependency>
   </dependencies>
 </project>
--- a/jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
+++ b/jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
@@ -13,10 +13,15 @@
  */
 package feign.jaxb;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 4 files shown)


jaxb/pom.xml
```diff
@@ -42,5 +42,10 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.vmlens</groupId>
+      <artifactId>concurrent-junit</artifactId>
+      <version>1.0.2</version>
```
velo: Shouldn't the scope for this be `test`?

jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
```diff
@@ -66,22 +81,31 @@ private void setMarshallerProperties(Marshaller marshaller) throws PropertyExcep
...
+          if (jaxbContext == null) {
+              jaxbContext = JAXBContext.newInstance(classes);
+              this.jaxbContexts.putIfAbsent(hashCode, jaxbContext);
+          }
+      }
+      return jaxbContext;
   }
 
+  public Set<Class<?>> getJaxbClasses() {
+      return jaxbClasses;
```
velo: Shouldn't this return a copy of `jaxbClasses`?

jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java
```diff
@@ -30,28 +35,38 @@
...
     return ctx.createUnmarshaller();
   }
 
   /**
    * Creates a new {@link javax.xml.bind.Marshaller} that handles the supplied class.
    */
-  public Marshaller createMarshaller(Class<?> clazz) throws JAXBException {
-    JAXBContext ctx = getContext(clazz);
+  public Marshaller createMarshaller(Class<?>... classes) throws JAXBException {
+    if (this.jaxbClasses.size() > 0)
```
kdavisk6: Since you have this test in both `create` methods, could you move this test to `getContext` instead?

kdavisk6:
Closing as duplicate of #762

PR #667, opened 2018-03-23
Added support for custom POJO query param encoding

### LINKED ISSUE #661: Add new type of parameter for non-body Objects via an Expander-like interface for groups of parameters
We have `@Param`, `@QueryMap` and `@HeaderMap`, but it would be nice if we could pass in objects that could be expanded via reflection rather than maps (giving us the flexibility of `@QueryMap` but with compile time safety checks of our client uses).

I would use them for authorization parameters. We generate JWT-like auth headers using some passed-in credentials like account id, so a global interceptor that doesn't have access to the object passed into the method doesn't work.

I would also use this to process objects through reflection to generate non-null query parameters. Some of our rest endpoints have near a dozen optional parameters so having to pass them all in, null or not, makes code unmaintainable.

rage-shadowman:
Added support for custom POJO parsing of optional query parameters.

Updated the existing `QueryMap` annotation to be used on custom POJO objects as well as `Map`s:
```java
@RequestLine("GET /")
V find(@QueryMap(encoded = BOOLEAN_VALUE) CustomObject customObject);
```

The `CustomObject` class must have fields that are named as they shall appear in the query parameter map. The object will be used to generate a `Map<String,Object>` through reflection based on all local fields that will then be handled in the same way as a regular `QueryMap` annotated `Map` object.

You cannot have more than 1 `QueryMap` annotated parameter (you cannot even have 1 that is a `Map` and another that is a non-map POJO). There is no way to specify that some fields are encoded, while others are not, and there is no way to specify a custom `Expander` for any fields.

Fixes #661 (may also fix #636 and #620 with some local modifications).

[EDIT: Updated per changes in later commits based on pull review comments]


core/src/main/java/feign/Contract.java
```diff
@@ -282,11 +282,26 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
           checkState(data.headerMapIndex() == null, "HeaderMap annotation was present on multiple parameters.");
           data.headerMapIndex(paramIndex);
           isHttpAnnotation = true;
+        } else {
+          CustomParam customParam = findCustomParam(annotation);
```
rage-shadowman: NOTE: This checks for extensions of `@CustomParam` (as explained in the initial pull request comment) as well as the simple `annotation instanceof CustomParam`, but it does not search recursively, as I don't believe it makes sense to extend an extension. This is the only thing that my untrained eye can see in this pull request that might be "iffy".

Of course, please let me know if this (or something I didn't see) stands out as needing changes. I would love to see some sort of custom param encoder (such as this) put into feign (and from some of the other open issues/pull requests, I can see that I'm not alone).

saintf:
hi @rage-shadowman  -
I took a peek and tried to understand the problem we're trying to solve. As I see it, what all of the above have roughly in common is that they're trying to pass complex objects as parameters and have them be then placed in the request as either query parameters, path parameters or headers.

That, however, is difficult without some code like above. However, my feeling is that even now the code is too difficult. Feign has always (at least to me) been about simplicity and making things declarative (instead of having to write complex code to achieve something). Most of the internals of feign are completely unknown to the common user (i.e. passing the template/etc seems wrong to me).

Today Feign already allows you to put values into headers dynamically, as well as parameters and query strings. It even allows you to pass a map to be a query param (as long as you use `@QueryParam`)

```
interface GitHub {
  List<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);

  @RequestLine("GET /users/{username}/repos?sort={sort}&filter={filterPattern}")
 @Headers("X-Ping: {token}")
  List<Repo> repos(@Param("username") String owner, @Param("sort") String sort,@Param("token") String token,@Param("filterPattern") String pattern);

@RequestLine("GET /find")
V find(@QueryMap Map<String, Object> queryMap);
}
`
That said - I think perhaps a more natural way to pass this without having to expand it is likely where we add value - so what if we decided to have something like this:
```
class RepoDetails {
  String owner;
  String repoName;
}

class QueryObject {
  String sort;
  String filter;
}

interface GitHub {
  @RequestLine("GET /repos/{repo.owner}/{repo.repoName}/contributors")
  List<Contributor> contributors(@Param("repo") RepoDetails repoDetails);

//this should yield GET /users/{username}/repos?sort={query.sort}&filter={query.filter}
  @RequestLine("GET /users/{username}/repos")
 @Headers("X-Ping: {token}")
  List<Repo> repos(@Param("username") String owner, @QueryMap QueryObject query);
}
`

Lastly, I think mixing parameters (query/path) with headers sounds slightly dangerous and likely something to be avoided. That said - if you have the usecase perhaps the library supports it enough to let you do it if you must. I would then do something of the sort:

```
class RepoDetails {
  String owner;
  String repoName;
 String accessToken;
} 

interface GitHub {
  @RequestLine("GET /repos/{repo.owner}/{repo.repoName}/contributors")
 @Headers("X-Ping: {repo.accesToken}")
  List<Contributor> contributors(@Param("repo") RepoDetails repoDetails);
}
`
This way everything remains declarative - with very little code to write from any of the users.

What do you think?

**Comment by rage-shadowman** [comment]
Regarding template access, if you write an `Encoder` or `Decoder` you get access to both the parameter object and the template (but only for un-annotated "body" parameters, of which you can only have 1) and if you write a `RequestInterceptor` you get access to the template (but no parameter objects). So you already have to know something about the template in order to use Feign's advanced functionality.

I could rename `CustomParam` to `ObjectParam` and include some `HeaderParam(header = "X-Header-Name", expander = OptionalExpander.class)`, `QueryParam(name = "param-name", expander = OptionalExpander.class, encoded = DEFAULTS_TO_FALSE)`, and `BodyParam(encoder = OptionalEncoder.class)` with a reflection-based `ObjectParamEncoder` class. Also allowing a `HeaderMap` and `QueryMap` style annotation as well. All annotations would work on either member variables (private access is fine) or 0-arg (getter style) methods.

Would that answer your concerns?

 As far as the `"{parent.child}"` format, I could see allowing an optional `name` parameter on the `ObjectParam` annotation that would prefix all child parameters (`parentName + "." + childName`).

I don't like the idea of parsing method names to determine the parameter inclusion characteristics, so I'd stick with annotations. I also like the idea of this object parameter not necessarily knowing all of the possible params ahead of time (allowing for "map" style parameters).

**Comment by kdavisk6** [comment]
My gut reaction to this is that addition of the `encoder` option in the annotation feels awkward and may be too complicated as @saintf has mentioned.  Also, adding an additional annotation where the general purpose of `QueryMap` feels like this **should** work here, makes we wonder, could we update `@QueryMap` to support `Map` and `Object` values?

In the case of an `Object`, we can use a technique similar to what you and @DanielYWoo have suggested to expand the object.  Users can then do something like the following

```java
public class SearchDetails {
	public String term;
	public List<String> filters;
	public String sortOn;
	public String sortDirection
}

interface Search {
        // the result here would be `/api/search?term={term}&filters={filter[0]}&filters={filter[1]}...&sortOn={sortOn}&sortDirection={sortDirection}
	@RequestLine("GET /api/search")
	List<SearchResults> search(@QueryMap SearchDetails searchDetails);
}
```

To me, this fits more within the *spirit* of the change you are requesting without making significant modifications to the current usage patterns.  Thoughts?

saintf:
So do we then agree that in this change (at least for first round) we focus on expanding `@QueryMap` to take on objects and expanding the parameters (not just taking a map)?

I'd vote for that as an initial change - then look at the next set of desires as separate changes (equally happy to get my hands dirty and do some coding if you'd like some help @rage-shadowman)

rage-shadowman:
@saintf I think that sounds good. I'd prefer to call it something else, but if that's the consensus, I'll try to get that working and submit the changes. I'll still need to hack in an authorization interceptor for my needs, but I'll keep that code separate. This should be a great start for what we need.

I'd also like to make it work only on the object's fields or methods annotated with `@Param`, rather than parsing the getter names like a bean or assuming all fields are included as named. This way we can also specify whether or not a parameter is encoded and if a custom `Expander` is needed. Is there any consensus on this?

kdavisk6:
I think that adding additional annotations to fine tune the encoding would be a welcome feature, however; let's start small.  Referring back to [The Rule of Three](https://blog.codinghorror.com/rule-of-three/), 

> a reusable component should be tried out in three different applications before it will be sufficiently general to accept into a reuse library.

we should focus on adding basic support for Objects via `@QueryMap` and see how adoption goes.  If it goes well, and with this particular ask I don't see that as being an issue, we can expand on it.  Sound good?

core/src/main/java/feign/ObjectParamMetadata.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+class ObjectParamMetadata {
+
+  private final static ConcurrentMap<Class<?>, ObjectParamMetadata> classToMetadata =
+      new ConcurrentHashMap<Class<?>, ObjectParamMetadata>();
```
rage-shadowman: I'm not sure if this cache is a case of premature optimization or not.

rage-shadowman:
I believe I have followed the prior input. Please look again and let me know what you think.

If this is good, then I can start over from a clean branch so this is all done in 1 commit if you like (to make the commit history cleaner).

core/src/main/java/feign/ObjectParamMetadata.java
rage-shadowman: ...or if it belongs in `MethodMetadata` instead.

rage-shadowman:
Thanks for the comments. I do think this simplifies things and that makes it better. A `Builder` pattern for your customo POJO objects should allow this to work for anything that just needs to modify query params (headers are another matter, but I think actually allowing `RequestInterceptor`s access to the method arguments would be the better solution there).

rage-shadowman:
So, where does this stand? Is it waiting for votes? Waiting for @adriancole or another maintainer to review?

velo:
Yes, I need to take a look on this, just busy this week =)

core/src/main/java/feign/Contract.java
```diff
@@ -132,6 +134,10 @@ protected MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method me
     private static void checkMapString(String name, Class<?> type, Type genericType) {
       checkState(Map.class.isAssignableFrom(type),
               "%s parameter must be a Map: %s", name, type);
+      checkMapKeys(name, genericType);
+    }
+
+    private static void checkMapKeys(String name, Type genericType) {
```
saintf: this looks like a good add - but I didn't see a test covering it in the case that it isn't a string (I assume tests already cover the case where it is)

README.md
```diff
@@ -443,6 +443,27 @@ A Map parameter can be annotated with `QueryMap` to construct a query that uses
...
+public class CustomPojo {
+  private final String name;
+  private final int number;
+
+  public CustomPojo (String name, int number) {
+    this.name = name;
+    this.number = number;
+  }
+}
+```
```
saintf: nice docs.

We should discuss with the wider audience if we want all member variables (including non-exposed private ones), or if we want those that are exposed in some way (public or have a getter), and if we handle transient ones or not. It's worth asking as I see below in the code that we're changing access modifiers on the fly (making accessible if not) - something that may not be something we do out-of-the box or so...

core/src/main/java/feign/ObjectParamMetadata.java
saintf: I think I start getting the point of this class. Is it basically a `QueryObjectParamMapper` or so? as in - it's job is to, for a single object type, generate a `Map<String,Object>`?

If so, I think I get where you're headed and would suggest the following changes:
* Rename it to something closer to the description above. I like `QueryObjectParamMapper`, but if you think of a better one, thumbs up.
* The caching is likely not a premature optimisation (based on how many other places we try to figure out a class and then cache its metadata to not have it done again) - but is likely in the wrong place (why would it be done in this object?). Perhaps moving it to the place where it gets called from (`ReflectiveFeign`?) is a more appropriate fit? It also doesn't have to be `static` there - which is usually a bit of a code-smell.
* likely rename `toQueryMap` to something more like `mapObject(object)` or `convertToQueryMap` - the mapper isn't becoming a query map itself - it's converting an object into one - so make it that descriptive
* likely rename `getMetadata` to `generateMapper` (if you end up calling this a mapper) - and you can likely then make it the single-entry point on startup that I mentioned.

Thoughts?

saintf: oh! and given the somewhat more complex nature - add some unit tests specifically to it - to make sure you capture intended behaviour and can replicate error cases easier. (test both correct behaviour and expected error cases)

core/src/test/java/feign/FeignTest.java
```diff
@@ -384,6 +383,48 @@ public void queryMapValueStartingWithBrace() throws Exception {
         .hasPath("/?%7Bname=%7Balice");
   }
 
+  @Test
```
saintf: I like :+1: :smile: . Nice, clean, easily readable test cases.

core/src/main/java/feign/ObjectParamMetadata.java
saintf: I read some more on your code. I agree on the right place being next to `indexToExpander`. If you want I can fork off your repo, give it a try, and then create PR into your repo for those changes so you can see what I mean/accept if you like.

README.md
kdavisk6: In another PR of a similar nature, I made the suggestion that we should support objects that via the Java Beans api.  That should allow for the widest range of support and interoperability.  However straight reflection is ok too from my perspective.

rage-shadowman: I'd still argue in favor of annotating the methods/fields that you want to use with something like `@Param`. The only reason to use all fields, was for simplicity (I guess additional annotations weren't simple enough). Personally, I'm less a fan of parsing all getter method names (javabeans style) than I am of reading all fields. But if it'll get this stuff merged...

saintf: @rage-shadowman let's get more thoughts on it from a few people. It may be that the concensus is this works just fine and we like it (and then just leave it alone). It may also be that we go Java Beans style or something else. let's see what others think before taking any drastic decision.

core/src/main/java/feign/Contract.java
rage-shadowman: This wasn't an addition, it was a DRY split. I suppose I could revert the `checkMapString` method and just 2x leave the `Map` instance check even though I just checked that. The test is in `DefaultContractTest.queryMapKeysMustBeStrings`.

core/src/main/java/feign/ObjectParamMetadata.java
rage-shadowman: I named it based on the other metadata class, because that's what it *stores*. But you are right, what it *does* is turn an object into a map, so I'll rename it.

Some possibilities: `QueryObjectParamMapper`, `ObjectQueryParamMapper`, `QueryMapBuilder`, or keep the cache internal and call it `QueryMapFactory` (either keep it static, or make it an interface default implementation that you can add as a `Feign.Builder` option and pass it into the `ReflectiveFeign` constructor)? It actually is building a `Map` of query params (so it is a `QueryMapBuilder` or "factory").

How does that sound?

README.md
rage-shadowman: If the map builder has an interface that can be passed into the feign builder, then I could even write my own in my local client that would use my annotations without complicating anything internal to feign. We could have an EJB implementation that parses getters as a separate possibility.

core/src/main/java/feign/ObjectParamMetadata.java
rage-shadowman: @saintf I don't mind working on this. It's a good learning experience for me. But if you want to write some code on it, I'm happy to do look at that too.

README.md
rage-shadowman: I would probably call it a `QueryMapBuilder` or `ObjectQueryMapBuilder` (to differentiate it from other serializers, such as the method param serializer `Expander`, or the body param serializer `Encoder`), but I like this idea. It could just have a `Map<String, Object> toQueryMap(Object)` method (or whatever better name you guys can come up with).

kdavisk6: I agree that a Serializer would fit nicely, so as long as we have a sensible default Serializer.

rage-shadowman: I created a `QueryMapEncoder` that can be specified via the `Feign.Builder`. I think it answers this concern and cleans up the naming issues.

I haven't pushed it into this pull request as I don't want to jump the gun on this. You can check that out in https://github.com/rage-shadowman/feign/tree/feature/shadowman/query-param-encoder

Should I push that into this pull request?

kdavisk6: @rage-shadowman I'd add your encoder to the PR and suggest that it be used by default, without any additional customization to start.

rage-shadowman: Done.

velo:
I was about to merge this, but there is one conflict. Ping me once fixed and I will hit the button.

rage-shadowman:
Fixed the conflicts. Though I am unable to test it as your master fails tests for me right now (and thus so does my branch now that I've merged your master back into it).

velo:
> Fixed the conflicts. Though I am unable to test it as your master fails tests for me right now (and thus so does my branch now that I've merged your master back into it).

Can you detail the failures and your system?
May be is a OS related issue and may be relevant to fix it.

rage-shadowman:
It had to do with redirects and retries (the retry tests failed because of a permanent redirection response). But now I'm at work, and it's working, so I'm guessing it must have been a DNS issue.

PR #668, opened 2018-03-26
Response is closed after decoder fails

cakofony:
Previously a decoder failure resulting in an exception would
fail to close responses when "doNotCloseAfterDecode" was enabled.


phymbert:
Can you please attach a unit test also ?

cakofony:
Updated with a test that fails against master :-)

PR #670, opened 2018-03-28
Updated README.md

SCrusader:
Made "Why Feign and not X" and "How does Feign work" more formal and concise.


kdavisk6:
You have my vote 👍

Issue #669, opened 2018-03-27
Unsupported JAX-RS 2.x annotations render entire interface useless


masc3d:
Causing ```IllegalStateException: Method has too many Body parameters``` when creating target.

I have noticed async is not supported (yet). Until then it would be beneficial to simply ignore them so regular endpoints of the interface remain accessible.

Maybe there is a way to mitigate this in 9.6.1 and I just haven't found it. Is there any?

masc3d:
`final class JAXRSContract` 😢

masc3d:
The same applies to `@Context` btw which is used for SSE endpoints.

kdavisk6:
I looked over your change, why not submit a PR for feedback?  From my perspective it looks good.  I would add a test case or update the existing test cases to verify.

masc3d:
`feign-jaxrs2` is available in `9.7`

PR #672, opened 2018-03-31
FIXED unsupported jaxrs-2.1 annotations should not break entire interface

masc3d:
resolving #669

### DIFF
--- a/httpclient/pom.xml
+++ b/httpclient/pom.xml
@@ -59,6 +59,7 @@
       <artifactId>jsr311-api</artifactId>
       <version>1.1.1</version>
       <groupId>javax.ws.rs</groupId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
--- a/httpclient/pom.xml
+++ b/httpclient/pom.xml
@@ -56,6 +56,12 @@
     </dependency>
 
     <dependency>
+      <artifactId>jsr311-api</artifactId>
+      <version>1.1.1</version>
+      <groupId>javax.ws.rs</groupId>
+    </dependency>
+
+    <dependency>
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>mockwebserver</artifactId>
       <scope>test</scope>
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -38,8 +38,8 @@
 
     <dependency>
       <groupId>javax.ws.rs</groupId>
-      <artifactId>jsr311-api</artifactId>
-      <version>1.1.1</version>
+      <artifactId>javax.ws.rs-api</artifactId>
+      <version>2.1</version>
     </dependency>
 
     <!-- for example -->
--- a/jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
+++ b/jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
@@ -13,24 +13,14 @@
  */
 package feign.jaxrs;
 
+import feign.Contract;
+import feign.MethodMetadata;
+
+import javax.ws.rs.*;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
-
-import javax.ws.rs.Consumes;
-import javax.ws.rs.FormParam;
-import javax.ws.rs.HeaderParam;
-import javax.ws.rs.HttpMethod;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
-import javax.ws.rs.container.Suspended;
-import javax.ws.rs.core.Context;
-
-import feign.Contract;
-import feign.MethodMetadata;
 
 import static feign.Util.checkState;
 import static feign.Util.emptyToNull;
@@ -39,7 +29,7 @@ import static feign.Util.emptyToNull;
  * Please refer to the <a href="https://github.com/Netflix/feign/tree/master/feign-jaxrs">Feign
  * JAX-RS README</a>.
  */
-public final class JAXRSContract extends Contract.BaseContract {
+public class JAXRSContract extends Contract.BaseContract {
 
   static final String ACCEPT = "Accept";
   static final String CONTENT_TYPE = "Content-Type";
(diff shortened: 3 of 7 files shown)


jaxrs/pom.xml
```diff
@@ -38,8 +38,8 @@
 
     <dependency>
       <groupId>javax.ws.rs</groupId>
-      <artifactId>jsr311-api</artifactId>
-      <version>1.1.1</version>
+      <artifactId>javax.ws.rs-api</artifactId>
```
spencergibb: Hmm. Dangerous thing to do in a minor, changing a dependency like this

masc3d: how? jax-rs-2.1 is fully backwards compatible down to 1.1.

velo:
Some testing would be nice, but my main concern is breaking for people using Jax RS 1

kdavisk6:
Agreed.  Changing a dependencies major version may cause users of the client headaches if they aren't prepared when they update.

masc3d:
while I don't share any concerns for breakage technically, I would agree it's formally not appropriate to bump this spec as part of a minor patch.

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
```diff
@@ -124,7 +126,12 @@ protected boolean processAnnotationsOnParameter(MethodMetadata data, Annotation[
     boolean isHttpParam = false;
     for (Annotation parameterAnnotation : annotations) {
       Class<? extends Annotation> annotationType = parameterAnnotation.annotationType();
-      if (annotationType == PathParam.class) {
+      // masc20180327. parameter with unsupported jax-rs annotations should not be passed as body params.
+      // this will prevent interfaces from becoming unusable entirely due to single (unsupported) endpoints.
+      // https://github.com/OpenFeign/feign/issues/669
+      if (annotationType == Suspended.class || annotationType == Context.class) {
```
velo: Can we use a String check here?

`annotationType.getClass().equals("javax.ws.rs.container.Suspended")`
Why? To prevent class def not found for anyone that still wanna use jax-rs 1

jaxrs/pom.xml
velo: I think one way to address this, is to keep the original dependency and add the new one as provided.

For those using jax-rs 2, they would need to explicitly declare the dependency. And for the ones already using jax-rs 1 no need for changes.

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
masc3d: yes that may work, but it's ugly and prone to breakage on maintenance so I won't supply this. I'd rather have `feign-jaxrs-2`. also I don't see any reason (for anyone) to even include `jax-rs-1`.

masc3d: how do you plan to support `jax-rs-2` in feign in general?

spencergibb:
>  I don't see any reason (for anyone) to even include jax-rs-1.

Except for the people who can't or aren't willing to upgrade.

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
velo: > it's ugly and prone to breakage on maintenance
agree

may be a `try-catch` for the `ClassDefNotFound`

masc3d: @velo is it universally predictable when jvm's throw this?

masc3d: I would assume this is thrown by jvm before `processAnnotationsOnParameter` is invoked even, but I could be wrong.

jaxrs/pom.xml
masc3d: @spencergibb how is this a minor btw? I'm actually considering jax-rs-2.x breakage pretty much a major considering there's no applicable workaround and the spec has been around for half a decade by now. In consideration of 2.x being backwards compatible thus applicable as drop-in substitute, would make  jsr-311 legacy support pretty much a minor instead.

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
velo: I made a PR with an "example" and the exception as well
https://github.com/masc3d/feign/pull/1

rage-shadowman: Not sure if it's overkill or not, but you could always use the maven-invoker-plugin to test compatibility.

kdavisk6:
@masc3d have you considered creating a `feign-jaxrs-2` library?  This approach could be more direct and address any concerns others changing the dependencies in `feign-jaxrs`.  If creating a new library just for these changes seems too much, another option could be to release a new major version of `feign-jaxrs` making the move to 2.x explicit.  

Thoughts?

spencergibb:
I think those are the only two options

masc3d:
@kdavisk6 yes I proposed this one comment above, still waiting for feedback.

masc3d:
> how do you plan to support jax-rs-2 in feign in general?

if you can agree on a major version bump anytime soon it probably doesn't make sense to spawn a new artifact now.

kdavisk6:
I'm in the camp of creating a major release of `feign-jaxrs`.

spencergibb:
It would be the whole project since modules are not released independently

kdavisk6:
Oh, I was not aware of that.

masc3d:
I decided I won't provide a jsr311 compatible setup. those consumers which are allegedly adversely affected in a way that can't be easily mitigated probably don't even exist.

kdavisk6:
After reading into this further, it's difficult to say with any confidence how many users of Feign are still using JAX-RS 1.x (jsr311) libraries.  With that in mind, my suggestion would be deprecate `feign-jaxrs` and create a new `feign-jarxrs2` library with first class JAX-RS 2.x support and beyond.

This will allow users of the library who are still on JSR-311 to receive core feign updates and those that want to take advantage of JAX-RS 2.0 features the ability to do so.  Thoughts?

httpclient/pom.xml
```diff
@@ -55,6 +55,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
velo: I don't think this is being used, is it?

jaxrs2/README.md
```diff
@@ -0,0 +1,2 @@
+# Feign JAXRS 2
+
```
velo: Can you just "copy" README from jaxrs?!

httpclient/pom.xml
masc3d: it's required for test as `jsr311-api` is not pulled in transitively any longer. will add `scope`.

Issue #656, opened 2018-03-17
Is it time to move on? Should feign move to `java 8`?


velo:
Right now, feign is compatibility is bound to `java 6`.

Is it time to release feign 10.0 with `java 8`? (10.x was @spencergibb suggestion)

@kdavisk6 suggested we start a discussion about it.

adriancole:
Up to you, but if this changes, please change the group ID and the package
ID as it won't be bytecode compatible anymore so definitely a major
release. Decisions like this are open, but because the collateral damage
extends well beyond those watching issues, just be very cautious if
implementing this. We don't want to create work for "innocent bystanders"
and changing package and group Id helps with this.

adriancole:
ps @spencergibb <https://github.com/spencergibb> might disagree with me
about if changing package names helps or not. It does invalidate other
layered tools.. so don't take my word as gospel

spencergibb:
Spring doesn't change group id or artifact id for minimum Java versions

adriancole:
>
> Spring doesn't change group id or artifact id for minimum Java versions
>
I guess what I mean to ask isn't so much about spring code, but spring
cloud impact. When feign is rewritten to java 8 it will not be compatible
anymore with java 7. Is this an issue for spring cloud? If so, maybe group
id, package. If not, then maybe unnecessary.

At any rate the android support statements should be dropped if this
happens, so yeah just update that part. In other projects I've avoided
bumping to java 8 (and dealing with this) via retrolambda. Anyway there are
a few folks who will be invalidated. For example, anyone using feign inside
a metrics library as an exporter, it will invalidate the ability to use it
(which might be fine as they just put a <10 update comment in their version
matrix).

velo:
> the android support statements should be dropped if this happens.
In my mind, android had java 8 support
https://developer.android.com/studio/write/java8-support.html
Am I missing something?

Yeah, https://github.com/orfjackal/retrolambda seems a good option.
We could even build the official packages with java 8 and offer retrofitted java 6 jars using classifier.

adriancole:
>
> In my mind, android had java 8 support
> https://developer.android.com/studio/write/java8-support.html
> Am I missing something?
>
>  There's an SDK version level concern. For example, the current code is
compatible with older versions
https://github.com/Netflix/denominator/blob/master/example-android/build.gradle#L14

spencergibb:
Nope, spring cloud (and framework and boot) are now all baselined to Java 8.

[...]

PR #704, opened 2018-05-16
Updated POM to set compiler and signatures to Java 8

kdavisk6:
* Updated the Cipher Suites in `TrustingSSLSocketFactory` to an appropriate
Java 8 Cipher.  
* Updated Versions to 10.0.0-SNAPSHOT

In support of #656

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.github.openfeign</groupId>
     <artifactId>parent</artifactId>
-    <version>9.8.0-SNAPSHOT</version>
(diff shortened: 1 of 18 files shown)


velo:
Now, shall we include retro lambda on this and at least try to keep java 6 support for a while, or just drop it?

kdavisk6:
I think that would a good start.

velo:
Absolutely.

I branched `9.x`

May be, just create a ticket about `retrolambda` and see if anyone expressed the desire to make it happen.

spencergibb:
I don't think you need retro lambda. If they need six, use the 9.x branch.

PR #707, opened 2018-05-19
Build feign with java 11

velo:
Since we move the build to java 8 and feign version to 10, I decided to verify if feign builds on the next LTS jdk 11.

For now, I exclude `jaxb` cause I can't get it working


kdavisk6:
Looks good.  I would suggest that we add some additional documentation to [RELEASE](https://github.com/OpenFeign/feign/blob/master/RELEASE.md) as well.

velo:
@kdavisk6 this shouldn't change the release process... just build... I added more info to the `CHANGELOG`

PR #708, opened 2018-05-24
modify Feign's connectTimeout and readTimeout props' priority while use Ribbon

caojiajun:
while use Feign-Ribbon, the conf for ribbon's readTimeout and connectTimeout prop will be override by Feign's Request.Options which is fixed from Feign.Builder. After the commit, if use the Feign's Request.Options' DEFAULT, readTimeout and connectTimeout will use Ribbon's readTimeout and connectTimeout prop which is Dynamic.

### DIFF
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -13,11 +13,6 @@
  */
 package feign;
 
-import java.io.IOException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.List;
 import feign.Logger.NoOpLogger;
 import feign.ReflectiveFeign.ParseHandlersByName;
 import feign.Request.Options;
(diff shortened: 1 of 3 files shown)


kdavisk6:
This is a very similar change to the one proposed in #554.  The same concerns apply here.  By providing a custom `RequestConfig` you are bypassing what is being set at the `Feign` level.  Should we consider refactoring the `Options` object into an interface so individual clients like Ribbon and ApacheHttp can provide this level of customization?

caojiajun:
There are two question here. 
First. Feign should not override Request Config all the time. Feign should give a way to ignore the config in Feign Level, so Ribbon/ApacheHttpClient level can use their config instead.
Second. Feign should give a more flexible Request Config as you say. As Ribbon/ApacheHttpClient have a lot of Config not only connectTimeoutMillis/readTimeoutMillis/followRedirects.

kdavisk6:
Also see #531

velo:
2 months no feedback. Closing, please re-open if necessary

PR #710, opened 2018-05-26
Update pom.xml

karlnicholas:
com.marvinformatics.formatter to version 2.7.0. Maybe cause of Eclipse-STS problems, but should be updated anyway.


karlnicholas:
Closed due to failure. Will update groupId.

PR #712, opened 2018-05-26
Update pom.xml for formatter-maven-plugin.

karlnicholas:
Older plugin causing problems with Eclipse-STS on Windows.


karlnicholas:
Ughh, spelling error in groupID?

PR #713, opened 2018-05-26
Update pom.xml for formatter-maven-plugin.

karlnicholas:
Update formatter-maven-plugin to groupId net.revelc.code.formatter and version 2.7.3 because current one causes Java version problems on Eclipse-STS for windows.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -443,9 +443,9 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>com.marvinformatics.formatter</groupId>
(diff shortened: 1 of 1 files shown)


velo:
I prefer to use this one as it also configure eclipse automatically.

karlnicholas:
On my system it just breaks eclipse, tries to set the java version to "9". Kablam.

velo:
Odd, could you file a bug on https://github.com/velo/maven-formatter-plugin/issues

PR #714, opened 2018-05-26
Added testing only fixes Windows newlines.

karlnicholas:
Added fixes to junit only tests for Windows platform newlines. Pretty much just put System.lineSeparator() everywhere except in jaxb where feign returns only UNIX specific newlines.


karlnicholas:
@velo Hi, It does build and pass all tests on windows. Only issue would be the formatter plugin but that doesn't prevent building and unit testing. I forked and built feign because I was going to look into adding support for Spring Pageable for the spring.cloud.openfeign project (https://github.com/spring-cloud/spring-cloud-openfeign/issues/26) and so want to look into that at this point.

Issue #676, opened 2018-04-10
Header parameters not removed when empty


rage-shadowman:
When you leave your header parameter null, the placeholder becomes the value.

Ex:
```java
@Header("X-Some-Header: {someHeader}")
@RequestLine("GET /?queryStuff={queryStuff}")
String someHeaderTest(@Param("someHeader") String someHeader, @Param("queryStuff") String queryStuff);
```

When passing in null values (`api.someHeaderTest(null, null);`) feign will remove the `queryStuff` parameter, but the header value for "X-Some-Header" ends up retaining the placeholder value of "{someHeader}".

If the null query parameters are removed, then the null header parameters should also be removed.

PR #724, opened 2018-06-25
Remove null empty headers

fmcejudo:
https://github.com/OpenFeign/feign/issues/676

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -266,7 +266,7 @@ public final class RequestTemplate implements Serializable {
   /* roughly analogous to {@code javax.ws.rs.client.Target.request()}. */
   public Request request() {
     Map<String, Collection<String>> safeCopy = new LinkedHashMap<String, Collection<String>>();
-    safeCopy.putAll(headers);
+    safeCopy.putAll(headers());
     return Request.create(
         method, url + queryLine(),
         Collections.unmodifiableMap(safeCopy),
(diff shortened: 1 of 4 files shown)


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -535,6 +535,8 @@ public RequestTemplate headers(Map<String, Collection<String>> headers) {
    * @see Request#headers()
    */
   public Map<String, Collection<String>> headers() {
+    headers.forEach((key, value) -> value.removeIf(e -> e == null));
```
velo: I don't like the idea of modifying the existing map...

may be just filter and create  new one?!

velo:
@kdavisk6 LGTM, what do you think?

Do you foresee any issues on removing empty headers?

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -535,7 +536,16 @@ public RequestTemplate headers(Map<String, Collection<String>> headers) {
    * @see Request#headers()
    */
   public Map<String, Collection<String>> headers() {
-    return Collections.unmodifiableMap(headers);
+
+    return Collections.unmodifiableMap(
+        headers.entrySet().stream().filter(h -> h.getValue() != null && !h.getValue().isEmpty())
+            .collect(toMap(
```
velo: just use default toMap... `toMap(Entry::getKey, Entry::getValue)` no need to handle duplicated keys or map creation

fmcejudo: if I do so, it returns the map in any order, with LinkedHashMap it collects them in same order. In fact that was my first approach and it broke tests

Issue #709, opened 2018-05-24
Retryer always throws RetrybaleException but not origin cause


metalpalo:
I implemented own decoder (similar to Default, overtaken possibility to retry, see bellow) to throw custom exceptions based on status codes.
These exceptions are subclasses of _RuntimeException_(not _RetryableException_) and I would like to them to be retryable via **RETRY_AFTER** flag.
This is ensured by following routine overtaken from _ErrorDecoder.Default_, but here my exception is wrapped to _RetrybaleException_

```java
@Override
public Exception decode(String methodKey, Response response) {

    RuntimeException exception = decodeException(response);

    //overtaken from Default
    Date retryAfter = retryAfterDecoder.apply(firstOrNull(response.headers(), RETRY_AFTER));
    if (retryAfter != null) {
        return new RetryableException(exception.getMessage(), exception, retryAfter);
    }
    return exception;
}
```

I understand that such type is necessary for _SynchronousMethodHandler_ to repeat this operation.

My problem is that when the default retryer reaches _maxAttemps_ it throws _RetrybaleException_ instead of _Throwable_ see

```java 
public void continueOrPropagate(RetryableException e) {
    if (attempt++ >= maxAttempts) {            
       throw e;
    }
}
```

What whould be good for me is something like this:

```java
public void continueOrPropagate(RetryableException e) {
  if (attempt++ >= maxAttempts) {
      if (e.getCause() instanceof RuntimeException)
          throw ((RuntimeException)e.getCause());
      throw e;
  }
```
What do you this about this? Is my idea wrong?

dinomite:
From [PR730](https://github.com/OpenFeign/feign/pull/730#issuecomment-403044441)

@kdavisk6:
> However, with that said, it may reasonable to consider refactoring the proxy to read and throw the exceptions returned from the decoder directly, instead of masking them.

Kevin is describing adding a new `catch` block in `ReflectiveFeign.invoke()` to throw the exception cause:

```
try {
} catch (Throwable th) {
    throw th.getCause();
}
```

kdavisk6:
I think it would be worth while to create a preliminary PR to see the scope of this change.  I don't think it will be very far reaching, but it's worth giving it a shot.  @dinomite see if you can create a first pass at this and see if it reacts the way you need.

dinomite:
Preliminary PR https://github.com/OpenFeign/feign/pull/737

kdavisk6:
Closing via #737

PR #730, opened 2018-07-03
Make Feign's MethodHandlerFactory configurable

dinomite:
I want to be able to provide my own `MethodHandler` factory as a way to change the way retries work[1].  Currently the factory for `MethodHandler`s is always [`SynchronousMethodHandler`'s internal `Factory`](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Feign.java#L248).

This PR pulls an interface out of `SynchronousMethodHandler.Factory` (`MethodHandlerFactory`), adds a `methodHandlerFactory()` to `Feign.builder` to allow setting this, and initializes the method handler to `SynchronousMethodHandler.Factory` in `Feign.Builder.build()` if it isn't set by the caller.

[1]: Specifically, I want to be able to throw checked exceptions from my `ErrorDecoder`s (see also: #709) so that users of the client can know what exceptions might be emitted.  Providing my own `MHF` will allow me to provide my own `MethodHandler` that, along with my `Retryer`, handles checked exceptions from my `ErrorDecoders`.

### DIFF
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -109,6 +109,7 @@ public abstract class Feign {
         new InvocationHandlerFactory.Default();
     private boolean decode404;
     private boolean closeAfterDecode = true;
+    private MethodHandlerFactory methodHandlerFactory;
 
     public Builder logLevel(Logger.Level logLevel) {
       this.logLevel = logLevel;
@@ -219,6 +220,14 @@ public abstract class Feign {
     }
 
     /**
+     * Allows you to override how reflective dispatch works inside of Feign.
+     */
+    public Builder methodHandlerFactory(MethodHandlerFactory methodHandlerFactory) {
+      this.methodHandlerFactory = methodHandlerFactory;
+      return this;
+    }
+
+    /**
      * This flag indicates that the response should not be automatically closed upon completion of
      * decoding the message. This should be set if you plan on processing the response into a
      * lazy-evaluated construct, such as a {@link java.util.Iterator}.
@@ -245,12 +254,13 @@ public abstract class Feign {
     }
 
     public Feign build() {
-      SynchronousMethodHandler.Factory synchronousMethodHandlerFactory =
-          new SynchronousMethodHandler.Factory(client, retryer, requestInterceptors, logger,
-              logLevel, decode404, closeAfterDecode);
+      if (methodHandlerFactory == null) {
+        this.methodHandlerFactory = new SynchronousMethodHandler.Factory(client, retryer,
+                requestInterceptors, logger, logLevel, decode404, closeAfterDecode);
+      }
       ParseHandlersByName handlersByName =
           new ParseHandlersByName(contract, options, encoder, decoder, queryMapEncoder,
-              errorDecoder, synchronousMethodHandlerFactory);
+              errorDecoder, methodHandlerFactory);
       return new ReflectiveFeign(handlersByName, invocationHandlerFactory, queryMapEncoder);
     }
   }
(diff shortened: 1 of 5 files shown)


kdavisk6:
Your original approach outlined in #709 is more appropriate than exposing `MethodHandler` onto the public interface.  That component is part of the internals of `feign` and not intended to be used outside.  

If your intent is to use checked exceptions over `RuntimeExceptions`, the best option is to register your own `Retryer` that can inspect the `RetryException` and throw from there.  For completeness, here is an example based off of your own comments from #709 

```java
public MyRetryer extends Retryer.Default {

   @Override
   public void continueOrPropagate(RetryableException re) {
      try {
         /* let the base class do it's thing */
         super.continueOrPropagate(re);
      } catch (RuntimeExceprtion ree) {
         /* retries have been exhausted, inspect and rethrow */        
         if (ree instanceof RetryableException) {
             /* get the cause of the retry */
             Throwable cause = retry.getCause();
             if (cause instanceof RuntimeException) {
                throw cause;
             } else {
                /* it's a checked exception, but the signature does not allow for checked it, so 
                 * the exception must be wrapped in order to make it out */
                 throw new RuntimeException(cause);
            }
         }
         /* rethrow */
         throw ree;
      }
   }
}
```

dinomite:
The issue is that we don't want to lose our checked exceptions—we like having the compiler help users of our clients know what exceptions to expect.  Our clients declare checked exceptions, so callers are forced to catch them, but if a request is retried then it comes out as a `RuntimeException` which skips the `catch` block and thus the exception handling.

kdavisk6:
In that case, the appropriate change would be alter the signature of `Retryer` to support checked exceptions.  

```java
public interface Retryer {
   public void continueOrPropagate(RetryableException e) throws Throwable;
}
```

That would be a breaking change and something that would need to be discussed by the community.  

Even so, for a framework like Feign and others such as Spring Data, using an Interface to generate a client has some drawbacks.  One of them being that the generated implementation cannot support checked exceptions on the interface, only `RuntimeExceptions` can be used, since they don't have the knowledge about how they are being used.

However, with that said, it may reasonable to consider refactoring the proxy to read and throw the exceptions returned from the decoder directly, instead of masking them.

In `ReflectiveFeign`
```java
public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
   ...
   try {
   } catch (Throwable th) {
      throw th.getCause();
   }
}
```

This would then allow you to do the following 
```java
public interface MyApi
   public stuff() throws MyException;
}

try {
   myFeign.stuff();
} catch (MyException) {
...
}
```
Is this what your goal?

dinomite:
First, thanks for the thoughtful & thorough replies!

> That would be a breaking change and something that would need to be discussed by the community.

This would absolutely be preferred, but I assumed untenable as the Feign retry architecture seems built around unchecked exceptions (which I presume was a knowing choice for ease of users).  Where might I propose such a change?

> One of them being that the generated implementation cannot support checked exceptions on the interface, only RuntimeExceptions can be used, since they don't have the knowledge about how they are being used.

I'm confused by this because we throw (and catch at the point of use) checked exceptions in our generated clients.  We're only now running into issues because we want to have retries but keep checked exceptions.

> However, with that said, it may reasonable to consider refactoring the proxy to read and throw the exceptions returned from the decoder directly, instead of masking them.

I like the sound of this because it gets to exactly what I'm looking for.  Looks pretty straightforward, though I haven't looked at the tests yet, is it worth me making a PR down this path?

kdavisk6:
I'd say let's move this conversation over to the issue so we can get more visibility and feedback before continuing.

PR #732, opened 2018-07-04
Moves Java8 into Core

### LINKED ISSUE #705: Move java8 to core
Since java 8 is now the baseline. See #656

kdavisk6:
Closes #705 

- Moves the `OptionalDecoder` and `StreamDecoder` into Feign Core
- Removes `java8` library


velo:
Looks good, could you add an entry about this change on changelog?

kdavisk6:
Changelog updated!

Issue #693, opened 2018-05-01
[SOLUTION] NullPointerException when using @RequestLine("GET /wsdl/testcase?wsdl")


GordonMcKinney:
This is using Feign-Vertx but the NPE is inside the Feign library.

```
/**
 * This demonstrates the behavior of a query param without a value
 * e.g. /wsdl/testcase?wsdl
 */
public interface WsdlTestApi {
    @RequestLine("GET /wsdl/testcase?wsdl")
    Future<String> getWsdl();
}
```

NPE is thrown /com/netflix/feign/feign-core/8.18.0/feign-core-8.18.0-sources.jar!/feign/RequestTemplate.java
```
      Collection<String> values = new ArrayList<String>();
      for (String value : entry.getValue()) {
[NPE]   if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
          Object variableValue = unencoded.get(value.substring(1, value.length() - 1));
```

To fix this, use the following:
```
      Collection<String> values = new ArrayList<String>();
      for (String value : entry.getValue()) {
        // Allow a pure param to be represented, e.g. /path/resource?wsdl
        if ( null == value ) {
          values.add(null);
        }
        else if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
          Object variableValue = unencoded.get(value.substring(1, value.length() - 1));
```

kdavisk6:
The latest version of Feign is 9.6.  Have you tried updating your dependencies to the latest version and reproduce?

renaz6:
The PR for this fix can be found here: https://github.com/OpenFeign/feign/pull/733

PR #733, opened 2018-07-05
Fixed NPE in RequestTemplate.java. Updated TrustingSSLSocketFactory.java to use cipher from master.

renaz6:
This PR fixes the NPE in RequestTemplate.java described in this open issue: https://github.com/OpenFeign/feign/issues/693#issuecomment-386323127
Included, there is also a simple unit test to check for the NPE and an update to TrustingSSLSocketFactory.java to use the cipher from master.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -670,7 +670,10 @@ public final class RequestTemplate implements Serializable {
       }
       Collection<String> values = new ArrayList<String>();
       for (String value : entry.getValue()) {
-        if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
+        // Allow a pure param to be represented, e.g. /path/resource?wsdl
+        if (null == value) {
+          values.add(null);
+        } else if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
           Object variableValue = unencoded.get(value.substring(1, value.length() - 1));
           // only add non-null expressions
           if (variableValue == null) {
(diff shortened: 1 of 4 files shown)


kdavisk6:
Two notes:

1. Please submit one change per PR
2. Please submit changes agains the `master` branch.

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -670,7 +670,10 @@ void replaceQueryValues(Map<String, ?> unencoded, Map<String, Boolean> alreadyEn
       }
       Collection<String> values = new ArrayList<String>();
       for (String value : entry.getValue()) {
-        if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
+        // Allow a pure param to be represented, e.g. /path/resource?wsdl
+        if (null == value) {
+          values.add(null);
```
kdavisk6: It's more appropriate to simply skip the value if it's `null` instead of adding a `null` value.

```java
if (value == null) {
   continue;
}
```

core/src/test/java/feign/client/TrustingSSLSocketFactory.java
```diff
@@ -43,7 +43,7 @@
   private static final Map<String, SSLSocketFactory> sslSocketFactories =
       new LinkedHashMap<String, SSLSocketFactory>();
   private static final char[] KEYSTORE_PASSWORD = "password".toCharArray();
-  private final static String[] ENABLED_CIPHER_SUITES = {"SSL_RSA_WITH_3DES_EDE_CBC_SHA"};
+  private final static String[] ENABLED_CIPHER_SUITES = {"TLS_RSA_WITH_AES_256_CBC_SHA"};
```
kdavisk6: This is a java 8 only change and your PR is against our 9.x branch, which is not java 8 compatible.  Please break this out into another PR against the `master` branch.

PR #735, opened 2018-07-06
Fixed NPE in RequestTemplate.Java

renaz6:
This PR fixes the NPE in RequestTemplate.java described in this open issue: https://github.com/OpenFeign/feign/issues/693#issuecomment-386323127
Included, there is also a simple unit test to check for the NPE.


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -670,7 +670,10 @@ void replaceQueryValues(Map<String, ?> unencoded, Map<String, Boolean> alreadyEn
       }
       Collection<String> values = new ArrayList<String>();
       for (String value : entry.getValue()) {
-        if (value.indexOf('{') == 0 && value.indexOf('}') == value.length() - 1) {
+        // Allow a pure param to be represented, e.g. /path/resource?wsdl
+        if (null == value) {
+          values.add(null);
```
kdavisk6: It's preferable to skip `null` values instead of propagating them.  

```java
if (value != null && !value.isEmpty()) {
   ...
}
```

This will still result in the parameter being kept as a literal.

renaz6: We discovered on our project that requests to a SOAP server were failing with an NPE. We found the need for valueless parameter. Our unit test captures this criteria. We are keeping the null value as a literal in order to satisfy our unit test that checks for parameters without a value.

kdavisk6: This speaks to solid data validation.  Whenever possible, we should be accepting only **known good** values and validating all information that comes in from a `Target`.  However, in this case, we do not want to assume too much of how the `Target` has been configured or the intentions of the developer behind it; which leads us to adopt a **reject known bad** or **blacklist** approach.

Your solution to keep the `null` does neither, it only pushes the responsibility for validating that information down the stack to other components and functions within `Feign`.  This is not only bad practice, but it is also fragile and may produce some very difficult, hard to track down bugs if the other components do not handle this situation.   Within the `RequestTemplate`, the method `doQuery` is saving you.

With that said, I still recommend that that the `null` values are removed, not propagated, and validated.  This may require more changes to ensure that `Param` object with no value are still available on the `Request`.   It may also require you to re-think how to represent a name-only parameter.

core/src/test/java/feign/client/AbstractClientTest.java
```diff
@@ -332,6 +332,9 @@ public void testAlternativeCollectionFormat() throws Exception {
     @RequestLine("POST /?foo=bar&foo=baz&qux=")
     @Headers({"Foo: Bar", "Foo: Baz", "Qux: ", "Content-Type: {contentType}"})
     Response postWithContentType(String body, @Param("contentType") String contentType);
+
+    @RequestLine("GET /wsdl/testcase?wsdl")
+    Response getWsdl();
```
kdavisk6: This is a poor name for a test.  This does not indicate what the test actually is.  A more appropriate name would be `getRequestWithNameOnlyParameter()` or something similar.

core/src/test/java/feign/client/DefaultClientTest.java
```diff
@@ -96,4 +99,20 @@ public void canOverrideHostnameVerifier() throws IOException, InterruptedExcepti
     api.post("foo");
   }
 
+  @Test
+  public void testWsdlApi() throws InterruptedException {
```
kdavisk6: Similar comment as above.  This test is not about a `wsdl` but that that the request template can correctly generate a url where there exists a parameter with no value.

core/src/main/java/feign/RequestTemplate.java
renaz6: Thank you for this comment! I have updated the test names per your request and we will investigate your suggestions further in a future sprint.

kdavisk6:
See #778

kdavisk6:
Closed via #778

PR #737, opened 2018-07-08
Unwrap RetryableException and throw cause

dinomite:
As described in #709, this changes Feign to unwrap `RetryableException`s to throw the underlying cause in the event retrying fails.

This will require documentation updates, but I wanted to make a PR to show the changes for discussion first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,7 +727,9 @@ public class Example {
 `Retryer`s are responsible for determining if a retry should occur by returning either a `true` or
 `false` from the method `continueOrPropagate(RetryableException e);`  A `Retryer` instance will be 
 created for each `Client` execution, allowing you to maintain state bewteen each request if desired.
-If the retry is determined to be unsucessful, the last `RetryException` will be thrown.
+
+If the retry is determined to be unsuccessful, the last `RetryException` will be thrown.  To throw the original
+cause that led to the unsuccessful retry, build your Feign client with the `propagateExceptions()` option.
 
 #### Static and Default Methods
 Interfaces targeted by Feign may have static or default methods (if using Java 8+).
--- a/README.md
+++ b/README.md
@@ -729,7 +729,7 @@ public class Example {
 created for each `Client` execution, allowing you to maintain state bewteen each request if desired.
 
 If the retry is determined to be unsuccessful, the last `RetryException` will be thrown.  To throw the original
-cause that led to the unsuccessful retry, build your Feign client with the `propagateExceptions()` option.
+cause that led to the unsuccessful retry, build your Feign client with the `exceptionPropagationPolicy()` option.
 
 #### Static and Default Methods
 Interfaces targeted by Feign may have static or default methods (if using Java 8+).
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -100,15 +100,7 @@ public class ReflectiveFeign extends Feign {
         return toString();
       }
 
-      try {
-          return dispatch.get(method).invoke(args);
-      } catch (Throwable th) {
-        Throwable cause = th.getCause();
-        if (th instanceof RetryableException && cause != null) {
-          throw cause;
-        }
-        throw th;
-      }
+      return dispatch.get(method).invoke(args);
     }
 
     @Override
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -98,7 +98,16 @@ public class ReflectiveFeign extends Feign {
       } else if ("toString".equals(method.getName())) {
         return toString();
       }
-      return dispatch.get(method).invoke(args);
+
+      try {
+          return dispatch.get(method).invoke(args);
+      } catch (Throwable th) {
+        Throwable cause = th.getCause();
+        if (th instanceof RetryableException && cause != null) {
+          throw cause;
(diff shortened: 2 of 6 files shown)


kdavisk6:
@dinomite the builds is failing during some of the test cases.  Can you please look at those and update the PR?

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -98,7 +98,16 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
-      return dispatch.get(method).invoke(args);
+
+      try {
+          return dispatch.get(method).invoke(args);
+      } catch (Throwable th) {
+        Throwable cause = th.getCause();
+        if (th instanceof RetryableException && cause != null) {
+          throw cause;
+        }
+        throw th;
```
velo: This feels like we are breaking the previous signature for all feign methods and interfaces... before we would throw `FeignException`, now we throw `Exception`.

I think we could introduce a new method into the `Retryer` called `propagateException(FeignException)`, and ship feign with 2 implementations.  

* `Legacy` throws the current `FeignException`
* `Cause` throws the root cause for a `FeignException`

kdavisk6: This is a breaking change, we should weigh that carefully.

Should we consider a mechanism similar to Hystrix Javanica?

https://github.com/Netflix/Hystrix/blob/7f5a0afc23aa5ff82320560a04d4c81a45efd67c/hystrix-contrib/hystrix-javanica/src/main/java/com/netflix/hystrix/contrib/javanica/command/AbstractHystrixCommand.java#L148-L158

There is an assumption made that unless a particular option is set, the **original** exception is thrown.  If the original exception cannot be thrown, it is wrapped and thrown.  In Feign, we **always** wrap the exception.  I think this is the core issue @dinomite and I have been discussing in #737.  Maybe we can add a top level option or annotation to `Target` methods that allow for finer control over what exceptions get wrapped and what exceptions do not.

kdavisk6: @dinomite what are your thoughts on our suggestions?

dinomite:
@kdavisk6 Sorry, traveling right now—I'll provide a proper reply next week.

core/src/main/java/feign/ReflectiveFeign.java
dinomite: Thanks for reminding me that I had forgotten about this.

You're definitely right that this is big change, so doing it in a backwards-compatible way is the most tenable—I'd be ok with either @velo's suggestion of a pluggable solution or what you have suggested, @kdavisk6.  I'm a bit out of context on Feign recently, do either of these look better to you all?  Should  Does one of them sound preferable for me to give a go at?

kdavisk6: We will leave that up to you to choose.  The goal is to limit our exposure to existing clients. 

I think we can expand on @velo's suggestion to add a new method to the `Builder` called `propagateExceptions` which will change the default `Retryer` registered.  This will allow existing integrations to work and those that want the behavior to enable it without having to add additional code.  Thoughts?

velo: I liked it

dinomite: Got it.  I'll have some updates to this PR in the next week or two.

dinomite:
Updated—I liked the suggestion of adding propagation configuration in the builder, thanks for getting me started.  The change is in `SynchronousMethodHandler`, `catch`ing `RetryableException`s that `continueOrPropagate()` may throw and, if `propagateExceptions` is enabled and there is a cause, the cause is thrown instead of the `RetryableException`.

dinomite:
Updated the README

README.md
```diff
@@ -727,7 +727,9 @@ public class Example {
 `Retryer`s are responsible for determining if a retry should occur by returning either a `true` or
 `false` from the method `continueOrPropagate(RetryableException e);`  A `Retryer` instance will be 
 created for each `Client` execution, allowing you to maintain state bewteen each request if desired.
-If the retry is determined to be unsucessful, the last `RetryException` will be thrown.
+
+If the retry is determined to be unsuccessful, the last `RetryException` will be thrown.  To throw the original
+cause that led to the unsuccessful retry, build your Feign client with the `propagateExceptions()` option.
```
velo: 👍

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -43,13 +43,14 @@
   private final ErrorDecoder errorDecoder;
   private final boolean decode404;
   private final boolean closeAfterDecode;
+  private final boolean propagateExceptions;
```
velo: I like the PR and the general idea....

The only think I would ask is to replace this `boolean propagateExceptions` with an enum `ExceptionPropagationPolicy propagationPolicy`

Loved the `builder. propagateExceptions()` idea

velo:
Nice, liked the new work, removed the `breaking change` label as now is a configuration

core/src/main/java/feign/SynchronousMethodHandler.java
dinomite: Sounds good to me, updates soon.

velo:
Looks good, I will merge once build passes

dinomite:
Hmm, looking into that error…

PR #758, opened 2018-08-08
Allow decoding of parameterizedTypes (generics) Fixes #759

### LINKED ISSUE #759: JAXB Decoder doesn't support generics
Trying to decode instances of generics objects throws an exception because JAXBDecoder can't produce a context. 
The raw type can be extracted and used for creating a JAXBContext for the generic type.

RicardoRdzG:
If type is a parameterizedType then extract rawType before proceeding with decoding instead of Trowing an exception

### DIFF
--- a/jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
+++ b/jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
@@ -15,6 +15,7 @@ package feign.jaxb;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.lang.reflect.ParameterizedType;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Unmarshaller;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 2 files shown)


jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
```diff
@@ -69,6 +70,10 @@ public Object decode(Response response, Type type) throws IOException {
       return Util.emptyValueOf(type);
     if (response.body() == null)
       return null;
+    while (type instanceof ParameterizedType) {
+      ParameterizedType ptype = (ParameterizedType) type;
```
velo: Would be nice to have a test covering this loop

RicardoRdzG:
@Velo Added a test and reformatted using eclipse formatter

jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
RicardoRdzG: test added

velo: Sweet, thanks! I will review it right now

jaxb/src/test/java/feign/jaxb/JAXBCodecTest.java
```diff
@@ -122,11 +123,14 @@ public void encodesXmlWithCustomJAXBNoNamespaceSchemaLocation() throws Exception
...
-    assertThat(template).hasBody("<?xml version=\"1.0\" encoding=\"UTF-8\" " +
-        "standalone=\"yes\"?><mockObject xsi:noNamespaceSchemaLocation=\"http://apihost/schema.xsd\" "
-        +
-        "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">" +
-        "<value>Test</value></mockObject>");
+    assertThat(template)
+        .hasBody(
+            "<?xml version=\"1.0\" encoding=\"UTF-8\" "
+                +
+                "standalone=\"yes\"?><mockObject xsi:noNamespaceSchemaLocation=\"http://apihost/schema.xsd\" "
```
velo: could you add the `+` sign in front of the string?

```
                + "standalone=\"yes\"?><mockObject xsi:noNamespaceSchemaLocation=\"http://apihost/schema.xsd\" "
                + "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">"
                + "<value>Test</value></mockObject>"
```

RicardoRdzG: I used the eclipse formatter on the file before uploading. Can that change be added to the formatting file?

kdavisk6:
@velo something is off with the Travis build for this PR.  It succeeded for JDK8 but not for JDK11.  Can you take a look?

RicardoRdzG:
build failure seems like a fluke, sent a small amend to kick the build

PR #764, opened 2018-08-10
support charset when build Reader for Response.Body

arebya:


kdavisk6:
Please add a description to this PR.  Also, please indicate which issue this closes in the description.

core/src/main/java/feign/Response.java
```diff
@@ -327,6 +332,11 @@ public Reader asReader() throws IOException {
       return new InputStreamReader(inputStream, UTF_8);
     }
 
+    @Override
+    public Reader asReader(Charset charset) throws IOException {
+      return new InputStreamReader(inputStream, charset == null ? UTF_8 : charset);
```
velo: Let's make `charset` mandatory when calling this method.

Use `Util.checkNotNull(charset, "'charset' must be not null")`

arebya: I will create a new pr

Issue #757, opened 2018-08-07
Response body charset 


arebya:
How can I pass the defined charset such as GBK to Response body when create InputStreamBody.
![image](https://user-images.githubusercontent.com/1525048/43760952-2ac8d8c2-9a56-11e8-904c-6285fb589330.png)

velo:
You are correct, we need another `asReader(Charset encoding)`

As work around, you can use `asInputStream()` to create your own.

If you create a pull request for this, I can include on `10.0` release

arebya:
I have created a pull request at [#764](https://github.com/OpenFeign/feign/pull/764)

kdavisk6:
Closed via #766

PR #766, opened 2018-08-13
support charset when build Reader for Response.Body

arebya:
As mentioned in issue[#757](https://github.com/OpenFeign/feign/issues/757)

We need a asReader(Charset charset) method in Response.Body

### DIFF
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -13,6 +13,11 @@
  */
 package feign;
 
+import static feign.Util.UTF_8;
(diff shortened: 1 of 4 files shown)


core/src/main/java/feign/Response.java
```diff
@@ -37,298 +37,315 @@
  */
 public final class Response implements Closeable {
 
-  private final int status;
```
velo: could you revert this format changes?

runninng `mvn clean install` should fix code format

arebya: done

PR #769, opened 2018-08-14
Added FeignRequestException, FeignResponseException 

### LINKED ISSUE #642: FeignException with content as separate field
FeignException includes returned content in message. While this is mostly OK (unless the returned content is very long), the exception could also provide a separate field to obtain the content itself for further parsing instead of having to parse the message for it.

finke-ba:
Idea was to use two different classes for holding Request and Response objects.

Couple of moments that I'm not sure about:
1. Add request object to RetryableException and to errorExecuting method. 
2. In FeignResponseException I'm copying response body to another input stream due to root input stream already closed on the moment of handling exception.
3. I deleted the response body from the message variable what makes this commit back incompatible.
4. Test added put in correct place/class.

I'm looking forward to questions and comments.

Fixes #642


kdavisk6:
The only difference between `FeignRequestException` and `FeignResponseException` is the inclusion of a `body` attribute.  I would reconsider this approach.  If the intent is to make the body available, we can add a `body` attribute to `FeignException` and get the same result, without the cascading changes.

Can you give that a try?

finke-ba:
@kdavisk6 Thanks for comment!
The main idea to create two additional exceptions was to have Request and Response object in Exception class. It gives more flexibility instead of adding just a body. Maybe it's already too much.
But please check carefully comments for this PR #664 first, because I used the ideas from that.

kdavisk6:
The notions of `Request` and `Response` are abstractions that are within OpenFeign.  User's of the library are defining interfaces, with methods, not Requests and Responses.  By exposing those abstractions through `FeignException`, we are **leaking** those out to users, requiring them to at least be aware of those abstractions and how to work with them.  This is the definition of [Leaky Abstractions](https://en.wikipedia.org/wiki/Leaky_abstraction), and it's something we should strive to avoid.

The original issue (#642), indicated that a separate field, with the appropriate body, would be sufficient, and I agree.  Another suggestion that might clear this up is to not call the field `body`, but instead call it `details` or something else.  `body` or `RequestBody` is another abstraction we should avoid **leaking**. 

User's can then inspect the Exception details for additional information.

finke-ba:
Ok, I'll create another implementation with one additional attribute in `FeignException `.

finke-ba:
That is awsome.
@kdavisk6 thank you alot for your help!

PR #783, opened 2018-09-10
httx.subSequence(0, 3).equals(http.substring(0, 3)), but httx is not …

suzhenyuan:
httx.subSequence(0, 3).equals(http.substring(0, 3)), but httx is not a valid http url format.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -95,7 +95,7 @@ public final class RequestTemplate implements Serializable {
   }
 
   private static boolean isHttpUrl(CharSequence value) {
-    return value.length() >= 4 && value.subSequence(0, 3).equals("http".substring(0, 3));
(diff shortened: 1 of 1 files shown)


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -95,7 +95,7 @@ static String urlEncode(Object arg) {
   }
 
   private static boolean isHttpUrl(CharSequence value) {
-    return value.length() >= 4 && value.subSequence(0, 3).equals("http".substring(0, 3));
+    return value.length() >= 4 && value.subSequence(0, 4).equals("http".substring(0, 4));
```
velo: change make sense,

Just wonder if you have a use case, so we could attach a test

suzhenyuan: It is my first time to pull a request. I will submit a test case next time.

kdavisk6:
This is no longer necessary per #778

PR #784, opened 2018-09-10
changed default query encoder result from POJO field to getter property

nonpool:
changed to getter property can use super class getter property, and also we are used to serialization and deserialization using the java bean specification.


kdavisk6:
This looks very good to me overall.  The use of `java.beans` has historically been discouraged, but this is a good use case for use to revisit that.  Thoughts @velo?

core/src/test/java/feign/CustomPojo.java
```diff
@@ -22,4 +22,12 @@
     this.name = name;
     this.number = number;
   }
+
+  public String getName() {
```
velo: @kdavisk6 this seems a breaking change to me.

Before `feign` would read fields... with this change, you need to have getters/setters

velo:
I think we should leave the current `QueryMapEncoder` as is (may be mark as deprecated, but not changed)

and offer a new `POJOQueryMapEncoder` to avoid a breaking change.

So, what I would like to see:
1 - new `POJOQueryMapEncoder` as presented on this PR
2 - rename `QueryMapEncoder.Default` to `FieldQueryMapEncoder`
3 - create a new `QueryMapEncoder.Default` that extends `FieldQueryMapEncoder` and mark it as `@Deprecated`

That would give current user a path forward w/o breaking compatibility

nonpool:
I agree with you @velo, this PR don't  consider compatibility.
I am very happy to submit it again as you wish.

nonpool:
about name, i think `PropertyQueryMapEncoder` is good.. what do you think? @kdavisk6

kdavisk6:
@nonpool Since this is your proposal, why don't you decide on the name and update the PR with @velo's suggestions.

nonpool:
update completed @kdavisk6 @velo

velo:
For some reason, your PR is including 73 files and over 3k lines changed.

Could you merge master into your branch?

rage-shadowman:
I always saw the word "properties" (wrt Objects) as referring to member variables or what you called "fields" (I think "field" encoder is a fine name for the original encoder). But maybe it is a more ambiguous term meaning both fields and things returned from getters.

If this new encoder encodes based solely on getter methods, then I wouldn't call it "property" encoder (just to avoid any ambiguity confusion). I'd vote for "bean" encoder unless my definition of "properties" is completely non-standard or contrary to general convention.

I would suggest adding a section on the standard query map encoders in the README.md file. I saw some cleanup in there, but I didn't see any mention of this new encoder.

nonpool:
I should have done a wrong operation and caused the other commits to be merged into this PR. So I should create a new PR and close this PR？ 
About encoder name,  i should follow your advice. `BeanQueryMapEncoder` is also a good choice

velo:
`git merge master` will probably solve this

nonpool:
I have tried this, but both the master branch and this branch are up to date. and `git merge master` already are up to date  :(

kdavisk6:
You will need to rebase instead of merge.

nonpool:
recommit to [PR#802](https://github.com/OpenFeign/feign/pull/802)

PR #788, opened 2018-09-13
Fix NPE in OptionalDecoder

thanus:
The decoded instance (from `delegate.decode()`) in OptionalDecoder will be null when a response body is null. This decoded instance (null) will result in a NPE in the Optional since Optional.of() is used. 
Changing this to Optional.ofNullable() will return an empty Optional when the decoded instance is null.


thanus:
@kdavisk6 Good point, I already tried to test this within the OptionalDecoderTests. Even when the body on MockServer is not set, the StringDecoder returns an empty String which will be then an Optional with an empty String by the OptionalDecoder.

```java
  @Test
  public void test200WhenResponseBodyIsNull() throws IOException, InterruptedException {
    final MockWebServer server = new MockWebServer();
    server.enqueue(new MockResponse().setResponseCode(200));

    final OptionalInterface api = Feign.builder()
        .decoder(new OptionalDecoder(new Decoder.Default()))
        .target(OptionalInterface.class, server.url("/").toString());

    Optional<String> response = api.get();

    assertThat(response.isPresent()).isFalse();
  }
```
Although when for example GsonDecoder is used it works, because it returns null when the response body is null. Anyone suggestions how you can test this?

velo:
Hi @thanus,
you could just create a simple mock `Decoder` that always returns null.

thanus:
@velo thnx, a test has been added using a mock decoder. I have also added tests when Optional is present and when decoding no Optional.

core/src/test/java/feign/optionals/OptionalDecoderTests.java
```diff
@@ -54,6 +57,45 @@ public void simple204OptionalTest() throws IOException, InterruptedException {
...
+    assertThat(response).isEqualTo(Optional.of("foo"));
+  }
+
+  @Test
+  public void test200WhenResponseBodyIsNull() throws IOException, InterruptedException {
+    final MockWebServer server = new MockWebServer();
+    server.enqueue(new MockResponse().setResponseCode(200));
+
+    final OptionalInterface api = Feign.builder()
+        .decoder(new OptionalDecoder(((response, type) -> null)))
```
velo: 🥇

PR #789, opened 2018-09-18
The @Headers does not work

likuankk:
The @Headers does not work when it has blank space around ":".
I'd expect to be able to do something like the following
`@RequestLine("POST /xml")`
`@Headers("Content-Type : text/xml")`
` String sentXml(String x);`
Unfortunately this doesn't seem to work because it has a blank space before ":".  The service receive a HTTP request that contains headers "Content-Type: application/x-www-form-urlencoded" because the HttpURLConnection cannot  recognize “Content-Type ”.

> if (!method.equals("PUT") && (poster != null || streaming())) {
                requests.setIfNotSet ("Content-type",
                        "application/x-www-form-urlencoded");
            }

It does not friendly to developer.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -333,11 +333,11 @@ public interface Contract {
           new LinkedHashMap<String, Collection<String>>(input.length);
       for (String header : input) {
         int colon = header.indexOf(':');
-        String name = header.substring(0, colon);
+        String name = header.substring(0, colon).trim();
         if (!result.containsKey(name)) {
           result.put(name, new ArrayList<String>(1));
         }
-        result.get(name).add(header.substring(colon + 2));
+        result.get(name).add(header.substring(colon + 1).trim());
       }
       return result;
     }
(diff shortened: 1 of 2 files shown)


kdavisk6:
The [Http Message Specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html) states field names must be followed immediately by a colon.  The server you are sending this request to is correct in informing you that `Content-Type ` is an invalid header field name, because it is. 

We cannot accept this change as is, but you do point out that our logic for parsing header values is incomplete.  Also stated in the specification is that field values may be preceded by any amount of leading white space (LWS), and our contract blindly expects only one.  Would you be willing to update this PR to correct the field value whitespace while reverting the field name whitespace changes?

kdavisk6:
Thank you for making the changes.  Can you please add a test case?

likuankk:
After PR , i read the RFC 2616 carefully. It states:

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive. The field value MAY be preceded by any amount of LWS, though a single SP is preferred.

> field-value    = *( field-content | LWS )

> The field-content does not include any leading or trailing LWS: linear white space occurring before the first non-whitespace character of the field-value or after the last non-whitespace character of the field-value. Such leading or trailing LWS MAY be removed without changing the semantics of the field value. 

I'm sorry for not thinking about it. Therefore, i update the code. Also, i add the test case for it.

Issue #791, opened 2018-09-19
SAXParseException: Premature end of file when Status Code is 204


selimok:
Hi There, 

I just came across a problem with feign clients JaxbDecoder. While trying to decode it checks first the response status and if it's 404 then it returns a empty value of target type. This should be also the case for status code 204, since it indicates by design a response with no body. 

The current implementation missed the case and cause a SaxParseException

```
[org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.]] with root cause
org.xml.sax.SAXParseException: Premature end of file.
```

Thanks.

kdavisk6:
I believe the root cause of this is that `SynchronousMethodHandler` assumes that all `2xx` responses should be decoded.  It may be more appropriate the logic there to explicitly skip decoding on `204 No Content`.  Thoughts?

PR #792, opened 2018-09-19
Return null (empty target object) if status code is 204

selimok:
As mention in #791 JAXBDecoder should check also for status code 204 before trying to decode response body.

### DIFF
--- a/jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
+++ b/jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
@@ -66,7 +66,7 @@ public class JAXBDecoder implements Decoder {
 
   @Override
   public Object decode(Response response, Type type) throws IOException {
-    if (response.status() == 404)
+    if (response.status() == 404 || response.status() == 204)
       return Util.emptyValueOf(type);
     if (response.body() == null)
       return null;


jaxb/src/main/java/feign/jaxb/JAXBDecoder.java
```diff
@@ -66,7 +66,7 @@ private JAXBDecoder(Builder builder) {
 
   @Override
   public Object decode(Response response, Type type) throws IOException {
-    if (response.status() == 404)
+    if (response.status() == 404 || response.status() == 204)
```
velo: Sounds fair.

Could you just add a test for `204` as well?

kdavisk6: Or better yet, we should inspect the body before sending it to the `Decoder` here:

https://github.com/OpenFeign/feign/blob/d7cc9b6c7ee34da15ee2cf0e1c0824b3040e5737/core/src/main/java/feign/SynchronousMethodHandler.java#L162-L170

selimok: I prefer a global way to handle it.

velo: > Should we do this globally or just in this decoder?

@kdavisk6 we could, but that would break current API for all decoders...

If you think of a `OptionalDecoder` the result for a 204 or 404 is `Optional.empty()`

For a `JAXB` decoder null
For a `List` decoder may be `Collections.emptyList()` (which I currently use for a binary format decoder.

So, although I would like to see this in a single location, I don't think we can.

kdavisk6: Fair enough.  Let's go with this for now and keep an eye out for other issues.

PR #799, opened 2018-09-28
fix #797 fix #798: JAXRSContract sets a single Content-Type value

### LINKED ISSUE #797: JaxRsContract sends multiple Content-Type values
Interfaces with multiple `@Consumes` annotations cause requests to be created with multiple Content-Type header values, which is invalid.

This regression was intruded with #763 (pr #765).  It makes sense for the `Accept` header, but I've had issues with interfaces annotated with multiple `@Consumes` values. Is there ever a case when we would want to send multiple content types? The request body can only be one type of content.

### LINKED ISSUE #798: RequestTemplate headers can no longer be cleared
Previously invoking `template.header("Foo", (String) null)` would clear existing headers, as expected by JAXRSContract:113, however that is no longer the case.

This causes class-level content-type and accept headers to be applied to all methods, even when those methods are annotated with a different set of values. We expect method annotations to override class level annotations.

This regression was introduced in #778

cakofony:
This change allows headers to be cleared by passing a null
value for backwards compatibility.

Multiple Content-Type values are not valid because the body
that we send with any given request will only have a single
type.

Updated header entry assertion to be agnostic to header name
order.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -634,8 +634,7 @@ public final class RequestTemplate implements Serializable {
    * @see RequestTemplate#header(String, Iterable)
    */
   public RequestTemplate header(String name, String... values) {
-    boolean clearValues = values == null || (values.length == 1 && values[0] == null);
-    return header(name, clearValues ? null : Arrays.asList(values));
+    return header(name, Arrays.asList(values));
   }
 
   /**
@@ -650,8 +649,7 @@ public final class RequestTemplate implements Serializable {
       throw new IllegalArgumentException("name is required.");
     }
     if (values == null) {
-      headers.remove(name);
-      return this;
+      values = Collections.emptyList();
     }
 
     return appendHeader(name, values);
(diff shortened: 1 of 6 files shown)


kdavisk6:
Thanks for working on this.  I think that we can address this more explicitly by adding an additional method to `RequestTemplate`, `header(String value, String... values, boolean replace)`, and update `appendHeader` accordingly.  This make the operation explicit and does not rely on rules such as `null` checking and so on.  The implementation would always create a new `HeaderTemplate` and add it to the map.  I've included `appendHeader` for reference.

https://github.com/OpenFeign/feign/blob/02d97371fe3a84bfd3222e18b94d791c5841170e/core/src/main/java/feign/RequestTemplate.java#L658-L674

kdavisk6:
Our preferred style for chained methods to omit the `set` prefix.   Do you have another suggestion for a method name or signature?

cakofony:
Perhaps rather than replace, a `clearHeader(String)` function? There aren't any other methods of that sort in `RequestTemplate` though.

It's a bit odd that the `header` functions behave differently from `query`, which do clear when an empty collection or null is applied. In its current state, this change provides parity between the two. I have some custom code in my projects which attempts to clear headers by setting null, I would wager that I'm not alone. I'd almost prefer to rename the function with the behavior change, otherwise api consumers will begin to see unexpected behavior after upgrading feign.

kdavisk6:
Looking over `RequestTemplate` again, you are correct.  `appendQuery` does look for an empty `Iterable` and will clear the query parameter.  `appendHeader` does not.  Would you be willing to update your PR to bring `appendQuery` and `appendHeader` in line?  That should restore the previous behavior.

cakofony:
Sure thing, I’ll put it together tomorrow.

PR #800, opened 2018-09-28
Proxied Feign's Client implementation [Tested]

herrhilmi:
Avoid missing with OkHttp dependecies to add proxied implementation.

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -18,7 +18,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.net.Proxy;
 import java.net.URL;
+import java.nio.charset.Charset;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -33,6 +36,7 @@ import static feign.Util.CONTENT_ENCODING;
 import static feign.Util.CONTENT_LENGTH;
 import static feign.Util.ENCODING_DEFLATE;
 import static feign.Util.ENCODING_GZIP;
+
 
 /**
  * Submits HTTP {@link Request requests}. Implementations are expected to be thread-safe.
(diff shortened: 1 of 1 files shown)


kdavisk6:
Thank you for your PR.  Can you please create an issue explaining what problem you are having and what the expected outcome of this change should be.  That will help provide the context we need to review this properly.

herrhilmi:
I've created this issue https://github.com/OpenFeign/feign/issues/801
should i push another PR ?

kdavisk6:
No, that won't be necessary.

velo:
What I would really like to see was an unit test =)

herrhilmi:
Yes, you'are absolutely right. I'll try to add some 👍

kdavisk6:
@herrhilmi If you could add some tests, I think this PR is in good shape to be accepted.

kdavisk6:
Thinking about this more, I suggest that, instead of modifying the existing default client, we provide a a `ProxyClient` implementation, allowing users to choose which targets should use a proxy and which should not.

Any thoughts?

velo:
> Thinking about this more, I suggest that, instead of modifying the existing default client, we provide a a `ProxyClient` implementation, allowing users to choose which targets should use a proxy and which should not.

There is a `Default` and a `Proxied`  implementation. 

Do you think that covers your request? 


The only thing I'm really missing is some testing.

kdavisk6:
@velo 

My concern is that `Default` was changed to accommodate `Proxy` by updating the `convertAndSend` method to accept a `Proxy`.  I would prefer that the `Proxy` Client be separated more.  For example:

```java
final boolean useProxy = proxy != null;
final URL url = new URL(request.url());  
final HttpURLConnection connection =  (HttpURLConnection) (useProxy ? url.openConnection(proxy) : url.openConnection());
```
is in `Default`, creating a dependency on the proxy flag for all cases.  I would prefer this logic be isolated into the `Proxy` client.  To be even more specific, here is what I would like to see here:

* Revert changes to `Default` client
* Isolate `Proxy` usage to the `Proxy` client only
* Unit Tests for the `Proxy` client added.

Issue #803, opened 2018-10-05
Enforce code format on CI


velo:
Once again, `master` source code deviated from formatter files included on build process.

Let's enforce it.

PR #804, opened 2018-10-05
Enforce format

velo:
Fix for #803


velo:
Confirmed on `887c6f4` that build will indeed fail!

Now, fixing broken files

PR #807, opened 2018-10-06
Make jaxb modules build on java 11

velo:
This was one of the last steps missing to get a functional java 11 build.

JAXB is now gone from java as part of JEP 320 http://openjdk.java.net/jeps/320

Added a dependency for java 11 builds

### DIFF
--- a/jackson-jaxb/pom.xml
+++ b/jackson-jaxb/pom.xml
@@ -70,4 +70,30 @@
     </dependency>
   </dependencies>
 
+  <profiles>
+    <profile>
+      <activation>
+        <jdk>11</jdk>
+      </activation>
+
+      <!--
+        JAXB was removed from java SDK on JEP 320
+        http://openjdk.java.net/jeps/320
+       -->
+      <dependencies>
+        <dependency>
+          <groupId>javax.xml.bind</groupId>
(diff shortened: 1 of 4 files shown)


jackson-jaxb/pom.xml
```diff
@@ -70,4 +70,24 @@
...
+
+      <dependencies>
+        <dependency>
+          <!--
+            JAXB was removed from java SDK on JEP 320
+            http://openjdk.java.net/jeps/320
+           -->
+          <groupId>org.glassfish.jaxb</groupId>
+          <artifactId>jaxb-runtime</artifactId>
+          <version>2.4.0-b180830.0438</version>
```
kdavisk6: Should we use this dependency or the general `jaxb` one?

```
<!-- https://mvnrepository.com/artifact/javax.xml.bind/jaxb-api -->
<dependency>
    <groupId>javax.xml.bind</groupId>
    <artifactId>jaxb-api</artifactId>
    <version>2.3.1</version>
</dependency>
```

velo: Done'ish...

still need the runtime to run tests... but dependencies are a bit better now

PR #813, opened 2018-10-14
Set JDK release when building with java 11

velo:


velo:
Nop, not going to work

Issue #774, opened 2018-08-17
java8 features break compatibility with android sdk <24


masc3d:
`feign-10` is now incompatible with sdk 24 or lower (when using eg. `JAXRSContract` which uses java8 lambdas), equalling 50% of android devices out there.

masc3d:
throwing `NoClassDefFoundError` on
```
removeValues(consumes.value(), (mediaType) -> emptyToNull(mediaType) == null, String.class);
```

maybe it's not the lambda though, as those should be supported on all devices by now.

masc3d:
caused by `java.util.function.Predicate` in `feign.Util` which is unsupported on sdk <24.

masc3d:
> We made a conscious decision to move to Java 8 knowing that we would have some issues with Android

was there a public discussion I could read up on @kdavisk6?

well it probably implies that feign >= 10 will not be feasible to use for majority of android projects for what I would estimate to be 5-10 years until < sdk 24 devices become irrelevant.

a pretty harsh decision. I would rather have opted to move feign to kotlin to maintain portability and have better language features at the same time.

is it too late to re-evaluate this?

kdavisk6:
It was discussed in #656.  We knew that this would be something that will come up.  The consideration towards Android was, let's see how it goes.  And now, here we are.    Moving to Kotlin is a more extreme move than moving to Java 8, and is something that we would encourage by creating a new library.  

Knowing that, what do you suggest we do?  From my perspective I see a few options.

- Restrict [Java 8 language usage to that of Android](https://developer.android.com/studio/write/java8-support)
- Require that Android developers use an SDK higher than 24 when using OpenFeign 10.x
- Create a `feign-android` specific version, that deals with this issue specifically.

We need to strike a balance between providing support for as many platforms as we can, while still moving the project forward, adding capabilities that allow us to do so.  Java, which has been stagnant for some time, is now moving very quickly forward.  From my perspective, which is just one of many, this is just the first step toward Java 10 and 11, which will bring another set of challenges.

Thoughts?

masc3d:
> Restrict Java 8 language usage to that of Android

not feasible if you want to extensively leverage language features plus requires verification mechanisms to prevent unsupported java8 features from leaking in.

> Require that Android developers use an SDK higher than 24 when using OpenFeign 10.x

won't happen with > 50% unsupported atm and current rate of android moving forward. I cant imagine any project which can afford to ignore 50% or say even 25% of android devices.

[...]

PR #816, opened 2018-10-18
Compatibility with Android SDK < 24

masc3d:
resolving #774

### DIFF
--- a/core/src/main/java/feign/java8/util/StringJoiner.java
+++ b/core/src/main/java/feign/java8/util/StringJoiner.java
@@ -1,0 +1,261 @@
+/*
+ * Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved. DO NOT ALTER OR REMOVE
+ * COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License version 2 only, as published by the Free Software Foundation. Oracle
+ * designates this particular file as subject to the "Classpath" exception as provided by Oracle in
+ * the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version 2 along with this work;
+ * if not, write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA or visit www.oracle.com
+ * if you need additional information or have any questions.
+ */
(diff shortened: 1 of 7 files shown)


velo:
Sounds fair, but CI is not happy about it

masc3d:
@velo @kdavisk6 as suspected, here comes the rabbit hole (eg `String.join`)

masc3d:
I would repackage predicates, `String.join` and any classes required for compat in the future (and there will be more) from openjdk into `feign.util`. is this ok with you?

masc3d:
can't do this without relaxing license header restrictions though

masc3d:
I also see `feign-10` makes extended use of `java.beans` and `hamcrest` all of which are not compatible with android. opinions on resolutions?

masc3d:
the way I see it we also need safeguards (eg. lint) to prevent potential android breakage on every minor change.

masc3d:
> I agree with @velo. Those implementations are quite simple, so we can add our own.

not applicable. this is a rabbit hole and more and more compat code will be added if android should be properly supported. repackaging is probably the only reasonable option.

velo:
What do you mean by `repackaging`?

masc3d:
> What do you mean by `repackaging`?

(very) selectively import classes from `streamsupport` and possibly `openjdk` (eg. `StringJoiner` including static utility methods) into a dedicated package `feign.java8` which has relaxed license header rules.

velo:
Ok, `streamsupport` is acceptable (personal opinion here) now, `apache-commons` no way jose

masc3d:
> `apache-commons` no way jose
not required when repackaging eg. `StringJoiner` from openjdk

core/src/main/java/feign/java8/util/StringJoiner.java
```diff
@@ -0,0 +1,261 @@
...
+ * This code is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * General Public License version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version 2 along with this work;
+ * if not, write to the Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA or visit www.oracle.com
```
velo: @kdavisk6 thoughts?

kdavisk6: I don't think this is appropriate.  I would prefer that we implement our own version of the `join` function over copying code like this.

Regarding the problem at hand, I'm still not comfortable with creating our own "polyfills" when the JDK supports these natively.

masc3d: >  I would prefer that we implement our own version of the join
@kdavisk6 if it was just about `join` I would agree

but in the end any solution will be kind of awkward. I already detected more incompatibilities.
maybe we should conclude that `feign-10` can only be compatible with SDK >= 24.

kdavisk6:
Based on our conversations, I'm going to close this PR in favor of documenting that, if you need to work on Android versions less than 24, that you use the Feign 9.x releases.

PR #822, opened 2018-10-24
Add support for java 8's Optional type to represent a HTTP 404 response.

czubin:


kdavisk6:
I'm not clear on how this change addresses the issue in the title.  How does adding `Optional` to the empty list replace the need to use the `Optional` decoder?  Is there another test case that can be added to demonstrate the desire behavior?

kdavisk6:
My only request is that we also update the documentation to reflect how to use Optional.

velo:
The usage is exactly the same... It just allow the default decoder to deal with empties returning the type, not null

PR #824, opened 2018-10-24
support PATCH with empty body paramter.

### LINKED ISSUE #665: PATCH does not allow a body that is null
I have been using the OkHttpClient for some time, and have been using a POST endpoint that takes no body. I recently changed this endpoint to use PATCH instead, but now the client throws an IllegalArgumentException, as it requires a PATCH request to have a body. I found the reason for not having this issue with POST in the code below, and I am wondering if there is any good reason why this does not also apply to PATCH? Also, the url in the code does not work.

feign/okhttp/src/main/java/feign/okhttp/OkHttpClient.java, line 77:
```
    byte[] inputBody = input.body();
    boolean isMethodWithBody = "POST".equals(input.method()) || "PUT".equals(input.method());
    if (isMethodWithBody) {
      requestBuilder.removeHeader("Content-Type");
      if (inputBody == null) {
        // write an empty BODY to conform with okhttp 2.4.0+
        // http://johnfeng.github.io/blog/2015/06/30/okhttp-updates-post-wouldnt-be-allowed-to-have-null-body/
        inputBody = new byte[0];
      }
    }
```

alzipursky:
Fixes #665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Version 10.1
+* Supports PATCH without a body paramter
+
 ### Version 10.0
 * Feign baseline is now JDK 8
   - Feign is now being built and tested with OpenJDK 11 as well. Releases and code base will use JDK 8, we are just testing compatibility with JDK 11.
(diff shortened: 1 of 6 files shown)


kdavisk6:
The idea of a `PATCH` request without a body is still awkward to me.  But, if folks want it, I'm cool with it.

matevskial:
@kdavisk6 Imagine Postman, for example, not allowing you to send a patch request without body because... it is awkward :)

PR #825, opened 2018-10-26
Add fine-grained HTTP error exceptions

jerzykrlk:
* a hierarchy of HTTP-related exceptions under FeignException.


jerzykrlk:
I'm sorry about that. Fixed.

core/src/main/java/feign/FeignException.java
```diff
@@ -85,4 +124,100 @@ static FeignException errorExecuting(Request request, IOException cause) {
         cause,
         null);
   }
+
+  public static class BadRequest extends FeignException {
```
rage-shadowman: I feel like all the 4xx ones should extend some sort of 4xx exception (say `FiengClientException`) and the same with 5xx errors (maybe `FeignServerException`).

jerzykrlk:
Yes, that's how Spring's exception hierarchy looks like.

I have personally neved had to catch Client/Server exceptions with Spring's RestTemplate. I rather catch one-two specific exceptions or all of them. So I skipped them here. 

That's my use case, though. If you see the point and would like them here, let me know - I'll try to add another PR with them, early next week.

rage-shadowman:
The only point I see is that 4xx means some form of bad request so don't try this again because it will never work, while 5xx means something failed on the server so if you try again later (say after some dependent service comes back online) it might work.

jerzykrlk:
Sure, here's another pull request: https://github.com/OpenFeign/feign/pull/854

PR #832, opened 2018-11-03
Fixes an issue with http-headers duplication when using RequestTemplate

### LINKED ISSUE #570: RequestTemplate does not conform to RFC7230 - case-insensitive header field names
Hi,
Feign RequestTemplate's header field names are not case-insensitive.

Here is the RFC fragment in question:
https://tools.ietf.org/html/rfc7230#section-3.2

Here is the code fragment from `RequestTemplate`:

```java
 private final Map<String, Collection<String>> headers =
      new LinkedHashMap<String, Collection<String>>();
```

And a JUnit test case:

```java
@Test
public void headerFieldNamesShouldBeCaseInsensitive() throws Exception {
    RequestTemplate request = new RequestTemplate();
    String value = "value";
    request.header("TEST", value);
    assertThat(request.headers().get("test"), contains(value)); // Actual   :null
}
```

tworogue:
Hello,

Here're my changes with intention to complete the issue: https://github.com/OpenFeign/feign/issues/570

Also, it's my first open-source PR, so please be kind to describe my faults, in case they are. 

Fixes #570


kdavisk6:
Travis is unhappy with the formatting on the files, please run `mvn clean install` and push the formatting changes.

core/src/test/java/feign/LoggerTest.java
```diff
@@ -76,16 +76,16 @@ public LogLevelEmitsTest(Level logLevel, List<String> expectedMessages) {
               "\\[SendsStuff#login\\] <--- HTTP/1.1 200 OK \\([0-9]+ms\\)")},
           {Level.HEADERS, Arrays.asList(
               "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP/1.1",
-              "\\[SendsStuff#login\\] Content-Type: application/json",
               "\\[SendsStuff#login\\] Content-Length: 80",
+              "\\[SendsStuff#login\\] Content-Type: application/json",
```
kdavisk6: Changes here are not related to this pull request.  Please refrain from submitting changes that are purely formatting.

tworogue: Hello, @kdavisk6 @velo 

Firstly, thanks for the fast review.

Actually, I suppose these changes **do** relate on current commit: it's not just a formatting, but the point is the changing of header fields order to alphabetical.

I reverted only the **LoggerTest** file, ran "mvn clean install" once again locally and got following output:

https://gist.github.com/tworogue/3b9262457101f8753de01db3d9d4ba84

So it seems to me that these changes are mandatory in this particular case, they're not cosmetic and they really make difference.

What would you advice in regarding of fixing this (with no changing LoggerTest file)?

velo: Make sense, since your you are shorting the header keys, this file must be changed.

Thanks for the detailed explanation

velo:
@kdavisk6 I think we are good to go, any other comments?

Issue #817, opened 2018-10-18
Maven Bill Of Materials (BOM)


fschmager:
Consider providing a [BOM project](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Importing_Dependencies).

Ensuring all used feign dependencies are of the same version can be a pain because each project needs to be included separately within `<dependencyManagement>`. This becomes especially obvious when using things like Spring Cloud, because it's opinionated about what version of feign it likes.

Here are two examples of what it would look like, each being a module of the parent pom:
* [Spring Cloud dependencies bom](https://github.com/spring-cloud/spring-cloud-release/blob/master/spring-cloud-dependencies/pom.xml)
* [Brave bom](https://github.com/openzipkin/brave/blob/master/brave-bom/pom.xml)

The spring one looks cleaner to me, since it just declares the projects provided and everything else is just inherited from parent.

velo:
I don't think that this is necessary for feign.

First reason being: feign is spread across 10+ modules, each with their own sub-set of dependencies.
Seconds feign is loosely dependent on the version. We publish feign-jackson using jackson `2.6.4`... but I replace it with `2.9.7`, just fine.
I was honestly considering going the other way around, setting feign dependencies as provided. So, you would ensure to bring them in.

I just didn't, cause, that would made the initial setup harder.


I think, it would make a sense a `bom` list all `feign-*` modules.  That way, you get a single dependency to change to perform a full feign upgrade.

fschmager:
Ah, I see what you mean. Creating a BOM for all of feign's dependencies would most likely be overkill. What I was specifically after was what you alluded to in you last paragraph: a BOM containing feign's own modules. Would be nice to only have to add that one module in `<dependencyManagement>` and be assured _that_ is the version _all_ feign modules will have even if they get pulled in transitively from some other dependency.

velo:
Ok, sound like a plan.... create a BOM that contain all feign modules

PR #833, opened 2018-11-04
Add maven bom

eddumelendez:
See gh-817

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,119 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2012-2018 The Feign Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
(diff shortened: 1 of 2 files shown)


eddumelendez:
Hi @velo according to the list of modules in the project there is no missing modules in the bom except for `example-github`, `example-wikipedia`, `benchmark`

    <module>core</module>
    <module>gson</module>
    <module>httpclient</module>
    <module>hystrix</module>
    <module>jackson</module>
    <module>jackson-jaxb</module>
    <module>jaxb</module>
    <module>jaxrs</module>
    <module>jaxrs2</module>
    <module>okhttp</module>
    <module>ribbon</module>
    <module>sax</module>
    <module>slf4j</module>
    <module>reactive</module>
    <module>example-github</module>
    <module>example-wikipedia</module>
    
    <module>java8</module>
    <module>mock</module>
    <module>benchmark</module>

Can you point to the modules I am missing?

velo:
`java11` is missing.
There is a new module under review that will become a miss if we don't have a process in place to catch it.

Any suggestions?

eddumelendez:
I guess `java11` will be released always with the rest of the modules right? if that the case I can add the module to the `bom`.

> There is a new module under review that will become a miss if we don't have a process in place to catch it.

I think we should track it into the issues section. I do not know to avoid missing a module into the bom.

velo:
I think a simple junit test (on `core`) that reads both pom files and check if they match, would be good enough.

What do you think?

eddumelendez:
parent bom could have more modules than dependencies declared into the bom file so they will always match at least we will declare in the test all the modules that should be present. I am not sure about that approach.

velo:
Well, we can have a exclusion list.... for instance for `example-github`, `example-wikipedia`, `benchmark`.... my main concern is new modules being created, specially by contributors and not making into the bom file.

PR #835, opened 2018-11-05
Fixes RecordedRequestAssert.hasNoHeaderNamed 

### LINKED ISSUE #679: RecordedRequestAssert.hasNoHeaderNamed(String... names) is broken
The `RecordedRequestAssert.hasNoHeaderNamed` method always returns true even if the header exists.

Also, fixing this method breaks the `AbstractClientTest.testPatch` test in the `ApacheHttpClientTest` in the "httpclient" module, since (apparently) the `ApacheHttpClient` adds a "Content-Type" header.

tworogue:
Now it returns 'false' in corresponding cases.

Fixes #679


rage-shadowman:
It seems silly to write unit tests for your test classes. There's probably a lot of helper code in src/test directories that isn't directly unit tested.

However, since it was known broken, a sanity check test couldn't hurt.

velo:
Exactly my feeling... =)

But, I had to ask

PR #844, opened 2018-11-19
Add unit tests for class feign.Util

petersear:
Hi,
I've analysed your codebase and added some unit tests for the class feign.Util.
I've written the tests for these functions with the help of [Diffblue](https://www.diffblue.com/) [Cover](https://static1.squarespace.com/static/5bc9dd6fb2cf795fa3511f70/t/5be2e84f8985836d62e8eb13/1541597264010/Diffblue+Cover+Datasheet.pdf).
These tests should help detect regressions caused by future code changes.

### DIFF
--- a/core/src/test/java/feign/UtilTest.java
+++ b/core/src/test/java/feign/UtilTest.java
@@ -13,6 +13,9 @@
  */
 package feign;
 
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.rules.ExpectedException;
 import org.junit.Test;
 import java.io.Reader;
 import java.lang.reflect.Type;
@@ -30,6 +33,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 
 public class UtilTest {
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
 
   @Test
   public void removesEmptyStrings() {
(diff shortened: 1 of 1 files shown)


core/src/test/java/feign/UtilTest.java
```diff
@@ -122,6 +128,114 @@ public void unboundWildcardIsObject() throws Exception {
...
+    final String errorMessageTemplate = "";
+    final Object[] errorMessageArgs = null;
+    // Act
+    thrown.expect(IllegalArgumentException.class);
+    Util.checkArgument(expression, errorMessageTemplate, errorMessageArgs);
+    // Method is not expected to return due to exception thrown
+  }
+
+  @Test
+  public void checkNotNullInputNullNotNullNullOutputNullPointerException() {
```
rage-shadowman: For future reference, these would make a lot more sense if there was more separation of terms. All of this "NullNotNullNull" stuff makes zero sense to new eyes on the project, but "checkNotNull_referenceNull_messageNotNull_messageArgsNull_nullPointerExceptionThrown" is a lot easier to read and understand the coder's intentions without looking at the code in the test (then when looking at the code, you can more easily tell if it really tests what was intended).

petersear: @rage-shadowman thanks for your feedback. Much appreciated.

PR #856, opened 2018-12-12
Exposes Charset on Request.Body for inspection

### LINKED ISSUE #847: Request.charset() is deprecated, but Request.Body does not provide that information
The method `Request.charset()` is marked `@deprecated` in v10.1.0, and the comment says we should use `Request.Body` instead. But `Request.Body` does not provide any charset information.

kdavisk6:
Fixes #847

When `Request.Body` was introduced, `Request.charset()` was deprecated
in favor of `Request.Body`, but `Request.Body` did not expose the
`Charset` publically.  This change adds a new `charset()` function to
`Request.Body` correcting this oversight.

### DIFF
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -82,6 +82,10 @@ public final class Request {
       return new Request.Body(null, null, null);
     }
 
+    public Charset charset() {
+      return this.encoding;
+    }
+
   }
 
   public enum HttpMethod {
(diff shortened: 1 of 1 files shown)


core/src/main/java/feign/Request.java
```diff
@@ -82,6 +82,10 @@ public static Body empty() {
       return new Request.Body(null, null, null);
     }
 
+    public Charset charset() {
```
velo: Hrmmm, why?!

if `encoding` is null usually mean this is a binary protocol

kdavisk6: Thanks for flagging that.  I didn't consider that use case.  I'll make the changes.

kdavisk6:
@velo I've ran into a number of other issues related to #857 and this PR.  I'm going to close this PR in favor of another.

PR #858, opened 2018-12-13
Refactored RequestTemplate Body Handling

### LINKED ISSUE #857: Remove use of deprecated internal APIs from sub modules
When refactoring `Request` and `RequestTemplate`, only the core module was updated to remove their usage.  These deprecated apis should be removed from the remaining maintained modules.

- [x] core
- [x] googlehttpclient
- [x] gson
- [x] hc5
- [x] httpclient
- [x] hystrix
- [x] jackson
- [x] jaxb
- [x] jaxrs
- [x] mock
- [x] okhttp
- [x] reactive
- [x] ribbon
- [x] sax
- [x] soap
- [x] spring

kdavisk6:
Fixes #857

Changes introduced in #821 deprecated a number of methods on `RequestTemplate`
whose usage throughout the rest of the library was quite prevalent.
While attempting to refactor that usage, a number of additional issues
related to #821 came to light.  The most important one is that the charset
at the `RequestTemplate` level applied only to the resulting `Request` and
did not need to be applied to the `Template` based classes.

It was this mistake, propagating the `RequestTemplate#charset` to the
`Template` objects, was the root cause of the NPE addressed in #821.

This change reverts #821 in favor of a simpler approach that addresses
the original issue and maintains the `Request.Body` abstraction, while
removing the `@deprecation` notices.

Revert "NPE when resolving a template with binary body (#821)"
This reverts commit 4da6d5cad674e4eeae5b56035ac1f94992499449.

### DIFF
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -13,6 +13,7 @@
  */
 package feign;
 
+import feign.querymap.FieldQueryMapEncoder;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
@@ -103,7 +104,7 @@ public abstract class Feign {
     private Logger logger = new NoOpLogger();
     private Encoder encoder = new Encoder.Default();
     private Decoder decoder = new Decoder.Default();
-    private QueryMapEncoder queryMapEncoder = new QueryMapEncoder.Default();
+    private QueryMapEncoder queryMapEncoder = new FieldQueryMapEncoder();
     private ErrorDecoder errorDecoder = new ErrorDecoder.Default();
     private Options options = new Options();
     private InvocationHandlerFactory invocationHandlerFactory =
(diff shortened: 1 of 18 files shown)


kdavisk6:
Some more context.  Attempting to remove all deprecated references from the remaining libraries exposed a few shortcomings with the initial `Request.Body` approach.  First, it muddled the boundaries between the `RequestTemplate` and the `Request`, which only became clear when the test cases started relying on the `RequestTemplate#charset` method.  Second, a lot of the test cases were evaluating the results by interrogating the `RequestTemplate#body` byte array and changing that to `requestBody().asBytes()` leaked the `Request.Body` abstraction. 

#821 linked the `Request#charset` to the `Body#charset`, which could be null, in the case of binary data, where previously, it was never null.  Refactoring the test cases became a slippery slope, causing changes to leak back into `Request`, `RequestTemplate`, and `Request.Body`.  I felt it was better to roll back #821 and re-apply the abstraction is a much smaller and targeted way, by explicitly stating that the charset on the `RequestTemplate` will be used as the charset in the `Request.Body` and that the URI related encoding be forced to UTF-8, per [RFC 3896 - Uniform Resource Identifier](https://tools.ietf.org/html/rfc3986)

They binary body test cases remain and pass.

kdavisk6:
Spring has just accepted 10.1.0 so I can't fix it this way, I'll need to come up with another solution that doesn't break the new changes.

velo:
Ok, I was concern that this change would break binary protocol again.

PR #864, opened 2018-12-20
Fix typo in check null message => HttpClient.java

cchacin:


java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -44,7 +44,7 @@ public Http2Client() {
   }
 
   public Http2Client(HttpClient client) {
-    this.client = Util.checkNotNull(client, "http cliet must be not unll");
+    this.client = Util.checkNotNull(client, "HttpClient must be not null");
```
rage-shadowman: FYI, proper English grammar is "must not be null" or "must be non null" (I'm not sure if grammar requires a hyphen in "non-null").

cchacin: Rephrased 50d209e

PR #871, opened 2018-12-31
Allow 1xx status codes

### LINKED ISSUE #836: IllegalStateException on 1xx response codes
Looks like there is no support for 1xx responses - Illegal state exception is thrown by Feign on any  response with code less than 200 during response building:
https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Response.java#L47

maly7:
Previously [Response](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Response.java#L47) didn't allow status codes in the 1xx series, only 200 and beyond were allowed. This fixes that by allowing status codes in the range from 100 <= status < 600.

Fixes #836

### DIFF
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -45,7 +45,7 @@ public final class Response implements Closeable {
   private final Request request;
 
   private Response(Builder builder) {
-    checkState(builder.status >= 100, "Invalid status code: %s", builder.status);
+    checkState(builder.status >= 100 && builder.status < 600, "Invalid status code: %s", builder.status);
     checkState(builder.request != null, "original request is required");
     this.status = builder.status;
     this.request = builder.request;
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -45,7 +45,8 @@ public final class Response implements Closeable {
   private final Request request;
 
   private Response(Builder builder) {
-    checkState(builder.status >= 100 && builder.status < 600, "Invalid status code: %s", builder.status);
+    checkState(builder.status >= 100 && builder.status < 600, "Invalid status code: %s",
+        builder.status);
     checkState(builder.request != null, "original request is required");
     this.status = builder.status;
     this.request = builder.request;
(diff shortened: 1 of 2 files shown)


kdavisk6:
Travis is unhappy due to formatting. Can you run Maven package locally?  The build will fix any formatting for you. Once complete, please push a new commit.

maly7:
Sorry about that! It should hopefully be fixed now

kdavisk6:
@maly7 our typical process is for one of the maintainers to review, and if approved, we give the others a few days to weigh in.  After that, if there are no additional comments, we'll merge your PR.  Thanks for your help!

core/src/main/java/feign/Response.java
```diff
@@ -45,7 +45,8 @@
   private final Request request;
 
   private Response(Builder builder) {
-    checkState(builder.status >= 200, "Invalid status code: %s", builder.status);
+    checkState(builder.status >= 100 && builder.status < 600, "Invalid status code: %s",
```
rage-shadowman: Personally, I don't see any reason to limit the response code at all. If somebody has a custom server that returns `65536` (or even `-654321`) then why not return the status code that was actually returned?

See [this stackoverflow post](https://stackoverflow.com/a/43094430/2167531) and [this random web discussion](https://community.softwaregrp.com/t5/Application-Perf-Mgmt-BAC-BSM/Possible-negative-status-codes-in-SiteScope/m-p/265881/highlight/true#M9346).

rage-shadowman: That said, I have no objections to this PR.

rage-shadowman: See also [this blog post](https://pokeinthe.io/2017/09/14/http-status-code-handling/).

maly7: @rage-shadowman Thanks for sharing the blog post! I'd be more than happy to make the change so that it doesn't limit the status code. I'll admit it feels a little awkward to be enforcing it here, and not allowing users to have a choice in the matter.

kdavisk6: @maly7 Are you considering making additional updates to the PR to address @rage-shadowman's feedback?

maly7: Yes, I'll have those changes pushed up tonight

PR #877, opened 2019-01-16
Adjusts copyright headers for this year

velo:


kdavisk6:
Good from me when you get Travis happy

PR #888, opened 2019-01-31
Fix ContentService example

okutane:
Currently ContentService example will trigger index out of bounds exception in feign.Contract.Default#toMap due to missing colon separator.


kdavisk6:
@okutane Thank you for the update!

PR #914, opened 2019-02-25
Fixes NullPointerException when accessing a FeignException's content

### LINKED ISSUE #912: NullPointerException if content in FeignException is null
I tried to parse and react on a FeignException and noticed that accessing `FeignException::contentUTF8` throws a NullPointerException if `content` is null:
https://github.com/OpenFeign/feign/blob/db44bfe7077705e6249565893f96470574815c38/core/src/main/java/feign/FeignException.java#L60

The value was null because there was an underlying connection issue (`status` was 0).

Could this be rewritten to return 'null' if the value of `content` is also null? 

Thank you very much!

robachmann:
If the content of a FeignException is null, `contentUTF8()` now returns `null` rather than throwing a NullPointerException. `null` is chosen to allow possible future releases to return an empty optional rather than a blank String. Fixes #912

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -60,7 +60,7 @@ public class FeignException extends RuntimeException {
     if (content != null) {
       return new String(content, UTF_8);
     } else {
-      return null;
+      return "";
     }
   }
 
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -57,7 +57,11 @@ public class FeignException extends RuntimeException {
   }
 
   public String contentUTF8() {
-    return new String(content, UTF_8);
+    if (content != null) {
+      return new String(content, UTF_8);
+    } else {
+      return null;
+    }
   }
 
   static FeignException errorReading(Request request, Response response, IOException cause) {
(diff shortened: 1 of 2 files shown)


robachmann:
@kdavisk6 I managed to implement a fix for the NPE and added a test for it. I hope I did everything correctly and am looking forward to your feedback.

core/src/main/java/feign/FeignException.java
```diff
@@ -57,7 +57,11 @@ public int status() {
   }
 
   public String contentUTF8() {
-    return new String(content, UTF_8);
+    if (content != null) {
+      return new String(content, UTF_8);
+    } else {
+      return null;
```
kdavisk6: It may be better to return an empty string here instead of `null`.  Unless others look into this code, we may be pushing the null pointer higher.  Thoughts?

robachmann: Good point. I thought one might want to differentiate whether the content was empty (`""`) or `null`. Eventually, the method could even return an `Optional<String>` with or without the content. I see your point though, and assume it's more convenient to simply return an empty String :)

robachmann:
@kdavisk6 I've just implemented the change to return an empty String rather than null.

kdavisk6:
@robachmann Thank you for updating the PR and congrats on your first contribution.

PR #930, opened 2019-03-27
Replaced comma with Constant Delimiter in Template

### LINKED ISSUE #924: Query param with comma produce multiple query params
Hi,

The behavior for query params containing a comma has changed in 10.2.0
Consider the following example:
```
@FeignClient(name = "postman", url = "https://postman-echo.com")
public interface PostmanEchoClient {
    @GetMapping("/get")
    void echo(@RequestParam String param);
}

@SpringBootTest
@RunWith(SpringRunner.class)
public class FeignCommaQueryParameterIssueTest {

    @Autowired
    PostmanEchoClient postmanEchoClient;

    @Test
    public void test() {
        postmanEchoClient.echo("first,second");
    }
}
```
10.1.0 produces the following request (what I think is correct): `https://postman-echo.com/get?param=first%2Csecond`
10.2.0 produces: `https://postman-echo.com/get?param=first&param=second`

I've prepared an example reproducing the issue [here](https://github.com/StasKolodyuk/feign-comma-issue). See `master` and `feign_10.1.0` brances to point the difference

Thanks

kdavisk6:
Fixes #924

Commas were used to identify iterable content, which conflicted when
a comma delimited literal was provided during expansion.  This change
switches commas for semi-colons, which are considered reserved secondary
delimiters in RFC 6750 and should not be used without being pct-encoded.

Should be a safer choice.

### DIFF
--- a/core/src/main/java/feign/template/Template.java
+++ b/core/src/main/java/feign/template/Template.java
@@ -32,6 +32,13 @@ import java.util.stream.Collectors;
  */
 public class Template {
 
+  /*
+   * special delimiter for collection based expansion, in an attempt to avoid accidental splitting
+   * for resolved values. semi-colon was chosen because it is a reserved character that must be
+   * pct-encoded and should not appear unencoded.
+   */
+  static final String COLLECTION_DELIMITER = ";";
+
   private static final Logger logger = Logger.getLogger(Template.class.getName());
   private static final Pattern QUERY_STRING_PATTERN = Pattern.compile("(?<!\\{)(\\?)");
   private final String template;
--- a/core/src/test/java/feign/template/QueryTemplateTest.java
+++ b/core/src/test/java/feign/template/QueryTemplateTest.java
@@ -19,6 +19,7 @@ import feign.CollectionFormat;
 import feign.Util;
 import java.util.Arrays;
 import java.util.Collections;
+import javax.management.Query;
 import org.junit.Test;
 
 public class QueryTemplateTest {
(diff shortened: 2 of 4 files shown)


core/src/test/java/feign/template/QueryTemplateTest.java
```diff
@@ -19,6 +19,7 @@
 import feign.Util;
 import java.util.Arrays;
 import java.util.Collections;
+import javax.management.Query;
```
velo: I don't think you really meant to include this =)

core/src/main/java/feign/template/Template.java
```diff
@@ -32,6 +32,13 @@
  */
 public class Template {
 
+  /*
+   * special delimiter for collection based expansion, in an attempt to avoid accidental splitting
+   * for resolved values. semi-colon was chosen because it is a reserved character that must be
+   * pct-encoded and should not appear unencoded.
+   */
+  static final String COLLECTION_DELIMITER = ";";
```
velo: Should we allow this to be configurable? I don't have a opinion on this.

core/src/test/java/feign/template/QueryTemplateTest.java
kdavisk6: Right.  I'll remove it.

core/src/main/java/feign/template/Template.java
kdavisk6: This is for our internal use.  User's can configure the output through the `CollectionFormat` enum already.  I chose semi-colon as to not interfere with those settings.  That turned out to be the cause of the bug, since comma was already taken.

tpatrizio:
> Fixes #924
> 
> Commas were used to identify iterable content, which conflicted when
> a comma delimited literal was provided during expansion. This change
> switches commas for semi-colons, which are considered reserved secondary
> delimiters in RFC 6750 and should not be used without being pct-encoded.
> 
> Should be a safer choice.

I need to preserve semicolon in the parameter value, is this possible with the SpringMvcContract ?

filters=(order=ASC;attributeFilter=false;key=modified;start=2019/12/0612:04:00;)

and it gets converted to:

filters=(order=ASC&filters=attributeFilter=false&filters=key=modified&filters=start=2019%2F12%2F0603:32:00&filters=)

kdavisk6:
Please see #1123 for a temporary workaround.

PR #931, opened 2019-03-27
Removed decoding from Body Template Expansion

### LINKED ISSUE #916: Body contains % ,  as JSON, UriUtils.decode error
Request Body
`{
      "percent" : "100%",
      "name" : "lee"
}`

Feign client  declarative invocation will decode throw exception
 URLDecoder.decode() code `throw new IllegalArgumentException(
                                   "URLDecoder: Illegal hex characters in escape "
                                            + "(%) pattern - negative value");`

UriUtils.decode()  should be contain IllegalArgumentException,  return the original value .

kdavisk6:
Fixes #916

In certain cases, a Body Template will contain a JSON payload.  To
support this we are asking users to pct-encode the beginning and the
end of the JSON object when providing it to the RequestLine so we don't
reject it as an expression.  Doing this requires that the we decode those
markers before submitting the request.

This change updates that logic to only decode the first and last characters
only and not decode the entire payload, since Body values don't require
any type of encoding.

### DIFF
--- a/core/src/main/java/feign/template/BodyTemplate.java
+++ b/core/src/main/java/feign/template/BodyTemplate.java
@@ -51,11 +51,13 @@ public final class BodyTemplate extends Template {
     String expanded = super.expand(variables);
     if (this.json) {
       /* decode only the first and last character */
-      expanded = expanded.substring(
-          expanded.indexOf(JSON_TOKEN_START_ENCODED) + JSON_TOKEN_START_ENCODED.length());
-      expanded = JSON_TOKEN_START + expanded;
-      expanded = expanded.substring(0, expanded.lastIndexOf(JSON_TOKEN_END_ENCODED));
-      expanded += JSON_TOKEN_END;
+      StringBuilder sb = new StringBuilder();
+      sb.append(JSON_TOKEN_START);
+      sb.append(expanded,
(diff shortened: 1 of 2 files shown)


core/src/main/java/feign/template/BodyTemplate.java
```diff
@@ -35,10 +41,24 @@ public static BodyTemplate create(String template) {
...
   public String expand(Map<String, ?> variables) {
-    return UriUtils.decode(super.expand(variables), Util.UTF_8);
+    String expanded = super.expand(variables);
+    if (this.json) {
+      /* decode only the first and last character */
+      expanded = expanded.substring(
+          expanded.indexOf(JSON_TOKEN_START_ENCODED) + JSON_TOKEN_START_ENCODED.length());
+      expanded = JSON_TOKEN_START + expanded;
+      expanded = expanded.substring(0, expanded.lastIndexOf(JSON_TOKEN_END_ENCODED));
+      expanded += JSON_TOKEN_END;
```
velo: Shouldn't we be using a `StringBuilder` ?

kdavisk6: Yes, that would be better.  I'll make the changes.

PR #937, opened 2019-04-03
Don't URL encode fragment identifiers

### LINKED ISSUE #936: Fragment identifiers are handled incorrectly
When making requests with fragment identifiers in the URL, the hash mark "#" gets URL encoded as "%23" and put in front of the the query.

E.g. what should be "/path?key1=value1#fragment" becomes "/path%23fragment?key1=value1".

You can read more about fragment identifiers here:
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax
https://en.wikipedia.org/wiki/Fragment_identifier

rejierd:
Fixes #936

This is a super simple fix to illustrate the issue.
If a different solution is preferred, then I'm open for suggestions of course.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -43,6 +43,7 @@ public final class RequestTemplate implements Serializable {
   private final Map<String, QueryTemplate> queries = new LinkedHashMap<>();
   private final Map<String, HeaderTemplate> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
   private String target;
+  private String fragment;
   private boolean resolved = false;
   private UriTemplate uriTemplate;
   private HttpMethod method;
@@ -70,6 +71,7 @@ public final class RequestTemplate implements Serializable {
    * @param collectionFormat when expanding collection based variables.
    */
   private RequestTemplate(String target,
+      String fragment,
       UriTemplate uriTemplate,
       HttpMethod method,
       Charset charset,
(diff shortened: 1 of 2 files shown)


flisky:
Sorry to chime in, but I think it's a wrong behavier!

Fragment is [purely client-side](https://tools.ietf.org/html/rfc3986#section-3.5), and the browser sends the url without fragment. Nginx will strip off the fragment, and [curl does this so](https://curl.haxx.se/mail/lib-2011-11/0179.html), and more.

I think we could choose one of the following ways:

1. strip off the fragment
2. throw an exception to reject fragment
3. encode the `#` as user want it (the current behavier)

I prefer to 3, keep as it is. And user should be responsible for their input.

flisky:
friendly ping @rejierd @kdavisk6 , any opinion?

rejierd:
We've been using OkHttp, which doesn't send the fragment either according to the [documentation](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html). The same can probably be said of every other underlying HTTP client that Feign delegates to - meaning this commit doesn't actually conflict with RFC3986 as you suggest.

Our use case is that we automatically generate Feign clients from Swagger files with the help of openapi-generator. Our partner uses fragments in some of their paths, and we'd rather not manually strip out them out every time they update their documentation.

Whether Feign actually keeps the fragment is not that important to us. The important thing is that Feign properly handles valid URLs, by not URL encoding the hash sign and putting the fragment before the query.

What are your thoughts @kdavisk6?

kdavisk6:
@rejierd 

Agreed.  It is our job to process and prepare the request based on the specification provided by the Target interface and it's underlying Contract, and to present consistent, repeatable behavior when doing so.  In short, as @rejierd has stated, if a user specifies a fragment on the uri template, we need to ensure that we process it as outlined in RFC 6570 and produce a valid URL, which Feign was not doing properly in this case.

PR #940, opened 2019-04-09
Separate full logging of request and response

akileev:
I have M2M service that makes small requests and receives huge responses. I want to full log requests and log only headers for responses.


kdavisk6:
@akileev 

Thank you for your PR.  Before we can consider this, we will need an issue created.  Please see [HACKING](https://github.com/OpenFeign/feign/HACKING.md) for more information regarding suggesting and submitting changes.

PR #949, opened 2019-04-11
Support get value with name, For more easier to expansion Expander

cdmamata:
Support get value with name, For more easier to expansion Expander, so that you can use Custom Model for Get or Post method without RequestBody.

### DIFF
--- a/core/src/main/java/feign/Param.java
+++ b/core/src/main/java/feign/Param.java
@@ -49,6 +49,14 @@ public @interface Param {
      * Expands the value into a string. Does not accept or return null.
      */
     String expand(Object value);
+
+    /**
+     * support get value with name
+     */
+    default String expandWithName(Object value, String name) {
+      return expand(value);
+    }
+
   }
 
   final class ToStringExpander implements Expander {
@@ -57,5 +65,6 @@ public @interface Param {
     public String expand(Object value) {
       return value.toString();
     }
+
   }
 }
--- a/core/src/main/java/feign/ReflectiveFeign.java
+++ b/core/src/main/java/feign/ReflectiveFeign.java
@@ -212,11 +212,12 @@ public class ReflectiveFeign extends Feign {
         int i = entry.getKey();
         Object value = argv[entry.getKey()];
         if (value != null) { // Null values are skipped.
-          if (indexToExpander.containsKey(i)) {
-            value = expandElements(indexToExpander.get(i), value);
-          }
+          Object expandValue = value;
           for (String name : entry.getValue()) {
-            varBuilder.put(name, value);
+            if (indexToExpander.containsKey(i)) {
+                expandValue = expandElementsWithName(indexToExpander.get(i), value, name);
+            }
+            varBuilder.put(name, expandValue);
           }
         }
       }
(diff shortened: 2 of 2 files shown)


kdavisk6:
@cdmamata 

I do not understand the intent of this PR.  It appears to add no additional functionality, just an alternate way of looking up the `Expander`, which, frankly, is unnecessary.  Please read HACKING and CONTRIBUTING before either re-opening this PR or contributing another.

cdmamata:
> @cdmamata
> 
> I do not understand the intent of this PR. It appears to add no additional functionality, just an alternate way of looking up the `Expander`, which, frankly, is unnecessary. Please read HACKING and CONTRIBUTING before either re-opening this PR or contributing another.

The purpose of adding this method is it doesn't support to send a **GET REQUEST** with User-defined model with out RequestBody.
    if use the pr, then i can extend an Expander like this :
-----------------------------------------------------------------------
**ModelExpander.java**， Here's the point of this PR. and you can also use other ways，
```
public class ModelExpander implements Param.Expander {

    public String expand(Object value) {
        String objectJson = JSON.toJSONString(value);
        return objectJson;
    }

    //Here's the point of this PR. and you can also use other ways
    @Override
    public String expandWithName(Object value, String name) {
        String valueExpand = null;

        if(value != null){
            if(name != null) {
                try {
                    Map<String, Object> jsonMap = (Map<String, Object>)JSON.toJSON(value);

                    Object getValue = jsonMap.get(name);
                    if(getValue != null){
                        valueExpand = getValue.toString();
                    }
                } catch (Exception e) {
                    log.error("GET VALUE ERROR:", e);
                }
            }else {
                valueExpand = value.toString();
            }
        }

        return valueExpand;
    }
}
```
**ParamModel.java**
```
@Target({ElementType.PARAMETER})
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface ParamModel {
    String value() default "";
}
```
**ParamModelParameterProcessor.java**
```
public class ParamModelParameterProcessor implements AnnotatedParameterProcessor {

    private static final Class<ParamModel> ANNOTATION = ParamModel.class;

    public Class<? extends Annotation> getAnnotationType() {
        return ANNOTATION;
    }

    @Override
    public boolean processArgument(AnnotatedParameterContext context, Annotation annotation, Method method) {

        int parameterIndex = context.getParameterIndex();
        Class parameterType = method.getParameterTypes()[parameterIndex];
        MethodMetadata data = context.getMethodMetadata();

        Field[] fields = parameterType.getDeclaredFields();

        for(Field field: fields) {
            String name = field.getName();
            context.setParameterName(name);

            Collection query = context.setTemplateParameter(name, (Collection)data.template().queries().get(name));
            data.template().query(name, query);
        }
        data.indexToExpander().put(context.getParameterIndex(), new ModelExpander());

        return true;
    }
}
```
finally inject the Contract Bean
```
    @Bean
    public Contract feignContract(){
        List<AnnotatedParameterProcessor> processors = new ArrayList<>();
        processors.add(new ParamModelParameterProcessor());
        processors.add(new PathVariableParameterProcessor());
        processors.add(new RequestHeaderParameterProcessor());
        processors.add(new RequestParamParameterProcessor());
        return new SpringMvcContract(processors);
    }
```
then we can send Request use User-defined model in feign client like this:
```
Result<String> hello(@ParamModel DefinedModel model);
```

cdmamata:
@kdavisk6  
So I can expand the Expander  more flexibly.

PR #953, opened 2019-05-03
 Testing feign-example-github jar 

velo:


kdavisk6:
An integration test?  I love it

kdavisk6:
@velo 

Could we update the builds to call this manually instead of from Maven to reduce the chances of the API Rate Limit?

velo:
I wanna create an account and include auth so we have a higher rate limit.

but, you know, procrastinating is an art =)

Also, need to change packaging to try to reproduce the original issue that sparkled this change

velo:
@kdavisk6 added a new encrypted github token, now test works on travis

PR #959, opened 2019-05-08
Feature/replace deprecated body

mstrYoda:
In this pr the old `body()` method calls replaced with `requestBody().asBytes()` method which both exists in `Request` class. The intention is to remove deprecated code and keep source code clean.

### DIFF
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -190,6 +190,17 @@ public final class Request {
     return body.encoding;
   }
 
+  /**
+   * If present, this is the replayable body to send to the server. In some cases, this may be
+   * interpretable as text.
+   *
+   * @see #charset()
+   * @deprecated use {@link #requestBody()} instead
+   */
+  public byte[] body() {
+    return body.data;
+  }
+
   public Body requestBody() {
     return body;
   }
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -190,17 +190,6 @@ public final class Request {
     return body.encoding;
   }
 
-  /**
-   * If present, this is the replayable body to send to the server. In some cases, this may be
-   * interpretable as text.
-   *
-   * @see #charset()
-   * @deprecated use {@link #requestBody()} instead
-   */
-  public byte[] body() {
-    return body.data;
-  }
-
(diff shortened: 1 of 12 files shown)


core/src/main/java/feign/Request.java
```diff
@@ -190,17 +190,6 @@ public Charset charset() {
...
   }
 
-  /**
-   * If present, this is the replayable body to send to the server. In some cases, this may be
-   * interpretable as text.
-   *
-   * @see #charset()
-   * @deprecated use {@link #requestBody()} instead
-   */
-  public byte[] body() {
```
velo: I don't recall when we deprecated this.

If was on 9.x think is ok to drop it.

mstrYoda: If deprecated method is completely removed from usage, is not that ok to remove the method itself? I think removing it prevents further usage. Should I keep it or remove it?

velo: A `decoder` may exists out there that still uses it.

mstrYoda: I see.. Sorry I am not fully familiar on source code. I'll put it back.

Just a question out of this context, is it acceptable to use a 3rd party library (like apache commons lang) for some requirement?

velo: core has being clean from external dependencies and we are most likely to want to keep it that way.

Particularly about apache commons-*, every time I see them I read "old deprecated code that somehow still survived" =)

kdavisk6:
@mstrYoda 

And one more small housekeeping note, can you please update the description of this PR with the following format:

```
A short description of the change, your intention, and what you are trying to solve.

Fixes [issue number]
```

This helps us keep track of what issue the PR is for and provides us with background on your thinking and the intended outcome of this change.

mstrYoda:
> @mstrYoda
> 
> And one more small housekeeping note, can you please update the description of this PR with the following format:
> 
> ```
> A short description of the change, your intention, and what you are trying to solve.
> 
> Fixes [issue number]
> ```
> 
> This helps us keep track of what issue the PR is for and provides us with background on your thinking and the intended outcome of this change.

I updated description.

kdavisk6:
@mstrYoda 

Thanks for your contribution!

Issue #927, opened 2019-03-26
@QueryMap doesn't work with inheritance


StasKolodyuk:
Hi,

I've raised an [issue](https://github.com/spring-cloud/spring-cloud-openfeign/issues/168) in spring-cloud-openfeign project related to @SpringQueryMap not working with inheritance.

However, seems like the root cause is in `FieldQueryMapEncoder` of feign-core https://github.com/OpenFeign/feign/blob/10a4eadf4f3312573354672c9f5f9e322a30e5c5/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java#L69-L75

It uses `type.getDeclaredFields()` which returns only fields of a class without inherited fields.

Thanks

kdavisk6:
@StasKolodyuk 

Thanks for bringing this up.  If I recall, this is by design, as Feign's support for bean properties is intentionally simple.  This would be an enhancement for us.  So I've marked this issue as such.  If you would like this feature, please feel free to open a PR.

FatTigerWang:
> @StasKolodyuk
> 
> 谢谢你提出这个问题。如果我记得，这是设计，因为Feign对bean属性的支持是故意简单的。这对我们来说是一种增强。所以我已经标记了这个问题。如果您想要此功能，请随时打开公关。

support inheritance is very important

kdavisk6:
Resolved via #960

PR #960, opened 2019-05-08
This pr resolves QueryMap inheritance issue #927

mstrYoda:
Add logic for finding fields which comes via inheritance on `FieldQueryMapEncoder#ObjectParamMetadata.parseObjectType()` method.  This provides appending parent object's fields as query parameter.

Fixes [#927]

### DIFF
--- a/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
+++ b/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
@@ -16,7 +16,6 @@ package feign.querymap;
 import feign.QueryMapEncoder;
 import feign.codec.EncodeException;
 import org.apache.commons.lang3.reflect.FieldUtils;
-
 import java.lang.reflect.Field;
 import java.util.*;
 import java.util.stream.Collectors;
@@ -70,7 +69,7 @@ public class FieldQueryMapEncoder implements QueryMapEncoder {
 
     private static ObjectParamMetadata parseObjectType(Class<?> type) {
       return new ObjectParamMetadata(
-              FieldUtils.getAllFieldsList(type).stream()
+          FieldUtils.getAllFieldsList(type).stream()
               .filter(field -> !field.isSynthetic())
               .peek(field -> field.setAccessible(true))
               .collect(Collectors.toList()));
--- a/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
+++ b/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
@@ -15,7 +15,6 @@ package feign.querymap;
 
 import feign.QueryMapEncoder;
 import feign.codec.EncodeException;
-import org.apache.commons.lang3.reflect.FieldUtils;
 import java.lang.reflect.Field;
 import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


mstrYoda:
@velo is it okey to adding dependency or I should remove it and implement the behaviour built in way?

kdavisk6:
@mstrYoda 

We would prefer to keep `core` dependency free.  If this is the approach you want to take, I recommend looking at what Commons Lang does and see if you can create a new implementation inspired by it, and update this PR.

mstrYoda:
> @mstrYoda
> 
> We would prefer to keep `core` dependency free. If this is the approach you want to take, I recommend looking at what Commons Lang does and see if you can create a new implementation inspired by it, and update this PR.
> 
> And one more small housekeeping note, can you please update the description of this PR with the following format:
> 
> ```
> A short description of the change, your intention, and what you are trying to solve.
> 
> Fixes [issue number]
> ```
> 
> This helps us keep track of what issue the PR is for and provides us with background on your thinking and the intended outcome of this change.

I've removed the dependency and add the logic itself as plain java code. The code is almost the same code as apache's. I think it should not be a problem since apache common has same licence with feign which is apache licence.

I want to hear your opinion too. Also thanks for review.

core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
```diff
@@ -67,11 +67,17 @@ private ObjectParamMetadata(List<Field> objectFields) {
...
+      List<Field> allFields = new ArrayList();
+
+      for (Class currentClass = type; currentClass != null; currentClass =
+          currentClass.getSuperclass()) {
+        Collections.addAll(allFields, currentClass.getDeclaredFields());
+      }
+
+      return new ObjectParamMetadata(allFields.stream()
+          .filter(field -> !field.isSynthetic())
+          .peek(field -> field.setAccessible(true))
```
kdavisk6: This is one area I'm unsure of.  This will allow serialization for all fields, regardless of access modifier.  I'm not sure we want make this decision for our users and take on the responsibility of doing so.  

My initial reaction is to avoid changing accessibility of fields and additionally filter on `isAccessible` fields.  Approaching it this way will prevent issues for users on JDK 11, as `setAccessible` is no longer considered _safe_, and we can document this behavior.  

The security implications of this are also important, but worth a separate discussion.

I do understand that this is existing logic, however; it is an important area to consider.

Thoughts?

mstrYoda: I agree. But changing the behavior is not affects the existing users? Also it might be written this way because of author did not want to deal serialization over getter&setters. As you said before it is another discussion.

mstrYoda: @velo I think we are waiting feedbacks to merge this pr. Thoughts?

velo: Well, I think for the `FieldQueryMap` it's ok to grab everything.

The only concern I had was the apache commons dependency, but, with that solved I'm good with it.

kdavisk6: Ok with me then.

mstrYoda:
I see that you add this pr to milestone 10.3.0, as a curiosity I want to ask you to when is this pr get merges :) @kdavisk6

PR #961, opened 2019-05-09
Adds support for per request timeout options. Fixes #562

### LINKED ISSUE #562: Per request timeout options
We have an API that we are using that expects us to use long polling for a GET request. We can change the timeouts to support that, but there are other calls in the API that don't support long polling and we want the timeouts to be smaller for those calls. Is there a way to support this with Feign?

meifans:
Adds support for per request timeout options
Fixes [#562](https://github.com/OpenFeign/feign/issues/562)

As @kdavisk6 said

```
interface MyClient {
   @RequestLine("GET /data")
   public MyResponse getMyData(Request.Options options);
}
```
In Spring Cloud 
```
interface MyClient {
   public MyResponse getMyData(Request.Options options);
}
```

### DIFF
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -187,8 +187,10 @@ final class SynchronousMethodHandler implements MethodHandler {
     if (argv == null || argv.length == 0) {
       return this.options;
     }
-    return Stream.of(argv).filter(o -> o instanceof Options).findAny().map(o -> (Options) o)
-                 .orElse(this.options);
+    return (Options) Stream.of(argv)
+        .filter(o -> o instanceof Options)
+        .findFirst()
+        .orElse(this.options);
   }
 
   static class Factory {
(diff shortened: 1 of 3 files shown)


kdavisk6:
@meifans 

Thank you for this PR.  Before we continue, please correct the formatting changes included in the PR by running `mvn clean install` and pushing the updated files.

core/src/main/java/feign/SynchronousMethodHandler.java
```diff
@@ -181,6 +183,14 @@ Object decode(Response response) throws Throwable {
     }
   }
 
+  Options findOptions(Object[] argv) {
+    if (argv == null || argv.length == 0) {
+      return this.options;
+    }
+    return Stream.of(argv).filter(o -> o instanceof Options).findAny().map(o -> (Options) o)
```
kdavisk6: I don't believe you need the `findAny()`.  This can be simplified:

```java
return (Options) Stream.of(argv)
         .filter(arg -> arg instanceof Options)
         .findFirst()
         .orElse(this.options);
```

velo:
😮

meifans:
A new PR: https://github.com/OpenFeign/feign/pull/970

PR #968, opened 2019-05-13
Migrating to Sonatype OSSRH from JCenter

### LINKED ISSUE #697: post request with url query parameters
I currently have replace httpurlconnection to feign-okhttp in my projects. I use the following edition:
        <dependency>
            <groupId>io.github.openfeign</groupId>
            <artifactId>feign-okhttp</artifactId>
            <version>9.6.0</version>
        </dependency>
but I got a problem when I post this url: "http://gclife-platform-service:20000/v1/user/login/log?userId=ea9dd2f707a54b11b03193c35b2ad3a4&error=SUCCESS"
The log from server side as following:
Received [1.1
cache-control: no-cache
postman-token: 8814eadd-5bb1-4cac-b072-8d608763ad01
devicechannel: gclife_agent_app
language: EN_US
ostype: Android
user-agent: PostmanRuntime/6.3.2
accept: */*
x-forwarded-host: 192.168.2.32:13000
x-forwarded-proto: http
x-forwarded-prefix: /app
host: 192.168.2.32:13000
x-forwarded-port: 13000
x-forwarded-for: 192.168.2.35
accept-encoding: gzip
connection: Keep-Alive
Content-Type: application/json
Content-Length: 212

{"deviceChannel":"gclife_agent_app","osType":"Android","deviceId":null,"appVersion":null,"district":null,"language":"EN_US","phoneModel":null,"osVersion":null,"coordiate":null,"networkType":null,"ipAddress":null}]

But "Received [1.1" should be: "Received [POST v1/user/login/log?userId=ea9dd2f707a54b11b03193c35b2ad3a4&error=SUCCESS HTTP/1.1 "
So this caused the parse problem. 
"2018-05-05 10:10:04.076 DEBUG 6525 --- [io-20000-exec-9] o.apache.coyote.http11.Http11Processor   : Error parsing HTTP request header
java.lang.IllegalArgumentException: Invalid character found in method name. HTTP method names must be tokens"

How can I get the correct log? Can you have this fixed?

### LINKED ISSUE #967: Replace JCenter with OSSRH
### Proposal
JCenter is great, however it's process for syncing to Maven Central can be cumbersome and is difficult to automate, as our most recent attempts to release have shown.  I would like to propose that we deploy our snapshots and releases to OSSRH instead.

### Benefits
- Signed artifacts prevent accidental or intentional package hijaking see [A Confusing Depdency](https://blog.autsoft.hu/a-confusing-dependency/)
- Automatic synchronization of releases to Central
- Easily add and remove authors by working with Sonatype.  In JCenter, only the owner of the package can manage the process.

### Drawbacks
- Artifacts must be signed by one of the approved developers.
- JCenter users will need to wait for JCenter to mirror new releases.
- Onboarding process for new maintainers will be longer, due to signing requirements.

I feel that the benefits of moving to OSSRH outweigh the drawbacks of leaving JCenter, specifically when it comes to the risks of fraudulent packages alone.

kdavisk6:
Fixes #967 

This change replaces JCenter with Sonatype OSSRH for snapshot and
release management.  To complete this change, we will need to add
additional information to Travis, specifically the GPG information
to sign the artifacts during a release.

It's my opinion that those values should be in Travis directly and
not here in this repository.  I'm open to other suggestions and
feedback on this approach.

Additional changes include:

* Updating BOM template to include SCM per OSSRH rules.
* Removed Bintray and GitHub credentials, they are no longer needed.


pom.xml
```diff
@@ -138,6 +139,18 @@
...
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+    <snapshotRepository>
+      <id>ossrh</id>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+    </snapshotRepository>
+  </distributionManagement>
+
+  <!--
   <distributionManagement>
```
velo: IIRC, this section is required

pom.xml
```diff
@@ -719,6 +743,20 @@
               </execution>
             </executions>
           </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
```
velo: Where are the signature files?

kdavisk6: I wanted to discuss this part.  I see two options, we can include the them in the repository, encrypted and reference them from the `/travis` folder, or we can include them at the Travis account level, and inject them as environment variables.

Here are links to the two approaches:

[Continuous Integration to maven central (for free)](https://www.phillip-kruger.com/post/continuous_integration_to_maven_central/) *this one uses environment variables at the account level*

[Complete guide to continuous deployment to maven central from Travis CI](http://www.debonair.io/post/maven-cd/) *this one uses encrypted files and values in the `travis.yml`*

I can see both working, but I would prefer not to have the values or encrypted files in the repository, if we can avoid it.

pom.xml
kdavisk6: Yes, I've replaced it.  I need to remove the commented section.  That contains the JCenter settings.

pom.xml
velo: I like `encrypted files and values in the travis.yml`

PR #970, opened 2019-05-16
Adds support for per request timeout options. Fixes #562

### LINKED ISSUE #562: Per request timeout options

meifans:
Adds support for per request timeout options 
Fixes #562

Change:
* Add Options UT
* Ignore Options when set bodyIndex

As @kdavisk6 said

```java
interface MyClient {
   @RequestLine("GET /data")
   public MyResponse getMyData(Request.Options options);
}
```
In Spring Cloud
```java
interface MyClient {
   public MyResponse getMyData(Request.Options options);
}
```

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -13,9 +13,6 @@
  */
 package feign;
 
-import static feign.Util.checkState;
-import static feign.Util.emptyToNull;
-import feign.Request.HttpMethod;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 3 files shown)


meifans:
@kdavisk6  I submit a new PR for #562

meifans:
Is anybody there?

kdavisk6:
@meifans 

I've approved this PR and marked it "ready to merge"  This indicated to any other contributors that, without any additional feedback, we will merge this request in a few days.

eugeniyk:
Hey @kdavisk6, I think it's quite important feature to be reflected in the docs! WDYT?

Also is my understanding correct - there's no way to specify default timeouts per RequestLine?

In my vision client settings hierarchy should be like that
1. request-level settings (has)
2. endpoint-level settings (doesn't have, and that's were I struggle)
3. client-instance-level settings (has)
4. default client in current jvm settings (doesn't have, (3) is enough)
5. defaults from feign repo (has)

PR #979, opened 2019-05-23
Adding support for X-RateLimit-Reset HTTP header

tandrup:
Some services use the `X-RateLimit-Reset` HTTP header instead of the `Retry-After` header. Try to parse both.


kdavisk6:
@tandrup 

The build for this PR is broken due to compilation errors.  If you are still interested in submitting this change, please update this PR to allow the build to pass and we will accept it.

Issue #932, opened 2019-03-28
feign.httpclient.ApacheHttpClient should not setEntity when request body is null in order to support application/x-www-form-urlencode


starrybleu:
In recently, I replaced "com.netflix.feign:feign-httpclient:8.18.0" with "io.github.openfeign:feign-httpclient" along with spring boot application, because which from netflix is deprecated.

I found a difference between two dependencies that one from netflix let entity null on building request, the other from openfeign(this repo.) let entity be an empty byte array.

So, `org.apache.http.client.methods.RequestBuilder` comes to finally build  a `HttpUriRequest` of `InternalRequest` skipping making entityCopy be `UrlEncodedFormEntity`, even if headers has 'Content-Type: application/x-www-form-urlencoded'.

So I finally got 400 Bad Request from other spring cloud server.

I think, as a wrapper of apache http client, feign.httpclient  should follow apache's implementation.

kdavisk6:
Support for forms including multi-part and x-www-form-urlencoded requests has been moved into another library, feign-form.  Please add that dependency, update your Feign Target [per their documentation](https://github.com/OpenFeign/feign-form) and try again.

starrybleu:
@kdavisk6 Thanks you for your response. I have already known feign-form library, you mentioned, and had a dependency in my project. However, I think the problem is not regarded with feign-form or en/decoding. The fact that `feign.httpclient.ApacheHttpClient` makes entity not null cause `org.apache.http.client.methods.RequestBuilder` not to UrlEncode parameters is the problem.

For additional information, spring boot 2.0.8.RELEASE, spring-cloud-starter-openfeign:2.0.2.RELEASE, io.github.openfeign:feign-httpclient:9.7.0, feign-form:3.3.0

kdavisk6:
@starrybleu 

I recommend, if at all possible, upgrading to the latest versions of those libraries.  Feign is now at 10.2.3 and Spring Cloud Stater Open Feign is at 2.1.0.RELEASE.

There is a strong possibility your issue has been addressed.

starrybleu:
@kdavisk6 

Thank you for your notice. 

If it was resolved when I check out, I'll close this issue.

starrybleu:
I have tried, but doesn't work.

I will create a pull request that I think what should be fixed is.

starrybleu:
[...]

PR #982, opened 2019-05-29
#932 To support application/www-form-urlencoded working with ApacheHt…

starrybleu:
…tpClient and RequestBuilder.

Only when entity is null, then RequestBuilder creates UrlFormEncodedEntity.

Ref. RequestBuilder.build()
```java
...
        HttpEntity entityCopy = this.entity;
        if (this.parameters != null && !this.parameters.isEmpty()) {
            if (entityCopy != null || !"POST".equalsIgnoreCase(this.method) && !"PUT".equalsIgnoreCase(this.method)) {
                try {
                    uriNotNull = (new URIBuilder(uriNotNull)).setCharset(this.charset).addParameters(this.parameters).build();
                } catch (URISyntaxException var5) {
                }
            } else {
                entityCopy = new UrlEncodedFormEntity(this.parameters, this.charset != null ? this.charset : HTTP.DEF_CONTENT_CHARSET);
            }
        }
...
```

### DIFF
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -68,7 +68,7 @@ public class ApacheHttpClientTest extends AbstractClientTest {
     @Path("/withBody")
     public String withBody(@QueryParam("foo") String foo, String bar);
 
-    @PUT
+    @GET
     @Path("/withoutBody")
     public String withoutBody(@QueryParam("foo") String foo);
   }
--- a/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
+++ b/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
@@ -145,8 +145,6 @@ public final class ApacheHttpClient implements Client {
       }
 
       requestBuilder.setEntity(entity);
-    } else {
-      requestBuilder.setEntity(new ByteArrayEntity(new byte[0]));
     }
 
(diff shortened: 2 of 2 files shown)


httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -68,7 +68,7 @@ public void queryParamsAreRespectedWhenBodyIsEmpty() throws InterruptedException
     @Path("/withBody")
     public String withBody(@QueryParam("foo") String foo, String bar);
 
-    @PUT
+    @GET
```
kdavisk6: I would prefer that we keep the `@PUT` test case and add the `@GET` test case.  I am surprised that the `@PUT` test case was testing without a body, that's not really valid anyway.

Could you please restore the `@PUT` and correct it to send a body?

starrybleu: Sure, I'll do that tomorrow. Thank you for feedback.

velo: Yeah, agree with Kevin on this one.

Add a new test and leave existing as is.

starrybleu: After some tries, I cannot figure out how to keep PUT with `@QueryParam` and `UrlEncodedFormEntity` by depending on `RequestBuilder`. For now, I kind of give up. Sorry for no step forward.

Issue #978, opened 2019-05-23
Document POST body example


wulftone:
After a bunch of experimentation, I figured out that you can use a POJO as a POST body, but I couldn't find it anywhere in the documentation.  e.g. You can do
```java
public interface Boop {
    @RequestLine("POST /boop/{someParam}")
    @Headers({
        "Accept: application/json",
        "Content-Type: application/json"
    })
    BoopResponse createBoopUsingPOST(Boop boop, @Param("someParam") String someParam);

    class Boop {
        public String value;
    }
}
```
But the tricky part is that the POJO param (`boop`, in the example), _must_ come first in the parameter list of the method.  That was difficult to find out!

kdavisk6:
@wulftone 

Thanks for this example.  It is missing from the documentation.  You can define the body of the request two ways:

* annotating the parameter with `@Body`
* the first method parameter without an annotation is assumed to be the `Body`

We should update the documentation to explain this and add you example.

eiselems:
@wulftone  @kdavisk6 is this like you had it in mind?

wulftone:
Pretty much, I commented on the pr: https://github.com/OpenFeign/feign/pull/986#issuecomment-500635251

eiselems:
@wulftone is it really failing for you when you put the boop paramater in any other position?
I updated the github-example in the repository and it works in all positions.

wulftone:
Yes, let me verify exactly why and get back to you... It is interesting that you found it doesn't matter in the example

sleroy:
Indeed I lost 10 minutes to find an Example. Could not guess it easily from the readme...

eiselems:
@sleroy hoped the example would be sufficient 

```
  @RequestLine("POST /repos/{owner}/{repo}/issues")
  void createIssue(Issue issue, @Param("owner") String owner, @Param("repo") String repo);

```

How should we improve it?

PR #986, opened 2019-06-10
Add POST example to README.md and example-github

eiselems:
Fixes: #978

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,23 @@ Usage typically looks like this, an adaptation of the [canonical Retrofit sample
 interface GitHub {
   @RequestLine("GET /repos/{owner}/{repo}/contributors")
   List<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
+
+  @RequestLine("POST /repos/{owner}/{repo}/issues")
+  void createIssue(Issue issue, @Param("owner") String owner, @Param("repo") String repo);
+
 }
 
 public static class Contributor {
   String login;
   int contributions;
+}
+
+public static class Issue {
+  String title;
+  String body;
+  List<String> assignees;
+  int milestone;
+  List<String> labels;
 }
 
 public class MyApp {
--- a/example-github/src/main/java/example/github/GitHubExample.java
+++ b/example-github/src/main/java/example/github/GitHubExample.java
@@ -13,14 +13,12 @@
  */
 package example.github;
 
-import feign.Feign;
-import feign.Logger;
-import feign.Param;
-import feign.RequestLine;
-import feign.Response;
+import feign.*;
 import feign.codec.Decoder;
+import feign.codec.Encoder;
 import feign.codec.ErrorDecoder;
 import feign.gson.GsonDecoder;
+import feign.gson.GsonEncoder;
 import java.io.IOException;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


wulftone:
Looks like a good example! I think it might help people to let them know explicitly that the parameter order matters, or they'll just get confused about it.. e.g. in development they might make a GET method with some params, then change it to a POST and add a third body parameter, which is wrong, then read the readme again and maybe notice the subtlety. I prefer spelling it out in sentences that this ordering is a hard requirement.

eiselems:
Good point. Since I am a bit spoiled since I use SpringBoot together with spring-feign ...
how would I do the following with just openFeign?

```
@RequestMapping(method = RequestMethod.POST, value = "/stores/{storeId}", consumes = "application/json")
Store update(@PathVariable("storeId") Long storeId, Store store);
```

@kdavisk6 mentioned the @body annotation, but that requires a templated body which is not what I want here 😕 ? Anybody could shine some light on it?


Also: Just took this from the spring cloud example page, there it seems that the order is not important - which I find slightly interesting.

kdavisk6:
`@Body` can be used two ways, one on the method with a template, the other is on a parameter.  To take your example and convert it to vanilla feign:

```java
@RequestLine("POST /stores/{storeId}")
@Header("Content-Type: application/json")
Store update(@Param("storeId") Long storeId, @Body Store store);
```
The information missing from the documentation is that `@Body` can be omitted if the last parameter in the method signature is the request body.  @eiselems, your intuition about the parameter location is on point.  

Here is an example where the `@Body` is required.

```java
@RequestLine("POST /stores/{storeId}/item/{item}")
@Header("Content-Type: application/json")
Store update(@Param("storeId") Long storeId, @Body Store store, @Param("item") Long item);
```
Hope this helps.

kdavisk6:
@velo @eiselems 

Not sure why Travis is not running the build.  It's breaking on the environment set up, however there are no environment changes in this PR.  Either of you have any insight on what we should look for to fix?  I kicked Master just to see if it something in the repo, but that build is working just fine.

eiselems:
Don't stress the build, not in a hurry to merge this asap. Take your time for figuring it out.

@kdavisk6 what is confusing me is that the interface approach doesn't work on my machine.
It is complaining that @Body requires a parameter and is not applicable to a parameter, is there another @Body for parameters which is not on the buildPath of the example?

velo:
> @velo @eiselems
> 
> Not sure why Travis is not running the build. It's breaking on the environment set up, however there are no environment changes in this PR. Either of you have any insight on what we should look for to fix? I kicked Master just to see if it something in the repo, but that build is working just fine.

I think is travis failing to install java... let's retry in a day or 2.

eiselems:
@kdavisk6 Sorry for probably needing too much hand holding for this one ...

I discovered in the example that all of:
```
@RequestLine("POST /repos/{owner}/{repo}/issues")
void createIssue(@Param("owner") String owner, Issue issue, @Param("repo") String repo);
```

eiselems:
So I think in the current state it should be good to merge

wulftone:
Huh. I definitely encountered issues if the pojo wasn't first in the list. I'll have to dig a little to find out why that is...

kdavisk6:
I found the issue with the build, so I'm going to fix that in another PR before the end of the week.  I'm ok with passing on that check for now.

@wulftone If you want, please create an issue asking for clarification on how to use the `@Body` parameter so we can track it.

PR #989, opened 2019-06-18
JacksonEncoder avoids intermediate String request body

schlosna:
Serialize Jackson request object directly to UTF-8 byte array without intermediate String representation.

See https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway

### DIFF
--- a/jackson/src/main/java/feign/jackson/JacksonEncoder.java
+++ b/jackson/src/main/java/feign/jackson/JacksonEncoder.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.Module;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import feign.RequestTemplate;
 import feign.codec.EncodeException;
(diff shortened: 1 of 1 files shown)


jackson/src/main/java/feign/jackson/JacksonEncoder.java
```diff
@@ -48,7 +49,7 @@ public JacksonEncoder(ObjectMapper mapper) {
   public void encode(Object object, Type bodyType, RequestTemplate template) {
     try {
       JavaType javaType = mapper.getTypeFactory().constructType(bodyType);
-      template.body(mapper.writerFor(javaType).writeValueAsString(object));
+      template.body(mapper.writerFor(javaType).writeValueAsBytes(object), StandardCharsets.UTF_8);
```
kdavisk6: The only change I think we should make here is use our `Util.UTF_8` constant for consistency.

schlosna:
Thanks for the review @kdavisk6 . I've updated the PR with your suggestion.

kdavisk6:
@schlosna 

I've marked this PR as ready to merge.  If there are no additional comments in the next few days, this PR will be merged and is attached to our 10.3.0 Release.

schlosna:
Excellent, thank you!

Issue #998, opened 2019-06-28
Don't add slash to URI  if it is a matrix parameter like (;matrixparam=value)


ranjitc5:
After updating **openfeign**  latest version I found that  matrix parameters behaviour got changed. Seems its similar like **OpenFeign#889**

**OpenFeign version : 10.1.0 or Latest**

**Please find below scenarios in which its creating a problem**
 When we have API end point with dynamic  path and we try to get details using matrix params its creating a problem 
For example:
```
Path ("{prefix}")
EmployeController {

  EmployeeList getEmployee(@PathParam("prefix") PathSegment pathSegment,
            @QueryParam("flag") String flag) throws Exception;
} 
```
Now when we try to call above mention api with latest version and after calling **RequestTemplate -->> uri(matrix paramete details, true):**  its generate template **uri** like below
```
api/gituser/;includeDetails=false
```

**Which was with earlier version  like below**
```
api/gituser;includeDetails=false
```
I have checked the same and I think we should add one more condition with  **semicolon**  in RequestTemplate --> url(string, boolean) method like 
 **!uri.startsWith(";")**  for the matrix parameter.

I have few more scenarios which are creating same issue with latest version of open feign
**Let me know if you need more details**

ranjitc5:
Can anyone please check above pull request ?

ranjitc5:
Hi @kdavisk6 ,

I can see my PR is merged and as per milestone it should be in  **10.2.4** and **10.3.0**  but I can't see any release with above mentioned version.

Can you please help me for the same ?

kdavisk6:
It's been built and released as we speak.

PR #999, opened 2019-07-01
Avoided url appending with slash when matrix parameter exists

ranjitc5:
Fixed an issue of passing matrix parameter using RequestTemplate

More details at  https://github.com/OpenFeign/feign/issues/998


ranjitc5:
Can anyone please check it ?

kdavisk6:
I’m just back from vacation. I will review this later today also.

ranjitc5:
> Could you also include a unit test? To prevent from breaking again in the future?

ranjitc5:
> Could you also include a unit test? To prevent from breaking again in the future?

I have added UT as well now.

PR #1001, opened 2019-07-04
Add unit tests for feign.auth.Base64

JohnLBergqvist:
I've analysed your codebase and noticed that `feign.auth.Base64` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.

### DIFF
--- a/core/src/test/java/feign/auth/Base64Test.java
+++ b/core/src/test/java/feign/auth/Base64Test.java
@@ -1,0 +1,479 @@
+/**
+ * Copyright 2012-2019 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


kdavisk6:
@JohnLBergqvist 

We appreciate the attempt, but this type of test generation is not that useful for us.  In particular, the class that is under test is something that we should replace with an internal JDK variant.  May I suggest that next time you would like to contribute with generated code, create an issue for us so we can discuss and work with you.

Also, I think what you are doing is really good and can be very useful to others, it's just not for us.  So keep going.

JohnLBergqvist:
@kdavisk6 Many thanks for your feedback. Are there any particular classes you would like us to test?

velo:
I agreed with @kdavisk6 , generated tests then to be hard to read and maintain. And even work, they may "lock" bugs as features.  I also would like to avoid that.

PR #1004, opened 2019-07-08
Fixes 1003: Do not wrap exceptions in RuntimeException

### LINKED ISSUE #1003: Reactive wrappers for feighn wraps everything in RuntimeException
Whenever you use feign reactive wrappers and custom error decoder, it will wrap the underlying exception in RuntimeException which makes it hard to react on. One way I fixed it in my own project is I make all my custom exceptions extend FeignException, but that adds extra load on support because FeignException contract is constantly changing.

It would be nice if Reactive feign stopped wrapping exceptions in RuntimeExceptions.

SimY4:
Fixes #1003

Initially, the ReactiveInvocationHandler.java was wrapping method invocation in a Callable. That has a few drawbacks, one of which is the fact that we cannot propagate Throwable property and we need to wrap it in some new Exception class.
This PR wraps methodInvocation in a generic Publisher instead. Publisher allows us to signal Throwables back as an error signal and without additional wrapping.

### DIFF
--- a/reactive/src/test/java/feign/reactive/ReactiveInvocationHandlerTest.java
+++ b/reactive/src/test/java/feign/reactive/ReactiveInvocationHandlerTest.java
@@ -19,28 +19,25 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyZeroInteractions;
-import feign.FeignException;
 import feign.InvocationHandlerFactory.MethodHandler;
 import feign.RequestLine;
 import feign.Target;
-import feign.reactive.ReactorInvocationHandler;
-import feign.reactive.RxJavaInvocationHandler;
 import io.reactivex.Flowable;
+
+import java.io.IOException;
 import java.lang.reflect.Method;
 import java.util.Collections;
-import org.junit.Rule;
+
+import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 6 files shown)


reactive/src/test/java/feign/reactive/ReactiveInvocationHandlerTest.java
```diff
@@ -62,17 +64,33 @@ public void invokeOnSubscribeReactor() throws Throwable {
...
+            .verify();
     verify(this.methodHandler, times(1)).invoke(any());
   }
 
-  @SuppressWarnings("unchecked")
   @Test
   public void invokeFailureReactor() throws Throwable {
-    this.thrown.expect(RuntimeException.class);
-    given(this.methodHandler.invoke(any())).willThrow(new RuntimeException("Could Not Decode"));
-    given(this.method.getReturnType()).willReturn((Class) Class.forName(Mono.class.getName()));
```
SimY4: This test didn't even run because it throws NullPointerException because this.method was never set. And the test passes because NullPointerException is runtime exception and ExpectedException was marking the test as passed.

kdavisk6:
Thanks for this PR. I will review and provide feedback later today.

reactive/src/test/java/feign/reactive/ReactiveInvocationHandlerTest.java
kdavisk6: Thank you for pointing that out.

PR #1012, opened 2019-07-19
Respect decode404 flag and decode 404 response body

### LINKED ISSUE #809: JacksonDecoder probably ignores decode404 flag
For this line of code, https://github.com/OpenFeign/feign/blob/6ad5584f037164aacafc72307e6a8e44a62b3f92/jackson/src/main/java/feign/jackson/JacksonDecoder.java#L48

Even decode404 is enabled, after this https://github.com/OpenFeign/feign/blob/2d761cb6a92efc6b4a95e7092c945f47431b76cd/core/src/main/java/feign/SynchronousMethodHandler.java#L133, calling the decoder, but JacksonDecoder still will ignore the 404 response.

Please correct me if it's not working as this way.

dharmeshjogadia:
Fix the following decoder to decode 404 response body

- GsonDecoder

- JacksonJaxbJsonDecoder

- JacksonDecoder

- JacksonIteratorDecoder

- JAXBDecoder

- SAXDecoder

Fixes #809


kdavisk6:
@dharmeshjogadia 

Thank you for this PR, as you can see, there are number of issues.  I suggest that you make sure that you are using the most recent version of master in your fork.  Please either update your fork or rebase against master before continuing.

dharmeshjogadia:
Hey @kdavisk6 
I have updated the PR against the master and fixes the test as well.

Issue #845, opened 2018-11-20
Url field in FeignException


maxvetrenko:
When RetryableException was caused we have information about url only in exception message. I think it will be useful to have url like an exception field. 

Some argumentation. Suppose we have several feign clients with same api type but different url. When RetryableException was caused we have information about unhealthy url only in exception message. Url  like field will give more possibilities to handle error more correctly

kdavisk6:
I think what we may want to do is what we've done elsewhere and store the original request as part of `FeignException`.  Would this work for you?

PR #1039, opened 2019-08-18
Request field added to FeignException

finnetrolle:
Fixes: #845 

Request field added to FeignException. Other exceptions changed too except Encode Exception (Request is null here)

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -25,27 +25,33 @@ public class FeignException extends RuntimeException {
   private static final long serialVersionUID = 0;
   private int status;
   private byte[] content;
+  private Request request;
 
-  protected FeignException(int status, String message, Throwable cause) {
+  protected FeignException(int status, String message, Request request, Throwable cause) {
     super(message, cause);
     this.status = status;
+    this.request = request;
   }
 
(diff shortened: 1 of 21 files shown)


finnetrolle:
Thank you for your feedback. I thought about making old constructors @Deprecated... Do you think the good way is to deprecate them or just return as is?

kdavisk6:
The older constructors are still relevant and there are additional use cases where we may want to allow exceptions without a request.  That being said, there is no need for deprecation, just additional options.

core/src/main/java/feign/FeignException.java
```diff
@@ -25,27 +26,59 @@
...
+    this.status = status;
+    this.request = checkNotNull(request, "request");
+  }
+
+  protected FeignException(int status, String message, Request request, Throwable cause,
+      byte[] content) {
+    super(message, cause);
+    this.status = status;
+    this.content = content;
+    this.request = checkNotNull(request, "request");
```
kdavisk6: We should have a real message here.  This will result in an NPE with a message of "request".  Can you change to read something like "request is required"

finnetrolle: I just grep project to check how checkNotNull invokes and made same. But ok, changed! )

finnetrolle:
done

PR #1045, opened 2019-08-23
GH-801: Adding support for JDK Proxy

### LINKED ISSUE #801: Proxied and non-proxied Feign client on the same jvm
I'm using Feign & spring to build a core component, which is interacting with a variety of rest APIs. One of which is behind a proxy. So, the issue is that I can not setup the proxy using the system properties since it will be applied on JVM level and will affect other clients.

In order to resolve this, i've added this library  "com.netflix.feign:feign-okhttp" to my project. Which resulted on a bad config and Spring's messages converters behaved wierdly. So, i guess the best solution's to provide another implementation of the Client interface that support proxy.

kdavisk6:
Fixes #801

Adding a `Proxied` client implementation that extends the `Default`
client allowing for a JDK Proxy, along with explict credential support.

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -13,26 +13,36 @@
  */
 package feign;
 
+import static feign.Util.CONTENT_ENCODING;
+import static feign.Util.CONTENT_LENGTH;
+import static feign.Util.ENCODING_DEFLATE;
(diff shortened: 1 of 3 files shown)


kdavisk6:
@herrhilmi

I've updated your pull request and created one here.  This contains your changes plus our suggestions.

Issue #985, opened 2019-06-08
Logging to other client's file


Donutellko:
Hello. 
I have two Feign clients in a Spring Boot project, they are both configured like this (Kotlin):
```
@Configuration
class FirstConfig {
    @Bean
    fun getFirstClient(): FirstClient {
        return Feign.builder()
                .logger(Logger.JavaLogger().appendToFile("logs/first.log"))
                .logLevel(Logger.Level.FULL)
                .decoder(JacksonDecoder())
                .encoder(JacksonEncoder())
                .target(FirstClient::class.java, "...")
    }
}
```
And similarly for `SecondClient` and `second.log` (all the names are changed).

The issue is that I get two log files mixed. Something like that:
logs/second.log:
```
[FirstClient#someMethod] ---> GET http://.../... HTTP/1.1
...
[FirstClient#someMethod] <--- END HTTP (3629-byte body)
[SecondClient#otherMethod] ---> POST https://.../... HTTP/1.1
...
[SecondClient#send] ---> END HTTP (767-byte body)
[FirstClient#oneMoreMethod] ---> GET http://.../... HTTP/1.1
...
```
And the second.log is full of FirstClient's logs and vice versa. `diff` says the files are **almost**  identical, but not completely -- two or three logs from first client does not appear in second's file.

I use version 10.1.0:
```
dependencies {
    compile 'io.github.openfeign:feign-core:10.1.0'
    compile 'io.github.openfeign:feign-jackson:10.1.0'
    ...
```

Thank you.

Donutellko:
I found that the problem is caused by 165th line in `feign\Logger.java`:
```logger = java.util.logging.Logger.getLogger(Logger.class.getName());```
So this method returns the same object for both instances of `feign\JavaLogger.java`. Would the change of an identifier be an appropriate solution?

dharmeshjogadia:
Hey @Donutellko 
I have one suggestion here to solve this problem, rather than using the default constructor of `feign\JavaLogger.java` we can introduce constructor with a parameter name or class. Something like
`public JavaLogger(Class<?> clazz)` and  `public JavaLogger(String name)`. what do you thing @kdavisk6 ?

PR #1047, opened 2019-08-24
Unexpected JavaLogger behavior fixed

finnetrolle:
Investigating #985 task I've created solution. I'm not sure this solution is good, but I tried to explain my thoughts and solution inside code. Also, I copy explanation here.
Also I'm not sure it's a good idea to create log files during test run. But I don't really know how to remove them after tests...

> This constructor can be used to create just one logger. In this way nothing will change.
> Example = Logger.JavaLogger().appendToFile("logs/first.log")
> 
> But if this constructor is used several times in different places with different files we will have bad situation
> We call constructor and that means for client that we created new instance, that will write logs to given file
> But in fact we created only wrapper. Our worker class java.util.logging.Logger instance will be same for all
> JavaLoggers because or getLogger() method logic. And each time we create and apply file to logger -
> we apply file handler to same logger.
> 
> Correct way to change this - is to ask client for unique logger name to give loggerManager change to differ files
> But some users already using this method for single call without any class name. So, I decided to make workaround
> 
> I'll create logger with Logger.class name for first instance.
> But each other calls of constructor will check if someone already created first logger. If so - I'll create
> logger with name Logger.class[1-..]. We will have Logger, Logger1, Logger2, etc. This looks not very good, but
> provides backward compatibility. But I mark this constructor as Deprecated and recommend other constructor with
> the name to provide inner logger for.

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -16,6 +16,7 @@ package feign;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.FileHandler;
 import java.util.logging.LogRecord;
 import java.util.logging.SimpleFormatter;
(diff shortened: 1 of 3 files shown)


kdavisk6:
@finnetrolle 

I think you may gone a little too far in this change.  I agree with deprecating the default constructor, but by attempting to provide an internal workaround, you are adding complexity that we don't need.  I think that by simply marking the default constructor as deprecated and add the new constructor will suffice.

We will remove the default constructor in Feign 11.

finnetrolle:
@kdavisk6 I absolutely agree with you about workaround. But I decided that it's better to make proposal and this is the best I can provide to keep default constructor and change its logic. I'll fix it now

kdavisk6:
I understand and appreciate the effort.  We can add additional documentation to let user's know of the limitation and our suggested workaround.

finnetrolle:
@kdavisk6 I'm going to change javadoc for deprecated method to warn users

core/src/main/java/feign/Logger.java
```diff
@@ -164,8 +165,35 @@ protected void log(String configKey, String format, Object... args) {
...
+    }
+
+    /**
+     * Constructor for JavaLogger class
+     * 
+     * @param loggerName - A name for the logger. This should be a dot-separated name and should
+     *        normally be based on the package name or class name of the subsystem, such as java.net
+     *        or javax.swing
+     */
+    public JavaLogger(String loggerName) {
```
velo: maybe a constructor that accepts a class instead? like most log frameworks

finnetrolle: maybe both? I think just a class could be not enough. For example if I have 2 same clients such 
```SomeService client1 = ...; 
SomeService client2 = ...;
```
And in some reason I want to write them to different files. It will force developers to make special class.

velo:
Makes sense

finnetrolle:
Done =)

Issue #944, opened 2019-04-11
Support ServiceLoader


SchulteMarkus:
Java has a class [java.util.ServiceLoader ](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/ServiceLoader.html) (in this package available from JDK6 onwards).

> A facility to load implementations of a service.
> A service is a well-known interface or class for which zero, one, or many service providers exist. A service provider (or just provider) is a class that implements or subclasses the well-known interface or class. A ServiceLoader is an object that locates and loads service providers deployed in the run time environment at a time of an application's choosing. Application code refers only to the service, not to service providers, and is assumed to be capable of choosing between multiple service providers (based on the functionality they expose through the service), and handling the possibility that no service providers are located.

**Feature request**: Please support ServiceLoader by placing the required files in META-INF/services.

**Background**

You may already have heard of [SubstrateVM](https://github.com/oracle/graal/tree/master/substratevm) (part of [GraalMV](https://www.graalvm.org/)) 

> Substrate VM is a framework that allows ahead-of-time (AOT) compilation of Java applications under closed-world assumption into executable images or shared objects (ELF-64 or 64-bit Mach-O).

SubstrateVM supports _META-INF/services_ as part of the compilation process (done in https://github.com/oracle/graal/issues/563, released https://www.graalvm.org/docs/release-notes/#10-rc8). 

So, if you support ServiceLoader, you add (part of) support for SubstrateVM, too.

**Example**

[Micronaut](https://micronaut.io/) can be compiled using SubstrateVM. One example for META-INF/services can be found at https://github.com/micronaut-projects/micronaut-core/tree/master/inject/src/main/resources/META-INF/services

velo:
@SchulteMarkus do you think a generated implementation would solve the graalVM problem?

https://github.com/velo/feign/pull/3/files

Could you test and help me automate tests with graal VM so I can continue developing this feature?

Issue #935, opened 2019-04-03
Illegal reflective access by feign.DefaultMethodHandler in Java 11


gavinws:
Similar to #393 

This can be reproduced on Java 11 (probably also 9 and 10) with the [Github example](https://github.com/OpenFeign/feign/tree/10a4eadf4f3312573354672c9f5f9e322a30e5c5/example-github) after renaming the package to `notfeign.example.github` (or anything that isn't under `feign.*`). Can be compiled with Java 8 or 11 (the compiler doesn't complain either way).

I reproduced on [Zulu](https://www.azul.com/downloads/zulu/) and [Corretto](https://docs.aws.amazon.com/corretto/latest/corretto-11-ug/downloads-list.html) 11.0.2 (all OpenJDK 11 implementations should behave the same):
```
$ java -version
openjdk version "11.0.2" 2019-01-15 LTS
OpenJDK Runtime Environment Zulu11.29+3-CA (build 11.0.2+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.29+3-CA (build 11.0.2+7-LTS, mixed mode)

$ java -version
openjdk version "11.0.2" 2019-01-15 LTS
OpenJDK Runtime Environment Corretto-11.0.2.9.1 (build 11.0.2+9-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.2.9.1 (build 11.0.2+9-LTS, mixed mode)
```

```
$ java -jar target/feign-example-github-10.2.1-SNAPSHOT.jar
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by feign.DefaultMethodHandler (file:/Users/gavin/test/feign/example-github/target/feign-example-github-10.2.1-SNAPSHOT.jar) to field java.lang.invoke.MethodHandles$Lookup.IMPL_LOOKUP
WARNING: Please consider reporting this to the maintainers of feign.DefaultMethodHandler
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Let's fetch and print a list of the contributors to this org.
[GitHub#repos] ---> GET https://api.github.com/users/openfeign/repos?sort=full_name HTTP/1.1
[GitHub#repos] <--- HTTP/1.1 200 OK (554ms)
...
```

Note that using the `--illegal-access=warn` JVM option just changes the warning to just the single line calling out `feign.DefaultMethodHandler`.

[...]

PR #1060, opened 2019-09-01
Declarative contracts

velo:
This is a preparation work for feign code generation that I'm working on this branch:
https://github.com/velo/feign/tree/apt-generator

With generated client, we will be able to generate clients that will work with graalVM and also don't have java 11 reflective warning. Fixing both:
https://github.com/OpenFeign/feign/issues/935
https://github.com/OpenFeign/feign/issues/944

The key work for the code generation is the new `DeclarativeContract`. Which would allow to reuse the annotation logic when generating client. So, same contract code used by feign reflectively would be used on APT code generation.

The idea of this PR is to review the design of this new `DeclarativeContract`

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -14,15 +14,21 @@
 package feign;
 
 import java.lang.annotation.Annotation;
-import java.lang.reflect.*;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
 import java.net.URI;
-import java.util.*;
-import java.util.function.BiConsumer;
-import java.util.function.BiFunction;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import feign.Contract.DeclarativeContract.ClassAnnotationProcessor;
 import feign.Request.HttpMethod;
 import static feign.Util.checkState;
 import static feign.Util.emptyToNull;
@@ -106,9 +112,15 @@ public interface Contract {
         if (parameterAnnotations[i] != null) {
           isHttpAnnotation = processAnnotationsOnParameter(data, parameterAnnotations[i], i);
         }
+
+        if (isHttpAnnotation) {
+          data.ignoreParamater(i);
+        }
+
         if (parameterTypes[i] == URI.class) {
           data.urlIndex(i);
-        } else if (!isHttpAnnotation && parameterTypes[i] != Request.Options.class) {
+        } else if (!isHttpAnnotation && parameterTypes[i] != Request.Options.class
+            && !data.isAlreadyProcessed(i)) {
           checkState(data.formParams().isEmpty(),
               "Body parameters cannot be used with form parameters.");
           checkState(data.bodyIndex() == null, "Method has too many Body parameters: %s", method);
@@ -220,6 +232,7 @@ public interface Contract {
     Map<Class<Annotation>, ParameterAnnotationProcessor<Annotation>> parameterAnnotationProcessors =
         new HashMap<>();
 
+    @Override
     public final List<MethodMetadata> parseAndValidatateMetadata(Class<?> targetType) {
       // any implementations must register processors
       return super.parseAndValidatateMetadata(targetType);
@@ -263,17 +276,17 @@ public interface Contract {
      * @return true if you called {@link #nameParam(MethodMetadata, String, int)} after finding an
      *         http-relevant annotation.
      */
+    @Override
     protected final boolean processAnnotationsOnParameter(MethodMetadata data,
                                                           Annotation[] annotations,
                                                           int paramIndex) {
-      return Arrays.stream(annotations)
+      Arrays.stream(annotations)
           .filter(
               annotation -> parameterAnnotationProcessors.containsKey(annotation.annotationType()))
-          .map(annotation -> parameterAnnotationProcessors
+          .forEach(annotation -> parameterAnnotationProcessors
               .getOrDefault(annotation.annotationType(), ParameterAnnotationProcessor.DO_NOTHING)
-              .process(annotation, data, paramIndex))
-          .collect(Collectors.reducing(Boolean::logicalOr))
-          .orElse(false);
+              .process(annotation, data, paramIndex));
+      return false;
     }
 
     @FunctionalInterface
@@ -332,7 +345,8 @@ public interface Contract {
     @FunctionalInterface
     public interface ParameterAnnotationProcessor<E extends Annotation> {
 
-      ParameterAnnotationProcessor<Annotation> DO_NOTHING = (ann, data, i) -> false;
+      ParameterAnnotationProcessor<Annotation> DO_NOTHING = (ann, data, i) -> {
+      };
 
       /**
        * @param annotation present on the current parameter annotation.
@@ -342,7 +356,7 @@ public interface Contract {
        * @return true if you called {@link #nameParam(MethodMetadata, String, int)} after finding an
        *         http-relevant annotation.
        */
-      boolean process(E annotation, MethodMetadata metadata, int paramIndex);
+      void process(E annotation, MethodMetadata metadata, int paramIndex);
     }
 
     /**
@@ -424,20 +438,17 @@ public interface Contract {
         if (!data.template().hasRequestVariable(name)) {
           data.formParams().add(name);
         }
-        return true;
       });
       super.registerParameterAnnotation(QueryMap.class, (queryMap, data, paramIndex) -> {
         checkState(data.queryMapIndex() == null,
             "QueryMap annotation was present on multiple parameters.");
         data.queryMapIndex(paramIndex);
         data.queryMapEncoded(queryMap.encoded());
-        return true;
       });
       super.registerParameterAnnotation(HeaderMap.class, (queryMap, data, paramIndex) -> {
         checkState(data.headerMapIndex() == null,
             "HeaderMap annotation was present on multiple parameters.");
         data.headerMapIndex(paramIndex);
-        return true;
       });
     }
 
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -14,18 +14,15 @@
 package feign;
 
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
+import java.lang.reflect.*;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import feign.Contract.DeclarativeContract.ClassAnnotationProcessor;
 import feign.Request.HttpMethod;
 import static feign.Util.checkState;
 import static feign.Util.emptyToNull;
@@ -98,7 +95,7 @@ public interface Contract {
       }
       checkState(data.template().method() != null,
           "Method %s not annotated with HTTP method type (ex. GET, POST)",
-          method.getName());
+          data.configKey());
       Class<?>[] parameterTypes = method.getParameterTypes();
       Type[] genericParameterTypes = method.getGenericParameterTypes();
 
@@ -169,7 +166,6 @@ public interface Contract {
       }
     }
 
-
     /**
      * Called by parseAndValidateMetadata twice, first on the declaring class, then on the target
      * type (unless they are the same).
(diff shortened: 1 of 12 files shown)


core/src/main/java/feign/Contract.java
```diff
@@ -211,99 +207,224 @@ protected void nameParam(MethodMetadata data, String name, int i) {
...
+   * processed and how each annotation modifies {@link MethodMetadata}
+   */
+  abstract class DeclarativeContract extends BaseContract {
 
-    static final Pattern REQUEST_LINE_PATTERN = Pattern.compile("^([A-Z]+)[ ]*(.*)$");
+    Map<Class<Annotation>, ClassAnnotationProcessor<Annotation>> classAnnotationProcessors =
+        new HashMap<>();
+    private Map<Class<Annotation>, MethodAnnotationProcessor<Annotation>> methodAnnotationProcessors =
+        new HashMap<>();
+    Map<Class<Annotation>, ParameterAnnotationProcessor<Annotation>> parameterAnnotationProcessors =
```
velo: this maps are awful, and kind useless, cause I'm force to loop the entries to figure out if an annotation is instance of class.  Need a better data structure to hold this data, but, I ran out of ideas.

Also, probably a single data structure would be enough, no need for one per processor.

core/src/main/java/feign/Contract.java
```diff
@@ -211,99 +207,224 @@ protected void nameParam(MethodMetadata data, String name, int i) {
...
+      void process(E annotation, MethodMetadata metadata);
+    }
+
+    /**
+     * Called while class annotations are being processed
+     *
+     * @param annotation to be processed
+     * @param processor function that defines the annotations modifies {@link MethodMetadata}
+     */
+    protected <E extends Annotation> void registerClassAnnotation(Class<E> annotation,
```
velo: `register*` methods modifies internal state, this is not good, maybe need to find a way to "register" processors during instance creation and then assign to a unmodifiable collection.

core/src/main/java/feign/Contract.java
kdavisk6: I don’t think we need explicit maps. The annotations declare where they can be used and if we scan the annotations on either a class method or parameter, we will only see valid ones.

velo: my concern was the same annotation being present on class and method, but, demanding different logic to process...

for instance jaxrs `@Path`... 
when you annotate class you will set that as uri
but, when you annotate method, you need to append.... get it could be the same code, but there can be other cases where it may not be

velo:
> This class has become quite large. What are your thoughts on breaking out the Annotation Processors into their own classes and possibly their own package?

ow yes, as a mater of fact, annotation process will go to it's own module, since it will have dependencies to all contracts.

core/src/main/java/feign/Contract.java
kdavisk6: I did something similar in Feignx. I delegated that responsibility to MethodMetadata.  I added a guard in the annotation processor to check the state of MethodMetadata and then make the appropriate modification. I’ll add more context once I can get back to my pc.

core/src/main/java/feign/Contract.java
kdavisk6: We can use this as long as we make it safe.  However, I agree that the processors should be provided during construction to make the contract immutable

core/src/main/java/feign/Contract.java
velo: I guess I could go ahead, change all contracts to be declarative.

If all can work with a single code path I could squash this into a single map.

core/src/main/java/feign/Contract.java
kdavisk6: Take a look at this: https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java

This is really close to what I think you are going for.

kdavisk6: I created my own POC here: https://github.com/kdavisk6/feign/tree/declarative-contract-poc/core/src/main/java/feign/contract

I think we can consolidate this down into a single Map and introduce the concept of a context to wrap up the execution state.  Take a look and let me know what you think.

velo: I did apply this "declarative" concept to jaxrs contract... and for that, I would need something more complex to evaluate if an annotation should be processed or not.

I'm thinking about a Predicate... if I go for a predicate, then, a single map (or list) should be enough for all cases.

kdavisk6: I looked through the JAXRS Contract.  I may be missing something, but I did not see anything that requires the contract to conditionally parse an annotation outside of skipping ignored annotations.  Is that what you are looking into?

velo: For jaxrs, we don't check if @GET is present, if check if method are annotated with any annotation that in turn is annotated with @HttpMethod

core/src/main/java/feign/Contract.java
```diff
@@ -211,99 +207,238 @@ protected void nameParam(MethodMetadata data, String name, int i) {
...
-        Map<String, Collection<String>> headers = toMap(headersOnType);
-        headers.putAll(data.template().headers());
-        data.template().headers(null); // to clear
-        data.template().headers(headers);
-      }
+    protected final void processAnnotationOnClass(MethodMetadata data, Class<?> targetType) {
+      Arrays.stream(targetType.getAnnotations())
+          .forEach(annotation -> classAnnotationProcessors
+              .getOrDefault(annotation.annotationType(), ClassAnnotationProcessor.DO_NOTHING)
+              .process(annotation, data));
```
kdavisk6: This feels odd to me.  I can see that you are going for functional style, but the empty function feels forced.  This can be accomplished with a simple loop and lookup and that doesn't require the empty function:

```java
for (Annotation annotation : targetType.getAnnotations()) {
   AnnotationProcessor processor = this.classAnnotationProcessors.get(annotation);
   if (processor != null) {
      process.process(annotation, data);
   }
}
```

Can you help me understand if this is more than just a style choice?

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
```diff
@@ -44,10 +41,12 @@ protected MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method me
     return super.parseAndValidateMetadata(targetType, method);
   }
 
-  @Override
-  protected void processAnnotationOnClass(MethodMetadata data, Class<?> clz) {
-    Path path = clz.getAnnotation(Path.class);
-    if (path != null && !path.value().isEmpty()) {
+  public JAXRSContract() {
+    super.registerClassAnnotation(Path.class, (path, data) -> {
+      if (path.value().isEmpty()) {
```
kdavisk6: `Path` cannot have an empty value, by specification.  this should be an error.

jaxrs2/src/main/java/feign/jaxrs2/JAXRS2Contract.java
```diff
@@ -16,23 +16,20 @@
...
-    // masc20180327. parameter with unsupported jax-rs annotations should not be passed as body
-    // params.
+  public JAXRS2Contract() {
+    // parameter with unsupported jax-rs annotations should not be passed as body params.
     // this will prevent interfaces from becoming unusable entirely due to single (unsupported)
     // endpoints.
     // https://github.com/OpenFeign/feign/issues/669
-    return (annotationType == Suspended.class ||
-        annotationType == Context.class);
+    super.registerParameterAnnotation(Suspended.class, (ann, data, i) -> true);
```
kdavisk6: If they are not supported, then do we even need to register them?

velo: You need to let the contract parser know that this parameters should have an special handling. In this case, should be ignored.

If nothing is done feign will assume that is the body, failing of there is more than one.

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
velo: `checkState` instead?

Will do the chance.

core/src/main/java/feign/Contract.java
velo: 100% pure stylish choice.

Original change had this if to check for null.

Not having it allows me to do a leaner lambda function.

jaxrs2/src/main/java/feign/jaxrs2/JAXRS2Contract.java
kdavisk6: Oh, right, I remember now.  Thanks.

core/src/main/java/feign/Contract.java
kdavisk6: I don't normally like to haggle over style, but I really thing that if we don't have a processor, we shouldn't create an empty one just to avoid a `null` check.  If you really want to do this, I suggest the following change:

```java
private Optional<AnnotationProcessor> getProcessor(Annotation annotation) {
   return Optional.ofNullable(this.processors.get(annotation));
}
```
then you can do this:
```java
Arrays.stream(targetType.getAnnotations())
   .forEach(annotation -> {
      getProcessor(annotation)
         .ifPresent(processor -> processor.process(annotation, data))
  });
```
keeps the functional style without the empty lambda.  Thoughts?

core/src/main/java/feign/Contract.java
```diff
@@ -211,99 +219,237 @@ protected void nameParam(MethodMetadata data, String name, int i) {
     }
   }
 
-  class Default extends BaseContract {
+  /**
+   * {@link Contract} base implementation that works by declaring witch annotations should be
+   * processed and how each annotation modifies {@link MethodMetadata}
+   */
+  public abstract class DeclarativeContract extends BaseContract {
```
velo: note to self:
Extract this class to a new source file before merging.  Enough is enough =)

jaxrs/src/main/java/feign/jaxrs/JAXRSContract.java
velo: I restored the original code, for now won't try to enforce the contract around this.

core/src/main/java/feign/Contract.java
velo: Take a look on the new approach using a guarded function.

core/src/main/java/feign/Contract.java
velo: Predicate for annotationProcessors solved this

core/src/main/java/feign/Contract.java
kdavisk6: An alternative approach that doesn't require the empty lambda is to use `Optional` and use `ifPresent`.

```java
private Optional<AnnotationProcessor> getAnnotationProcessor(
   Annotation annotation) {
      return Optional.ofNullable(this.processors.get(annotation));
}
```
```java
Arrays.stream(target.getAnnotations())
   .forEach(annotation -> {
      getAnnotationProcessor(annotation)
         .ifPresent(processor -> processor.process(annotation, data));
   });
```

This removes the need for the empty lambda.  What do you think?

kdavisk6:
I'm still confused and I think your current approach relies on others having sufficient mental space to interpret what's going on.  There are three areas where I'm still not able to fully understand:

1.  `DO_NOTHING`

This says to me that I want to skip processing but I'm doing my best to avoid a `if !null` check.  `null` checks are not evil and are completely appropriate.  However, I do understand that you are going for functional style.  I'll reiterate that I think that using `Optional` is a better approach here.  It states clearly that we are not going to process an annotation unless we have a processor registered.

```java
public void processAnnotationsOnClass(MethodMetadata data, Class<?> target) {
  Arrays.stream(target.getAnnotations())
    .forEach(annotation -> getProcessor(annotation)
        .ifPresent(
            annotationProcessor -> annotationProcessor.process(annotation, data)));
}

private Optional<AnnotationProcessor> getProcessor(Annotation annotation) {
  return Optional.ofNullable(this.processors.get(annotation));
}
```

2.  Guarding using a `Predicate`

I'm still unclear on how this is intended to be used.  If its to determine if the processor should be called, we should leave that decision up to processor implementations.  Having the `Predicate` muddies the intention of `AnnotationProcessor` leaving it unclear what is the expected way to guard against execution.  It's still possible to skip processing in the `process` function with the `test` method present, bypassing the predicate entirely.  If we want to keep it, we'll need to find a way to make it's use crystal clear.

3.  Ignoring Parameters

I don't think ignoring is the right verb.  I suspect what you are looking for is to identify which parameters are not annotated and how to deal with them.  The current implementation relies on the result of `processAnnotationsOnParameter` to indicate if the parameter was processed.  With the annotation processor approach, that's no longer necessary as we will only process parameters that have annotations.  We can refactor `processAnnotationsOnParameter` to process the `Parameter`, allowing implementations to deal with those special scenarios.

```java
protected void processAnnotationsOnParameter(MethodMetadata methodMetadata,
    Parameter parameter, int index) {
  Annotation[] annotations = parameter.getAnnotations();
  if (annotations != null && annotations.length != 0) {
    for (Annotation annotation : annotations) {
      AnnotationProcessorContext context =
          new AnnotationProcessorContext(methodMetadata, parameter, index);
      this.getAnnotationProcessor(annotation)
          .ifPresent(processor -> processor.process(annotation, context));
    }
  } else {
    /* no annotations, special processing, like request options, etc... */
    if (Request.Options.class.isAssignableFrom(parameter.getType())) {
      /* process */
    }
  }
}
```

I created a gist that demonstrates these thoughts and my approach.  Now, I could be completely off here.  If I am, please let me know.

https://gist.github.com/kdavisk6/bb6db782c8fe5284ed0ea3ffb1a5d012

velo:
> 1. `DO_NOTHING`

Ow, there is one left for `ParameterAnnotationProcessor`, I will remove it. Thanks.

> 2. Guarding using a Predicate

So, my goal for `DeclarativeContract` is separate Class + annotation navigation from "how each annotation affects MethodMetadata". Main reason for that is that Annotation Processor Tool does NOT have classes, so, for APT, I need to navigate the annotations using a different API.

This translates into `DeclarativeContract` having the 3 process `processAnnotationsOnClass`, `...OnMethod` and `...OnParam` all being final. So the decision if an annotation should be processed need to be placed on this Predicate so APT can reuse this code.

> 3. Ignoring Parameters

This is used for JAXRS contract. On jaxrs you can inject things like the `@Context HttpRequest request`. When `JAXRSContract` detects `@Context` it needs to inform feign that this parameter must not be processed. There, my idea of calling it ignored.

Now, if the contract didn't made any decision about a parameter, feign assumes that one is a BODY. That can either be due to no annotation or some annotation that contract don't care about.

> gist

`AnnotationProcessorContext` supporting both `param` and `non-param` annotations would lead to index being null some times, I prefer to have 2 possible processors with their own signature... that way, when processing a Parameter you do have the the extra parameter to `index` and must deal with it.

velo:
> Thanks for the explanation and now I understand your approach. I have some reservations about the ignoring approach, it still feels awkward, but it works and I don't want to hold this up any more. 

Ok, going to merge it as is.

PR #1065, opened 2019-09-08
Create new module for apache http 5

velo:


hc5/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>feign-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <version>5.0-beta5</version>
```
kdavisk6: Should we add this while the version is still beta?

hc5/src/main/java/feign/hc5/ClassicHttpClient.java
```diff
@@ -0,0 +1,230 @@
...
+ * This module directs Feign's http requests to Apache's
+ * <a href="https://hc.apache.org/httpcomponents-client-5.0.x/index.html">HttpClient 5</a>. Ex.
+ *
+ * <pre>
+ * GitHub github = Feign.builder().client(new ClassicHttpClient()).target(GitHub.class,
+ * "https://api.github.com");
+ */
+/*
+ */
+public final class ClassicHttpClient implements Client {
```
kdavisk6: Is `ClassicHttpClient` appropriate?  This may be confusing to users.  I think it's OK for this to also be called `ApacheHttpClient` since it's in different packages.  What do you think?

hc5/src/main/java/feign/hc5/ClassicHttpClient.java
```diff
@@ -0,0 +1,230 @@
...
+    ClassicHttpRequest httpUriRequest;
+    try {
+      httpUriRequest = toClassicHttpRequest(request, options);
+    } catch (final URISyntaxException e) {
+      throw new IOException("URL '" + request.url() + "' couldn't be parsed into a URI", e);
+    }
+    final HttpHost target = HttpHost.create(URI.create(request.url()));
+    final HttpClientContext context = new HttpClientContext();
+    // per request timeouts
+    final RequestConfig requestConfig =
```
kdavisk6: This may be better in a function, just to simplify this method.

hc5/pom.xml
velo: Have being using the Alpha here for a while.

At some point we need to bring it in. Now seems a good time :)

hc5/src/main/java/feign/hc5/ClassicHttpClient.java
velo: Well, I'm hoping to eventually implement an async client as well...

So far, we have being avoiding class names conflicts for contacts.

hc5/pom.xml
kdavisk6: Fair enough

hc5/src/main/java/feign/hc5/ClassicHttpClient.java
kdavisk6: Then maybe `ApacheHttp5Client?`

velo:
@kdavisk6 changes in place

PR #1067, opened 2019-09-10
Ignore GPG Keys in Travis

kdavisk6:
Travis needs a local copy of the gpg key to install the library
in Maven Central.  This change ignores that local file because
our formatting checks consider this file "changed" creating
false-negative build results.


velo:
👍

PR #1069, opened 2019-09-12
Spring4 contract

velo:
I wanna test the declarative contract a bit further, and decide to bring an old project for spring 4 contracts under feign umbrella.
https://github.com/velo/feign-spring

This is does not replace/compete with `spring-cloud-openfeign` as it only works with older versions of spring.


velo:
I will merge as is so I can cut a release with final set of changes to contract. Worse case scenario I can deleted spring4 support later.

PR #1081, opened 2019-09-23
Updating Apache HttpClient to 4.5.10 (OpenFeign#1080)

mxmlnglt:


mxmlnglt:
I noticed Both jobs failed: https://travis-ci.org/OpenFeign/feign/builds/588377888
with such errors:
> 1. Exception in thread "main" example.github.GitHubExample$GitHubClientError: API rate limit exceeded for 34.66.50.208. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
> 
> 2. The command "./mvnw clean install -B" exited with 1.

Is this normal?

velo:
A env variable was missing, I fixed. Good to go.

PR #1090, opened 2019-10-04
Add composed Spring annotations support

beargummy:
Add Spring's composed annotations support (like `@GetMapping`). These are convenience methods that are quite frequently used.

Changeset might look a bit repetitive, however, to process annotations differently it would require bigger refactoring which is not intended by this minor auxiliary PR. As was mentioned in #1069, worse case this entire spring4 support can be deleted later.

### DIFF
--- a/spring4/src/main/java/feign/spring/SpringContract.java
+++ b/spring4/src/main/java/feign/spring/SpringContract.java
@@ -45,39 +45,39 @@ public class SpringContract extends DeclarativeContract {
     });
 
 
-    registerMethodAnnotation(GetMapping.class, (postMapping, data) -> {
-      appendMappings(data, postMapping.value());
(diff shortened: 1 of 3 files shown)


spring4/src/main/java/feign/spring/SpringContract.java
```diff
@@ -51,6 +44,42 @@ public SpringContract() {
         data.template().method(requestMapping.method()[0].name());
     });
 
+
+    registerMethodAnnotation(GetMapping.class, (postMapping, data) -> {
```
velo: Can you rename all the `postMapping` to `mapping`?

beargummy: sure, missed this one ;)

PR #1095, opened 2019-10-12
1048 simple illustration - more details in the default feignexception…

jerzykrlk:
… message


core/src/main/java/feign/FeignException.java
```diff
@@ -131,6 +136,11 @@ public static FeignException errorStatus(String methodKey, Response response) {
     } catch (IOException ignored) { // NOPMD
     }
 
+    String message = new FeignExceptionMessageBuilder()
+        .withResponse(response)
+        .withRequestMethod(response.request().httpMethod())
```
velo: could read the method from the response, eliminating the need to call `withRequestMethod` method

jerzykrlk: fixed

jerzykrlk:
Hi @velo, thanks for feedback.

I've polished up the code, and I've abbreviated the response body when it's too long (well, more than 200 chars didn't seem readable to me).

I'm not sure about #1091 - give me some time to understand that.

Can I have one more question for the future?

Quite often, I want to further modify the response body preview. For example - some APIs render 5xx of 4xx pages with HTML. Well, not the perfect world, but I have to live with that. So I need to strip all the HTML to find out, what happened.

I'd like to make extracting the response body customisable and injectable, but I can't seem to find a good way with current code (static method of `FeignException` doing all the hard work). How could I do that?

velo:
> How could I do that?

ErrorDecoder may be?!

PR #1098, opened 2019-10-21
Fixes a typo in Contract.java method

cezar-tech:
* The method parseAndValidatateMetadata has been renamed to parseAndValidateMetadata and has been deprecated;
* Replaced all usages along the project;
* Documented which method to use instead of the deprecated one.

Signed-off-by: Cézar Augusto <cezaraugustome@gmail.com>

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -33,13 +33,16 @@ public interface Contract {
    *
    * @param targetType {@link feign.Target#type() type} of the Feign interface.
    */
-  // TODO: break this and correct spelling at some point
-  List<MethodMetadata> parseAndValidatateMetadata(Class<?> targetType);
+  List<MethodMetadata> parseAndValidateMetadata(Class<?> targetType);
 
   abstract class BaseContract implements Contract {
 
+    /**
+     * @param targetType {@link feign.Target#type() type} of the Feign interface.
+     * @see #parseAndValidateMetadata(Class)
+     */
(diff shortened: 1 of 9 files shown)


cezar-tech:
It appears the build failed because of some rate limit ?

> Exception in thread "main" example.github.GitHubExample$GitHubClientError: API rate limit exceeded for 104.197.122.201. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

core/src/main/java/feign/Contract.java
```diff
@@ -31,15 +31,35 @@
   /**
    * Called to parse the methods in the class that are linked to HTTP requests.
    *
+   * @deprecated Use {@link #parseAndValidateMetadata(Class)} instead
    * @param targetType {@link feign.Target#type() type} of the Feign interface.
    */
   // TODO: break this and correct spelling at some point
-  List<MethodMetadata> parseAndValidatateMetadata(Class<?> targetType);
+  @Deprecated
+  default List<MethodMetadata> parseAndValidatateMetadata(Class<?> targetType){
```
velo: I would say to take out the typo method

cezar-tech:
I have ammended the commit to not pollute the master branch

core/src/main/java/feign/Contract.java
cezar-tech: I have ammended the commit to not pollute the master branch

PR #1105, opened 2019-11-03
Bump reactive dependencies.

SimY4:
Make reactive contract work with the latest project reactor class hierarchy.

UPD: related change in project reactor: https://github.com/reactor/reactor-core/pull/1527

### DIFF
--- a/reactive/pom.xml
+++ b/reactive/pom.xml
@@ -28,9 +28,9 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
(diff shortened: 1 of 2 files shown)


SimY4:
@velo I can, but the problem here is that the current version of reactive extensions will not work with the latest projectreactor release.  They've changed their class hierarchy and added an extra layer between Mono and `Publisher` interface. So direct interface lookup is failing. I changed the logic to check not only direct interfaces but whether the class is an instance of `Published` in general.

SimY4:
Can this be merged in? and released? This is currently blocking the spring boot upgrade for us.

PR #1106, opened 2019-11-03
404 + Optional = empty()

michaldo:
In HTTP language missing resource is explained by 404. In Java, the same is Optional.empty(). Feign should by default map 404 to empty

Currently, the mapping can be forced by property `decode404=true`. It means that `Optional` without `decode404=true` is useless - `Optional.empty` in never returned, because Feign transforms 404 into `FeignException`.

But applying `decoder404=true` has bad impact. Let's see:

There are two methods:

```
Optional<User> findByLogin(String login);
User findById(Long id);
```

When user is searched by login, it is normal business case that user is not found - for example, wrong login is typed. In such case I expect endpoint should return 404 - in HTTP language user not found.
Because method returns Optional<User>, it means 'empty' is normal case. Optional forces developer to handle that case. I expect 404 will be translated into Optional.empty

But when user is searched by id, it means user was already found - maybe in previous action. Missing user is not normal case. I expect that endpoint should return 404, but Feign should throw FeignException. That is because method returns User, and null in not acceptable - if acceptable, Optional<User> will be applied. Contrary to previous case, I want force developer to NOT handle the case. The case - no user for given id - is runtime error which should be handled like any other runtime error - by general handler.

Conclusion:
If method returns Optional, 404 should be mapped to Optional.empty.
If method returns concrete class, 404 should trigger FeignException.
"decode404=true" force developer to null check every time and is not a solution.
Solution is this commit

### DIFF
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -22,6 +22,7 @@ import feign.Request.Options;
 import feign.codec.DecodeException;
 import feign.codec.Decoder;
 import feign.codec.ErrorDecoder;
+import feign.optionals.OptionalDecoder;
 import static feign.ExceptionPropagationPolicy.UNWRAP;
 import static feign.FeignException.errorExecuting;
 import static feign.FeignException.errorReading;
@@ -148,6 +149,10 @@ final class SynchronousMethodHandler implements MethodHandler {
           shouldClose = closeAfterDecode;
           return result;
         }
+      } else if (response.status() == 404 && OptionalDecoder.isOptional(metadata.returnType())) {
+        Object result = decode(response);
+        shouldClose = closeAfterDecode;
+        return result;
       } else if (decode404 && response.status() == 404 && void.class != metadata.returnType()) {
         Object result = decode(response);
         shouldClose = closeAfterDecode;
--- a/core/src/main/java/feign/optionals/OptionalDecoder.java
+++ b/core/src/main/java/feign/optionals/OptionalDecoder.java
@@ -42,7 +42,7 @@ public final class OptionalDecoder implements Decoder {
     return Optional.ofNullable(delegate.decode(response, enclosedType));
   }
 
-  static boolean isOptional(Type type) {
+  public static boolean isOptional(Type type) {
     if (!(type instanceof ParameterizedType)) {
       return false;
     }


velo:
I think assuming `404` is same as empty (`Optional` or `Collection`) is problematic.

Recently on the project I work at, we configure feign to translate 404 to empties... we had to revert that.

404 can happen if you hit the wrong URL. By silently converting to empty you may not see some problem.

I favor current behavior. Unless @kdavisk6 has strong feeling to change this, I would say to leave as is.

michaldo:
Indeed, that is one of the worst concept in REST implementation: using protocol code 404 (wrong url) to satisfy application needs - not found.

_(Unfortunately, nobody ask me at the beginning of REST boom and now it too late to fix the world)_

@velo, If you have Feign client signature returning `Optional`, and you do **not** translate 404 to `empty`, how `empty` can appear in your application? Do you transform blank JSON `{}` to `empty`?

arielsegura:
There's a decoder for this already
https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/optionals/OptionalDecoder.java

michaldo:
> There's a decoder for this already
> https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/optionals/OptionalDecoder.java

My solution use this decoder. The decoder is quite good, the problem is that is hard to use it to handle 404:

- if decode404=false, `OptionalDecoder` is skipped
- if decode404=true, `OptionalDecoder` is used, but side effect is that if Optional is **NOT** used, 404 do not cause Feign exception -  in most cases exception is more handy than manual null check

michaldo:
If I understood correctly, your doctrine is 
`204 + Optional = empty`

This doctrine is consistent. Considering that it is not acceptable the same time have `404=empty` and `204=empty`, a one solution must be  selected.

The biggest problem with approach `404=empty` is that 404 may also mean wrong URL. Comparing, 204 always means URL is correct. Second big problem is consistency: as the same time `404` and `204` cannot map to `empty`. For `404` Feign exception can be thrown, but for `204` there is no alternative.

The biggest problem with approach `204=empty` is limited usablity: it is effective only for URL which sometimes return 200 and sometimes 204 - rare case IMHO.

From application developer perspective, `204=empty` is useless, comparing to `404=empty`

What do you think about property `decodeOptional404`, which would be like `decode404` but limited to Optional class only?

michaldo:
> This change breaks the encapsulation afforded by the `Decoder` separation. 

I you refer to 
```
} else if (response.status() == 404 && OptionalDecoder.isOptional(metadata.returnType())) {
```
I do not agree. Return type check already exists in code
```
if (void.class == metadata.returnType()) {
```

kdavisk6:
The key point here is your original assumptions that developers who choose `decode404` are unaware of the consequences.  Your change enforces a singular view on how `Optional` should behave.   @velo and I agree that this change is not appropriate.

However, I do recognize that how `decode404` interacts with `Optional` is confusing and should be revisited.  One alternative is to enable `decode404` at the method level instead of the client level.  This would allow users to decide which methods would get the behavior or not, resulting in something like the following:

```java
interface UserService {
   @RequestLine("GET /search{?query+}", decode404=true)
   Optional<User> search(Query query);
  
   @RequestLine("GET /user/{user}")
   User findById(UUID user);
}
```

This balances your request without requiring a global decision.  Thoughts?

michaldo:
I was thinking a lot about the issue. 

My case is user login - user enters a login and system check the user exists or not. REST endpoint return 0 or 1 entity. Code 404 is used to indicate 0 entity and my purpose was connect 404 with Optional.empty

But now I think better way is use code 204 to indicate 0 entity. 204 is mapped to Optional.empty automatically by Feign.

Conclusion:
If REST endpoint returns 0 or 1 entity and both answers are correct from business perspective, HTTP code 204 should be used to indicate "0 entity case"

However, if for some reason 404 means "0 entity case", it is good idea to have option to customize particular endpoint by parameter `decode404=true`. 

I agree with your doctrine a I'm closing this PR.

PR #1117, opened 2019-11-19
Makes iterator compatible with Java iterator expected behavior

ItamarBenjamin:
both next() and hasNext() should read from stream if needed. both also inspect 'current' member, next() resets it after consuming. exception is thrown when no more elements are available to return.


ItamarBenjamin:
@velo CI is fixed

PR #1118, opened 2019-11-20
added a method to get a target name from RequestTemplate

joongs4:
Hello, I'm now in a huge MSA project and need a method like below.

In our MSA project, I need to specify the target instance of a microservice by using EurekaClient. To do it, I should figure out what the target microservice name is. This is the reason why I added a method returning target service name.

Thanks.


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -966,6 +966,15 @@ public MethodMetadata methodMetadata() {
   public Target<?> feignTarget() {
```
velo: Why not use this method instead?

PR #1130, opened 2019-12-12
Force followRedirects on the OkHttpClient when needed

fpavageau:
This is a small fix: `followRedirects` was only applied when combined with a connect or read timeout override, which the test did.


kdavisk6:
Looks good.  Travis is unhappy due to formatting being off.  Can you please fix the formatting?

fpavageau:
@kdavisk6 Sure, it's fixed. Too bad the formatting check is attached to the `install` phase, as I'd only run `package`.

PR #1145, opened 2019-12-27
Remove deprecated code usage from core module

### LINKED ISSUE #857: Remove use of deprecated internal APIs from sub modules

kdavisk6:
Fixes #857

To simply removal, Request.Body was returned back to an internal
component and additional methods were added to Request to expose
it's capabilities outside of the object.

All other deprecated usage in core modules has been removed.
Deprecated code still exists in the test cases and will be
removed once the deprecated methods are removed in our next
major release.

This is a very large PR that has very little changes.

### DIFF
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -13,83 +13,20 @@
  */
 package feign;
 
-import static feign.Util.checkNotNull;
-import static feign.Util.valuesOrEmpty;
 import java.net.HttpURLConnection;
 import java.nio.charset.Charset;
-import java.util.*;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-import feign.template.BodyTemplate;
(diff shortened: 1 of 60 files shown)


core/src/main/java/feign/Request.java
```diff
@@ -232,17 +184,29 @@ public Charset charset() {
...
+    return body.isBinary();
+  }
+
+  /**
+   * Request Length.
+   *
+   * @return size of the request body.
+   */
+  public int length() {
+    return this.body.length();
```
adrian-skybaker, line 202: What if body is null?

Issue #1153, opened 2020-01-09
Update okhttp from 3.6.0 (due to security issues)


svenhaag:
Currently okhttp 3.6.0 is used which is from January 2017. Please update to a newer version as it has some serious security issues.

See:
[Maven Central](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp)
and
[CVE](https://nvd.nist.gov/vuln/detail/CVE-2018-20200)

kdavisk6:
Understood, however, users are free to override our default versions as they please.

Please feel free to open a PR with this dependency change if you like.  Otherwise, I'll add this to our like of small changes and add it to 10.8

kdavisk6:
Closed by #1158

PR #1158, opened 2020-01-14
Update okhttp client version from 3.6.0 to 3.14.4.

svenhaag:
PR for #1153 

Updating only the okhttp client version, not the mockwebserver version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,8 @@
 
     <main.basedir>${project.basedir}</main.basedir>
 
-    <okhttp3.version>3.6.0</okhttp3.version>
+    <okhttp3.client.version>3.14.4</okhttp3.client.version>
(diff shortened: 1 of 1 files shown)


kdavisk6:
@svenhaag 

I looks like the pom file contains `CLRF` values.  Can you please update this PR normalizing the line endings to `LF`?

svenhaag:
> 
> 
> @svenhaag
> 
> I looks like the pom file contains `CLRF` values. Can you please update this PR normalizing the line endings to `LF`?

Line endings seem all fine to me (checked in repo dierctly). However I had a tab instead of spaces in front of mockserver version line. Could this the build causing to fail?

kdavisk6:
Yeah, seems so.

PR #1160, opened 2020-01-17
Correct Encoding and restore decodeSlash in QueryTemplate

### LINKED ISSUE #1156: Query params will be encoded twice in newest Version 10.7.2
After the release of version 10.7.2, values of query params in the expression will be encoded twice. 
Following URL will be generated when I provide the string `:§$%&/()=` as parameter value:
`https://www.abc.com/foo?bar=%253A%25C2%25A7%2524%2525%2526%2F%2528%2529%253D`
When I decode the URL I get following result:
`https://www.abc.com/foo?bar=%3A%C2%A7%24%25%26/%28%29%3D`
And exactly this result shows a URL that is encoded, so therefore when I decode the result again, I get following string:
`https://www.abc.com/foo?bar=:§$%&/()=`

Unfortunatly, changing the HTTP client did not help. Am I the only one who got this problem?

kdavisk6:
Fixes #1156

Collection Format was encoding query string values unnecessarily
due to changes introduced in #1138 and #1139 that encode template
values before appending them to the query string.

In addition, `decodeSlash` flags that were accidentally removed,
have been restored in QueryTemplate.


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -21,6 +21,8 @@
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
```
velo: Any particular reason for this imports?

core/src/main/java/feign/template/UriUtils.java
```diff
@@ -153,16 +154,21 @@ private static String encodeChunk(String value, Charset charset, boolean allowRe
...
+      for (byte b : data) {
+        if (isUnreserved((char) b)) {
+          bos.write(b);
+        } else if (isReserved((char) b) && allowReserved) {
+          bos.write(b);
+        } else {
+          pctEncode(b, bos);
+        }
       }
+      return new String(bos.toByteArray());
```
velo: Should we include a charset when creating the string?

core/src/main/java/feign/RequestTemplate.java
kdavisk6: I removed all unused imports.

core/src/main/java/feign/template/UriUtils.java
kdavisk6: Sure thing.  Updated.

PR #1165, opened 2020-01-28
Use LinkedHashMap for deterministic iterations

contextshuffling:
Test `shouldPrintHeaders()` asserts the string version of `RequestHeader`. `RequestHeader` class uses a HashMap to store the headers. However, HashMap does not guarantee any specific order of entries. Test can fail if the iteration order changes.
This PR proposes to fix the test by changing HashMap to LinkedHashMap that makes the iteration order deterministic.

### DIFF
--- a/mock/src/main/java/feign/mock/RequestHeaders.java
+++ b/mock/src/main/java/feign/mock/RequestHeaders.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2019 The Feign Authors
+ * Copyright 2012-2020 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


kortov:
I guess to change implementation for tests is not a good approach. `HashMap` is lighter than `LinkedHashMap` ofc. I guess there should be a better approach to test it though I agree on a deterministic approach

contextshuffling:
@kortov Thanks for the feedback. I guess another approach is to assert for both orders. It gets messy when there are more but in this test, there are only two possible cases.

kortov:
Maybe you're right with your implementation because in https://github.com/OpenFeign/feign/blob/17735a357ecf0efbb655bd7b50b0344bf861c51f/core/src/main/java/feign/Response.java#L40 binded `LinkedHashMap` . And more than that, your PR is to feign.mock  package, so I guess performance is not bigger issue than determenistic results.

contextshuffling:
@velo  I fixed the format issue.

PR #1174, opened 2020-02-17
issue 361 - async feign variant supporting CompleteableFutures

motinis:
Adds CompleteableFuture support.
The targeted API must be an interface, where each method has a return type of CompleteableFuture<T> for some type T which can be decoded by the configured decoder.

Context is explicit, providing support for sessions.
A default synchronous -> asynchronous client is provided, although the different modules should add their own implementations of AsyncClient.

Retry is not supported - in particular, there can be many different ways one could handle the threading (scheduling the try on an appropriate thread). resilience4j could be an option, in particular to provide a wrapper to AsyncClient.

### DIFF
--- a/core/src/main/java/feign/AsyncClient.java
+++ b/core/src/main/java/feign/AsyncClient.java
@@ -68,4 +68,28 @@ public interface AsyncClient<C> {
     }
   }
 
+  /**
+   * A synchronous implementation of {@link AsyncClient}
+   * @param <C> - unused context; synchronous clients handle context internally
+   */
+  class Pseudo<C> implements AsyncClient<C> {
+
+    private final Client client;
+
+    public Pseudo(Client client) {
+      this.client = client;
+    }
+
+    @Override
+    public CompletableFuture<Response> execute(Request request, Options options, Optional<C> requestContext) {
+      CompletableFuture<Response> result = new CompletableFuture<>();
+      try {
+        result.complete(client.execute(request, options));
+      }catch(Exception e) {
+        result.completeExceptionally(e);
+      }
+
+      return result;
+    }
+  }
 }
--- a/core/src/main/java/feign/AsyncClient.java
+++ b/core/src/main/java/feign/AsyncClient.java
@@ -1,5 +1,15 @@
 /**
- * 
+ * Copyright 2012-2020 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
  */
 package feign;
 
@@ -7,7 +17,6 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-
 import feign.Request.Options;
 
 /**
@@ -15,45 +24,48 @@ import feign.Request.Options;
  */
 public interface AsyncClient<C> {
 
-	/**
-	 * Executes the request asynchronously. Calling {@link CompletableFuture#cancel(boolean)} on the result
-	 * may cause the execution to be cancelled / aborted, but this is not guaranteed.
-	 * 
-	 * @param request safe to replay
-	 * @param options options to apply to this request
-	 * @param contextOpt - the optional context. The client should update this appropriately based
-	 * on the received response before completing the result.
-	 * @return a {@link CompletableFuture} to be completed with the response, or completed exceptionally
-	 * otherwise, for example with an {@link java.io.IOException} on a network error connecting to {@link Request#url()}.
-	 */
-	CompletableFuture<Response> execute(Request request, Options options, Optional<C> contextOpt);
+  /**
+   * Executes the request asynchronously. Calling {@link CompletableFuture#cancel(boolean)} on the
+   * result may cause the execution to be cancelled / aborted, but this is not guaranteed.
+   * 
+   * @param request safe to replay
+   * @param options options to apply to this request
+   * @param requestContext - the optional context, for example for storing session cookies. The client should update
+   *                       this appropriately based on the received response before completing the result.
+   * @return a {@link CompletableFuture} to be completed with the response, or completed
+   *         exceptionally otherwise, for example with an {@link java.io.IOException} on a network
+   *         error connecting to {@link Request#url()}.
+   */
+  CompletableFuture<Response> execute(Request request, Options options, Optional<C> requestContext);
 
-	class Default<C> implements AsyncClient<C> {
-		
-		private final Client client;
-		private final ExecutorService executorService;
-		
-		public Default(Client client, ExecutorService executorService) {
-			this.client = client;
-			this.executorService = executorService;
-		}
+  class Default<C> implements AsyncClient<C> {
 
-		@Override
-		public CompletableFuture<Response> execute(Request request, Options options, Optional<C> contextOpt) {
-			CompletableFuture<Response> result = new CompletableFuture<>();
-			Future<?> future = executorService.submit(() -> {
-				try {
-					result.complete(client.execute(request, options));
-				}catch(Exception e) {
-					result.completeExceptionally(e);
-				}
-			});
-			result.whenComplete( (response, throwable) -> {
-				if (result.isCancelled())
-					future.cancel(true);
-			});
-			return result;
-		}
-	}
-	
+    private final Client client;
+    private final ExecutorService executorService;
+
+    public Default(Client client, ExecutorService executorService) {
+      this.client = client;
+      this.executorService = executorService;
+    }
+
+    @Override
+    public CompletableFuture<Response> execute(Request request,
+                                               Options options,
+                                               Optional<C> requestContext) {
+      CompletableFuture<Response> result = new CompletableFuture<>();
+      Future<?> future = executorService.submit(() -> {
+        try {
+          result.complete(client.execute(request, options));
+        } catch (Exception e) {
+          result.completeExceptionally(e);
+        }
+      });
+      result.whenComplete((response, throwable) -> {
+        if (result.isCancelled())
+          future.cancel(true);
+      });
+      return result;
+    }
+  }
+
 }
--- a/core/src/main/java/feign/AsyncClient.java
+++ b/core/src/main/java/feign/AsyncClient.java
@@ -1,0 +1,59 @@
+/**
+ * 
+ */
+package feign;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+import feign.Request.Options;
+
+/**
+ * Submits HTTP {@link Request requests} asynchronously, with an optional context.
+ */
+public interface AsyncClient<C> {
+
+	/**
+	 * Executes the request asynchronously. Calling {@link CompletableFuture#cancel(boolean)} on the result
+	 * may cause the execution to be cancelled / aborted, but this is not guaranteed.
+	 * 
+	 * @param request safe to replay
+	 * @param options options to apply to this request
+	 * @param contextOpt - the optional context. The client should update this appropriately based
+	 * on the received response before completing the result.
+	 * @return a {@link CompletableFuture} to be completed with the response, or completed exceptionally
+	 * otherwise, for example with an {@link java.io.IOException} on a network error connecting to {@link Request#url()}.
+	 */
+	CompletableFuture<Response> execute(Request request, Options options, Optional<C> contextOpt);
+
+	class Default<C> implements AsyncClient<C> {
+		
+		private final Client client;
+		private final ExecutorService executorService;
+		
+		public Default(Client client, ExecutorService executorService) {
+			this.client = client;
+			this.executorService = executorService;
+		}
+
+		@Override
+		public CompletableFuture<Response> execute(Request request, Options options, Optional<C> contextOpt) {
+			CompletableFuture<Response> result = new CompletableFuture<>();
+			Future<?> future = executorService.submit(() -> {
+				try {
+					result.complete(client.execute(request, options));
+				}catch(Exception e) {
+					result.completeExceptionally(e);
+				}
+			});
+			result.whenComplete( (response, throwable) -> {
+				if (result.isCancelled())
+					future.cancel(true);
+			});
+			return result;
+		}
+	}
+	
+}
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -1,0 +1,328 @@
+
+package feign;
+
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import feign.Logger.NoOpLogger;
+import feign.Request.Options;
+import feign.Target.HardCodedTarget;
+import feign.codec.Decoder;
+import feign.codec.Encoder;
+import feign.codec.ErrorDecoder;
+
+/**
+ * Enhances {@link Feign} to provide support for asynchronous clients.
+ * Context (for example for session cookies or tokens) is explicit,
+ * as calls for the same session may be done across several threads.
+ * <br><br>
+ * {@link Retryer} is not supported in this model, as that is a blocking API.
+ * {@link ExceptionPropagationPolicy} is made redundant as
+ * {@link RetryableException} is never thrown.
+ * <br>
+ * Alternative approaches to retrying can be handled through {@link AsyncClient clients}.
+ * <br><br>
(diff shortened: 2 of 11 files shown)


core/src/test/java/feign/AsyncFeignTest.java
```diff
@@ -0,0 +1,1067 @@
...
+	}
+	
+	static interface UpperWildApi {
+		@RequestLine("GET /")
+		CompletableFuture<? super Object> x();
+	}
+	
+	static interface WildApi {
+		@RequestLine("GET /")
+		CompletableFuture<? extends Object> x();
```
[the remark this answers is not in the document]
motinis: You're right - I missed that. It should be:

```suggestion
		CompletableFuture<?> x();
```

core/src/test/java/feign/AsyncFeignTest.java
```diff
@@ -0,0 +1,1067 @@
...
+		checkCFCompletedSoon(cf);
+	}
+
+	interface TestInterfaceAsync {
+
+		@RequestLine("POST /")
+		CompletableFuture<Response> response();
+
+		@RequestLine("POST /")
+		CompletableFuture<String> post() throws TestInterfaceException;
```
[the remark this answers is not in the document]
motinis: It is possible, and easy to implement. The proxy ReflectiveAsyncFeign would need to detect these kinds of methods and use CompleteableFuture.join() in a try/catch block with appropriate exception handling.

I though would exercise some caution here.

1. I think that interfaces should either be asynchronous or synchronous - I really can't think of a case where it makes sense to mix between the two.
2. If this is to be used as a replacement for the synchronous approach today, then I think it does require adding in a "blessed" way for doing retries (as exists today). Note though that it would break existing integrations, which would need to implement AsyncClient, as well as client code, in particular because context is now explicit. One could get around this by having the client also implement the context supplier (and handle under the covers in the builder by checking if the client implements Supplier<C>). This isn't too bad, since the synchronous clients today handle this internally generally on a thread local anyway.

core/src/main/java/feign/AsyncClient.java
```diff
@@ -0,0 +1,59 @@
...
+	 * may cause the execution to be cancelled / aborted, but this is not guaranteed.
+	 * 
+	 * @param request safe to replay
+	 * @param options options to apply to this request
+	 * @param contextOpt - the optional context. The client should update this appropriately based
+	 * on the received response before completing the result.
+	 * @return a {@link CompletableFuture} to be completed with the response, or completed exceptionally
+	 * otherwise, for example with an {@link java.io.IOException} on a network error connecting to {@link Request#url()}.
+	 */
+	CompletableFuture<Response> execute(Request request, Options options, Optional<C> contextOpt);
```
kdavisk6: `contextOpt` is a confusing and ambiguous to me.  What is this context for?

core/src/main/java/feign/AsyncFeign.java
```diff
@@ -0,0 +1,328 @@
...
+ */
+public abstract class AsyncFeign<C> extends Feign {
+
+	public static <C> AsyncBuilder<C> asyncBuilder() {
+		return new AsyncBuilder<>();
+	}
+
+	private static class LazyInitializedExecutorService {
+
+		private static final ExecutorService instance = Executors.newCachedThreadPool(r -> {
```
kdavisk6: Cached Thread Pools are OK to start, but I think we may be better off with a Fixed Thread Pool.

core/src/main/java/feign/AsyncFeign.java
```diff
@@ -0,0 +1,328 @@
...
+		
+		public <T> T target(Target<T> target) {
+			return build().newInstance(target);
+		}
+
+		public <T> T target(Target<T> target, C context) {
+			return build().newInstance(target, context);
+		}
+		
+		private static <C> AsyncClient<C> initDefaultClient() {
```
kdavisk6: Using a Builder, I don't see why we would need to have a static method to do this.  Methods like this are fine to be in the `build` method, given that it's being called locally.

core/src/main/java/feign/AsyncResponseHandler.java
```diff
@@ -0,0 +1,102 @@
...
+		super();
+		this.logLevel = logLevel;
+		this.logger = logger;
+		this.decoder = decoder;
+		this.errorDecoder = errorDecoder;
+		this.decode404 = decode404;
+		this.closeAfterDecode = closeAfterDecode;
+	}
+	
+	void handleResponse(CompletableFuture<Object> resultFuture, AsyncInvocation<?> invocationContext,
```
kdavisk6: There is a large amount of duplication here from the regular method handler.  We should consider refactoring this logic into a component that can be used by both execution modes.

core/src/main/java/feign/ReflectiveAsyncFeign.java
```diff
@@ -0,0 +1,126 @@
...
+import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+
+
+public class ReflectiveAsyncFeign<C> extends AsyncFeign<C> {
+
+	private static class MethodInfo {
```
kdavisk6: `MethodMetadata` already exists to serve this purpose.  We should consider refactoring that component before creating additional components.

core/src/test/java/feign/AsyncFeignTest.java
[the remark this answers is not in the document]
kdavisk6: > I think that interfaces should either be asynchronous or synchronous - I really can't think of a case where it makes sense to mix between the two.

You may not, but our user's may, so we should consider it.

See my other comment on Retry.  We will have to break it.

core/src/main/java/feign/AsyncClient.java
motinis: Generally a request happens within some context - for example, there may be session cookies or headers that need to be sent and maintained as part of the calls.

core/src/main/java/feign/AsyncFeign.java
motinis: I chose a CachedThreadPool to avoid needing to define additional configuration. Users can of course elect to pass in their own executor here for converting a sync client for use in async code (although they may have issues there with the fact that context may be shared in strange ways).

core/src/main/java/feign/AsyncFeign.java
motinis: This could be inlined into the lazyInits() method right now, as long as the number of things going on there is small, it's probably clearer inlined.

core/src/main/java/feign/AsyncResponseHandler.java
motinis: Yes, there is. I first wanted to bring this in cleanly for review without doing major changes to unify everything.

core/src/main/java/feign/AsyncClient.java
kdavisk6: Thanks for the explanation, can we rename it to `requestContext` to be more explicit?

core/src/main/java/feign/ReflectiveAsyncFeign.java
motinis: Agreed - it was doing much too much for my needs, hence the duplication. Much preferable to refactor.

core/src/test/java/feign/AsyncFeignTest.java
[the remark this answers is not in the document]
motinis: There is really only one reason that I can see to want a mixed interface and that's when wants to extend an existing synchronous interface to add on new async features (and one wants to be able to pass the instance to existing code).

Otherwise, it's generally trivial to do a join on the async call. The other way, changing client code from sync to async would likely take additional effort. But you're right that users could potentially have other constraints or reasons they need things a certain way.

However, it's not difficult to support - best practices can certainly be captured in documentation.

motinis:
Thank you for the positive feedback, and yes, I think a branch to work through this is a great idea. There are several aspects of Feign internals that I'm not familiar enough with, so please be patient if some questions look "too basic".


> Here's what I think we can do
> 
> * [ ]  Refactor `MethodMetadata` to be thread-safe
> * [ ]  Refactor `handleResponse` to be independent functions
> * [ ]  Refactor `Contract` to choose the right `InvocationHandler` based on return type.
> 

- In terms of making MethodMetadata threadsafe - I have a feeling rather that this class should be split in two. One part should be completely immutable - it could potentially be created when `target` is called (alternatively it could be created at call time but cached). The second part could be created per request  and would be mutable as part of the processing, but then I don't think it would need to be thread-safe. First part could still be called `MethodMetadata` or be renamed, second part I would think as `RequestMethodMetadata`
- For `handleResponse` to be "independent" - could you please explain here. Do you mean having different method signatures for sync vs async?

core/src/test/java/feign/AsyncFeignTest.java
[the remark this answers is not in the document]
velo: > I think that interfaces should either be asynchronous or synchronous - I really can't think of a case where it makes sense to mix between the two.

My point at supporting both at same time, would be mainly to force a single workflow inside feign.

I don't really expect anyone using async for 3 methods and having one not async.  Maybe someone will.

velo:
>     * n terms of making MethodMetadata threadsafe - I have a feeling rather that this class should be split in two. One part should be completely immutable - it could potentially be created when `target` is called (alternatively it could be created at call time but cached). The second part could be created per request  and would be mutable as part of the processing, but then I don't think it would need to be thread-safe. First part could still be called `MethodMetadata` or be renamed, second part I would think as `RequestMethodMetadata`

Agree. I poke around it recently and it has mutable and immutable data in a single class.  Moving all mutable data to a new `RequestMethodMetadata` makes lot's of sense.

I guess one side effect of that, would be us cutting a feign 11 release =)

motinis:
I've now updated the commits to take into account some of the comments - in particular, AsyncResponseHandler is used internally by SynchronousMethodHandler for processing of the response.

Next up will be expanding Async to support synchronous methods (as I said, this should be pretty easy) - for testing, the tests in FeignTest not dealing with Retry can be used, just going through AsyncFeign instead.

Finally would be looking to take MethodInfo and pull it out into first class citizen and then having MethodMetadata reference it, or extend it as the case may be.

I'm trying to get this pretty clean for all the non-retry before looking to do major rewriting.

core/src/main/java/feign/AsyncClient.java
motinis: Done in latest commit

core/src/main/java/feign/AsyncFeign.java
motinis: Inlined now

core/src/main/java/feign/AsyncResponseHandler.java
motinis: Refactored.

motinis:
I added now sync through async. One thing that was necessary as part of that was coercing the SynchronousMethodHandler to explictly call the decoder (which is really the stageDecode in AsyncFeign) - this is because there were some edge cases in terms of Response/void return types which have correct special handling in AsyncResponseHandler which would be quite ugly to handle otherwise. I thought this way would be the most robust, absent a more extensive rewrite.

core/src/test/java/feign/AsyncFeignTest.java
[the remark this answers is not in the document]
motinis: I added now sync through async. One thing that was necessary as part of that was coercing the SynchronousMethodHandler to explictly call the decoder (which is really the stageDecode in AsyncFeign) - this is because there were some edge cases in terms of Response and void return types which have correct special handling in AsyncResponseHandler which would be quite ugly to handle otherwise. I thought this way would be the most robust, absent a more extensive rewrite. To keep things consistent and simple, the coercion is also used for the async case, even though it works without it.

velo:
BTW, is there any tasks you would like coding assistance with?

Feel free to throw anything you need at me, I can make PRs to your branch to help out

core/src/main/java/feign/AsyncFeign.java
```diff
@@ -0,0 +1,335 @@
...
+      result.setDaemon(true);
+      return result;
+    });
+  }
+
+  public static class AsyncBuilder<C> {
+
+    private final Builder builder;
+    private Supplier<C> defaultContextSupplier = () -> null;
+    private AsyncClient<C> client;
```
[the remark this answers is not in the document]
motinis: Sure - that could work - although there would need to be enforcement that then you're not using a context supplier (i.e., the client would always be called with an empty requestContext Optional)

motinis:
> BTW, is there any tasks you would like coding assistance with?
> 
> Feel free to throw anything you need at me, I can make PRs to your branch to help out

Originally, I had this added as a "clean" extension - it leveraged standard Feign internally, but didn't make any changes. Since we're unifying the sync path onto this, it does require changes, and so we should do some cleanup here which I'd appreciate assistance with. One thing is I wouldn't want to break any existing code using the Feign class right now. What I suggest is to do this in 2 steps:

1) Instead of the forceDecoding option, have the ability to provide a MethodHandlerFactory as an internal part of Feign.Builder. This way, the Async codepath can be cleaned up from the stagedExecute and stagedDecode (these should be combined into a single method that's basically the AsyncMethodHandler implementation).
2) Refactor out a FeignDispatcher part - this should cover from proxy through to MethodMetadata creation at minimum, and probably including some more generic hook to the dispatch via a MethodHandler. AsyncFeign would then use that internally instead of Feign. Feign too should leverage FeignDispatcher (although whether via inheritance or other means is unclear). FeignDispatcher could possibly be public API, resulting in a separation of dispatch from handling.

core/src/main/java/feign/AsyncFeign.java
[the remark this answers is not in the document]
motinis: I delivered AysncClient.Pseudo instead as the fully synchronous AsyncClient implementation.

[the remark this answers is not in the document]
velo: Fair point, that is an excellent delivery. We can expand on that later

velo:
> Originally, I had this added as a "clean" extension

You are absolutely right.
Let's get this in, as is (or at least w/o Wishlist) and cut a release.  Put it out there in front of people =)

I made a few minor changes to your code that should make travis CI happy
https://github.com/motinis/feign/pull/1

The only think missing is some instructions on readme =)

motinis:
> I made a few minor changes to your code that should make travis CI happy
> [motinis#1](https://github.com/motinis/feign/pull/1)

Thanks, I've merged those in.

velo:
There are still some formatting problems on `core/src/main/java/feign/ReflectiveAsyncFeign.java` running maven locally should fix it.

motinis:
> There are still some formatting problems on `core/src/main/java/feign/ReflectiveAsyncFeign.java` running maven locally should fix it.

Can you please do that and I'll merge? Getting an error on my local where it can't find pomSortOrder.xml in the classpath.

velo:
Could you please run maven with `-e` and post the exception?

motinis:
> Could you please run maven with `-e` and post the exception?

I seem to be able to work-around this (by running from the feign parent dir with -f to core/pom.xml) - but the change it made to this file seems odd, in that it moved the opening brace to its own indented line:

```
      if (!CompletableFuture.class.isAssignableFrom(retType))
       {
        continue; // synchronous case
      }
```
If that's what you want, then sure I can push that...

velo:
Odd, but let's see if that fixes the build

motinis:
Looks like it did fix the build

velo:
@kdavisk6 I'm happy to merge this as is and start working on feign 11 with

- [ ] Refactor `MethodMetadata` to be thread-safe
- [ ] Refactor `handleResponse` to be independent functions
- [ ] Refactor `Contract` to choose the right `InvocationHandler` based on return type.

What do you think?

kdavisk6:
I’ll take a look tonight, but I trust you. Let’s create a new 11.x branch
after this is merged and refactor there.

velo:
Ok, once merged, I think we should release another 10.x

motinis:
> @kdavisk6 I'm happy to merge this as is and start working on feign 11 with
> 
> * [ ]  Refactor `MethodMetadata` to be thread-safe
> * [ ]  Refactor `handleResponse` to be independent functions
> * [ ]  Refactor `Contract` to choose the right `InvocationHandler` based on return type.
> 
> What do you think?

I think the actual refactoring should look different... basically, the current method handler approach should be used. This depends then if you want to remove the current SynchronousMethodHandler or not.
Basically, there should be an AsyncMethodHandler and corresponding factory that gets used. The client, context supplier, etc. would get pushed down into the factory to pass over to the handler. There may need to be some work on fixing up MethodMetadata, but otherwise the determining of the method call mode (sync vs async) and underlying return type for async could be done in the method handler.
If one wants to keep around the existing SynchronousMethodHandler, I would basically suggest having the builder take a supplier for the factory (so like a factory factory), so the creation can be delayed until build is actually taking place. Then the default could be to create the existing Factory, while it could be overridden by async builder to create the async factory version.
Probably needs some diagrams, etc. and should be started in a new issue and branch :)

kdavisk6:
@motinis and @velo I've approved this PR.  Feel free to move forward with this.  I'll create a new issue for us to discuss how to move forward with Feign 11

velo:
cool, I merged as is. Will work on releasing later today (social life permitting =] )

I will be sending a PR that added async capabilities to apache http 5 integration.

@motinis lemme know if you would be interested in reviewing that one.

nallabheem:
Hi @velo, @motinis  , is Async is available for Feign-Retry now. We are currently using OpenFeign3.0.0 and If we want to use the version of OpenFeign which support Async for Feign-Retry which version is that ? and what are all the changes between version 3.00 and 11 .

Could you please help me with the info. 
Thanks in advance for the help and support.

velo:
Hi @nallabheem , feign 3 is very, very, very, veeery old.

I can't start listing what would be the impact of such change. Most relevant thing that come to mind is that we dropped java 6 support.  I would suggest you go ahead and try it.

PR #1179, opened 2020-02-27
Async client for apache http 5

velo:
@motinis in case you wanna see what I did based on your work

### DIFF
--- a/hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
+++ b/hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
@@ -1,0 +1,196 @@
+/**
+ * Copyright 2012-2020 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package feign.hc5;
+
+import org.apache.hc.client5.http.async.methods.SimpleHttpRequest;
+import org.apache.hc.client5.http.async.methods.SimpleHttpResponse;
+import org.apache.hc.client5.http.config.Configurable;
+import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
+import org.apache.hc.client5.http.protocol.HttpClientContext;
+import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.io.CloseMode;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import feign.*;
+import feign.Request.Options;
+
+/**
+ * This module directs Feign's http requests to Apache's
+ * <a href="https://hc.apache.org/httpcomponents-client-5.0.x/index.html">HttpClient 5</a>. Ex.
+ *
+ * <pre>
+ * GitHub github = Feign.builder().client(new ApacheHttp5Client()).target(GitHub.class,
+ * "https://api.github.com");
+ */
+/*
+ */
+public final class AsyncApacheHttp5Client implements AsyncClient<HttpClientContext>, AutoCloseable {
+
+  private static final String ACCEPT_HEADER_NAME = "Accept";
+
+  private final CloseableHttpAsyncClient client;
+
+  public AsyncApacheHttp5Client() {
+    this(createStartedClient());
+  }
+
+  private static CloseableHttpAsyncClient createStartedClient() {
+    final CloseableHttpAsyncClient client = HttpAsyncClients.custom().build();
+    client.start();
+    return client;
+  }
+
+  public AsyncApacheHttp5Client(CloseableHttpAsyncClient client) {
+    this.client = client;
+  }
+
+  @Override
+  public CompletableFuture<Response> execute(Request request,
+                                             Options options,
+                                             Optional<HttpClientContext> requestContext) {
+    final SimpleHttpRequest httpUriRequest = toClassicHttpRequest(request, options);
+
+    final CompletableFuture<Response> result = new CompletableFuture<>();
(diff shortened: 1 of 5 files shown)


hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -0,0 +1,183 @@
...
+ * This module directs Feign's http requests to Apache's
+ * <a href="https://hc.apache.org/httpcomponents-client-5.0.x/index.html">HttpClient 5</a>. Ex.
+ *
+ * <pre>
+ * GitHub github = Feign.builder().client(new ApacheHttp5Client()).target(GitHub.class,
+ * "https://api.github.com");
+ */
+/*
+ */
+public final class AsyncApacheHttp5Client<C> implements AsyncClient<C> {
```
motinis: I think here this should actually be:
```suggestion
public final class AsyncApacheHttp5Client implements AsyncClient<HttpClientContext> {
```
This then enables the user to create their own external context as part of AsyncFeign - thus, they can provide their own CookieStore for example and have session management.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -0,0 +1,183 @@
...
+        result.cancel(false);
+      }
+    };
+
+    client.execute(httpUriRequest, callback);
+
+    return result;
+  }
+
+  protected HttpClientContext configureTimeouts(Request.Options options) {
```
motinis:
```suggestion
  protected HttpClientContext configureTimeouts(Request.Options options, Optional<HttpClientContext> requestContext) {
```

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -0,0 +1,183 @@
...
+    final HttpClientContext context = new HttpClientContext();
```
motinis:
```suggestion
    final HttpClientContext context = requestContext.orElseGet(HttpClientContext::new)
```
Use the client supplied context, else make a new default just for this request.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -0,0 +1,183 @@
...
+
+  private final CloseableHttpAsyncClient client;
+
+  public AsyncApacheHttp5Client() {
+    this(HttpAsyncClients.custom().build());
+  }
+
+  public AsyncApacheHttp5Client(CloseableHttpAsyncClient client) {
+    this.client = client;
+    client.start();
```
motinis: Doing client.start() here is a bit of an issue, because then Feign has responsibility for the lifecycle of the client (if it's doing start() it then needs to do close()). As well, the user may want to use that client in other places. So I would recommend moving that responsibility to the user.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -0,0 +1,183 @@
...
+        result.completeExceptionally(ex);
+      }
+
+      @Override
+      public void cancelled() {
+        result.cancel(false);
+      }
+    };
+
+    client.execute(httpUriRequest, callback);
```
motinis:
```suggestion
    client.execute(httpUriRequest, configureTimeouts(options, requestContext), callback);
```
use the HttpClientContext here of course :)

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
kdavisk6: I'm not convinced we need to add another construct like the `Context` when we have `Request` already, which represents the canonical information required for a `Client` to execute a request.  It's already thread-safe and immutable.  We should consider extending `Request` before adding another concept.

Another counter point, this `Context` provides users a way to circumvent our internal separation and, if history is any indication, will become a cause for confusion.  If you decide to continue to use this `Context`, I strongly suggest that we make it part of our private API usage.

Cookies and Session management can be something that we should support first at the `Contract` level, ensuring compatibility throughout the ecosystem and preserve the domain separation Feign is good at, which we should address separately from this PR.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
kdavisk6: This is typically how our internal APIs are constructed.  Feign controls the lifecycle

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
kdavisk6: `Optional` as a method parameter is an [anti-pattern](http://dolszewski.com/java/java-8-optional-use-cases/).  A more appropriate use would be to have method overloads

```java
protected HttpClientContext configureTimeouts(Request.Options options) {
 ...
}

protected HttpClientContext configureTimeouts(Request.Options options, HttpClientContext context) {
  ....
}
```

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
velo: Well, I think what I will do is only start the client when no args constructor is invoked.

So, when feign creates the http client it will also start it. If you are using the overwritten constructor, you manage the lifecycle of `CloseableHttpAsyncClient`.

Also, will make this client closeable, to make possible for people using the no-args constructor to close it.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
velo: I will resolve the optional on the execute method.  Will only include `configureTimeouts(Request.Options options, HttpClientContext context)`

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
velo: For now I will keep the Context, but I agree that is something we need to revisit

velo:
> @velo You have some changes here that aren't about the HC5 client, can we limit this PR to just the HC5 changes?

My bad, IDE automatically fixed stuff.

Only change required on feign core was make a few classes public so I could run the same tests on hc5

motinis:
I think that putting context onto the Request itself is preferable to
having an extra parameter.

One could indeed to this via Contract; and indeed the common cases could be
supported with explicit enums in annotations. I think as well that these
common cases should likely be fulfilled by the async client class. I also
think that there does need to be an ability for the user to specify via
Contract or Builder an explicit context provider for use-cases which
wouldn't make sense to put into Feign.

hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
kdavisk6: agreed.

PR #1181, opened 2020-02-28
Metric API

velo:
Provide a first-class Metrics API that users can tap into to gain insight into the request/response lifecycle. 

My plan is:
Extend the capabilities to Async and Reactive (before 10.8 release)
Add metrics support for OpenTracing

After this, I feel we are in a good place to cut 10.8 and being 11 branch

### DIFF
--- a/core/src/main/java/feign/Capability.java
+++ b/core/src/main/java/feign/Capability.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2012-2020 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package feign;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.List;
+import feign.Logger.Level;
+import feign.Request.Options;
+import feign.codec.Decoder;
+import feign.codec.Encoder;
+
+public interface Capability {
+
+  static <E> E enrich(E target, List<Capability> capabilities) {
+    return capabilities.stream()
+        .reduce(
+            target,
+            (c, capability) -> invoke(c, capability),
+            (a, b) -> b);
+  }
+
+  static <E> E invoke(E target, Capability capability) {
+    return Arrays.stream(capability.getClass().getMethods())
+        .filter(method -> method.getName().equals("enrich"))
+        .filter(method -> method.getReturnType().isInstance(target))
+        .findFirst()
+        .map(method -> {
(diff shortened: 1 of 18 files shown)


core/src/main/java/feign/Capability.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.List;
+import feign.Logger.Level;
+import feign.Request.Options;
+import feign.codec.Decoder;
+import feign.codec.Encoder;
+
+public interface Capability {
```
kdavisk6: I don't really know what this component is meant to to do?  Can we add some documentation here?

core/src/main/java/feign/Capability.java
```diff
@@ -0,0 +1,94 @@
...
+
+  static <E> E enrich(E target, List<Capability> capabilities) {
+    return capabilities.stream()
+        .reduce(
+            target,
+            (c, capability) -> invoke(c, capability),
+            (a, b) -> b);
```
kdavisk6: This reads like very old `C` code.  Can we please update this to be more specific?  I can't understand what's happening here enough to comment.

metrics5/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.github.openfeign</groupId>
+    <artifactId>parent</artifactId>
+    <version>10.7.5-SNAPSHOT</version>
+  </parent>
+  <artifactId>feign-metrics5</artifactId>
+  <name>Feign Metrics5</name>
+  <description>Feign Dropwizard Metrics 5</description>
```
kdavisk6: I'd like to suggest an alternative, [Micrometer](https://micrometer.io/docs).  It can wrap Dropwizard and will allow this to be integrated directly into Spring Boot/Cloud, since Micrometer is their preferred metrics collector.

metrics5/src/main/java/feign/metrics5/MeteredBody.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  public boolean isRepeatable() {
+    return delegate.isRepeatable();
+  }
+
+  @Override
+  public InputStream asInputStream() throws IOException {
+    // TODO, ideally, would like not to bring guava just for this
+    final CountingInputStream input = new CountingInputStream(delegate.asInputStream());
```
kdavisk6: This implementation is pretty simple, we could wrap and write this ourselves.

core/src/main/java/feign/Capability.java
velo: Yeah, this look odd, I will add some docs explaining.

metrics5/pom.xml
velo: My plan is to add that one later, added metrics 5 first cause is the one I'm using and gets values for me faster

kdavisk6: They both do the same thing. If you want to have multiple implementations I’d like to suggest that we rename this to metrics-dropwizard

PR #1182, opened 2020-03-03
Configurable to disable streaming mode for Default client

### LINKED ISSUE #260: Empty body on HTTP error response
I had a problem where the response body would come up empty, when the server responded with 401:

```
2015-08-31 17:11:01.168 [MyInterface#myMethod] <--- HTTP/1.1 401 Unauthorized (12ms)
2015-08-31 17:11:01.169 [MyInterface#myMethod] Date: Mon, 31 Aug 2015 15:11:01 GMT
2015-08-31 17:11:01.169  [MyInterface#myMethod] Content-Length: 71
2015-08-31 17:11:01.170  [MyInterface#myMethod] Content-Type: application/json
2015-08-31 17:11:01.170  [MyInterface#myMethod] <--- END HTTP (0-byte body)

```

As indicated by content-length header, the body should be 71 bytes, and it is (verified with tcpdump).
The empty body caused a NPE when the ErrorDecoder had nothing to decode and returned null, which SynchronousMethodHandler then tried to throw (line 121).

During previous testing, with 200 OK it worked fine, if I recall correct.

This was with the Default client. Using OkHttpClient instead, the issue disappeared.

verils:
* Add a property `allowStramingMode` in `Request.Option`.

* Optional to disable the streaming mode for `HttpURLConnection`.

* The property is default to `true` for compatibility.

Fix #260

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -19,8 +19,10 @@ import static feign.Util.ENCODING_DEFLATE;
 import static feign.Util.ENCODING_GZIP;
 import static feign.Util.checkArgument;
 import static feign.Util.checkNotNull;
+import static feign.Util.isBlank;
 import static feign.Util.isNotBlank;
 import static java.lang.String.format;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -35,9 +37,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.zip.DeflaterOutputStream;
 import java.util.zip.GZIPOutputStream;
+
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLSocketFactory;
+
 import feign.Request.Options;
 
 /**
@@ -161,10 +165,12 @@ public interface Client {
       }
 
       if (request.body() != null) {
-        if (contentLength != null) {
-          connection.setFixedLengthStreamingMode(contentLength);
-        } else {
-          connection.setChunkedStreamingMode(8196);
+        if (options.isAllowStreamingMode()) {
+          if (contentLength != null) {
+            connection.setFixedLengthStreamingMode(contentLength);
+          } else {
+            connection.setChunkedStreamingMode(8196);
+          }
         }
         connection.setDoOutput(true);
         OutputStream out = connection.getOutputStream();
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -234,6 +234,7 @@ public final class Request {
     private final long readTimeout;
     private final TimeUnit readTimeoutUnit;
     private final boolean followRedirects;
+    private final boolean allowStreamingMode;
 
     /**
      * Creates a new Options instance.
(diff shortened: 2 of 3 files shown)


core/src/main/java/feign/Client.java
```diff
@@ -161,10 +165,12 @@ HttpURLConnection convertAndSend(Request request, Options options) throws IOExce
...
       if (request.body() != null) {
-        if (contentLength != null) {
-          connection.setFixedLengthStreamingMode(contentLength);
-        } else {
-          connection.setChunkedStreamingMode(8196);
+        if (options.isAllowStreamingMode()) {
+          if (contentLength != null) {
+            connection.setFixedLengthStreamingMode(contentLength);
+          } else {
+            connection.setChunkedStreamingMode(8196);
```
velo: maybe make this configurable?!

core/src/main/java/feign/Request.java
```diff
@@ -269,6 +291,7 @@ public Options(long connectTimeout, TimeUnit connectTimeoutUnit,
       this.readTimeout = readTimeout;
       this.readTimeoutUnit = readTimeoutUnit;
       this.followRedirects = followRedirects;
+      this.allowStreamingMode = true;
```
velo: like the idea of defaulting to true

core/src/main/java/feign/Client.java
```diff
@@ -161,10 +165,12 @@ HttpURLConnection convertAndSend(Request request, Options options) throws IOExce
       }
 
       if (request.body() != null) {
-        if (contentLength != null) {
-          connection.setFixedLengthStreamingMode(contentLength);
-        } else {
-          connection.setChunkedStreamingMode(8196);
+        if (options.isAllowStreamingMode()) {
```
velo: From the documentation, the 2 `set*StreamingMode()` states:

> This method is used to enable streaming of a HTTP request body without internal buffering, when the content length

what if we call this new option `disableInternalBuffering`?

`allowStreamingMode` makes me scratch the head.

Also, this seem to be something very specific to `java.net.HttpURLConnection`.  May be move to `Client.Default`?

or is this something we can enable for ok_http, hc5, apache http 4, google http and so on.

verils: Yes, indeed, it's a specific feature for `java.net.HttpURLConnection`, moving the option to `Client.Default` is better. Why didn't I think of that? :-D

I call it `allowStreamingMode` because the word is just literal. I think if one knows the 2 `set*StreamingMode()` methods in `HttpURLConnection`, so he can understand what streaming mode is.

After moving `allowStreamingMode` into `Client.Default`, it will be related to that specific implementation. Then it may be easier to understand.

velo: yeah, in my mind, `allowStreamingMode` doesn't say much, like `allowFlushingMode`... I have no idea what either mean.

now, `disableInternalBuffering`, at least for me, has an exact meaning... client has some kind of internal buffer, and it's being turned off.

Anyway, I won't be picky about this, do what you feel best about.

verils: Discussed with my colleagues. Now we all agree `allowStreamingMode` is not so good, and we checked the source code found the buffering only happens in writing request output stream, so the final desicion is call it `disableRequestBuffering`

velo: Sound great

core/src/test/java/feign/client/AbstractClientTest.java
```diff
@@ -146,6 +146,22 @@ public void parsesErrorResponseBody() {
...
+
+    server.enqueue(new MockResponse().setResponseCode(401).setBody("ARGHH"));
+
+    TestInterface api = newBuilder()
+        .target(TestInterface.class, "http://localhost:" + server.getPort());
+
+    try {
+      api.postForString("HELLO");
+    } catch (FeignException e) {
+      assertThat(e.contentUTF8()).isEqualTo(expectedResponseBody);
```
velo: Is there anything we can check on the server?

verils: I don't think so, it responses as normal just fine

velo:
You need to run `mvn license:format` to fix this error

verils:
> 
> 
> You need to run `mvn license:format` to fix this error

Thanks for the tip. 

I found a funny thing. The `GoogleHttpClient` however uses `java.net.HttpURLConnection` as underlying transport implementation and it causes the same problem. I can do nothing in this repo to deal with this situation but ignore the test I've added in `GoogleHttpClientTest`.

I know it's not very well to do such thing, but the `GoogleHttpClient` is not my concern. The only thing we can do is waiting for a new released version of google client which fix this issue.

Issue #1183, opened 2020-03-06
Allow Overrides for Parameter Name when expanding Beans or POJO


boggard:
Now `FieldQueryMapEncoder` put request parameter with name equals to field name.

What about add some annotation to allow override parameter  name in query map?

Something like `@JsonAlias` in jackson.

boggard:
The issue relates to `BeanQueryMapEncoder` too.

kdavisk6:
Can you please add the following so we can understand your issue better. 

* expected behavior with observed behavior
* provide a minimum working example that demonstrates your issue

boggard:
I provide pull request to show what i mean and how i see the implementation

kdavisk6:
An update to this issue, the PR talks about allowing for an override to the naming scheme used by both the `FieldQueryMapEncoder` and `BeanQueryMapEncoder`, which are used when a Java Bean or POJO is being expanded as part of `@QueryMap` usage.

The suggestion from the maintainers is to use the `@Param` annotation and update the name resolution logic in both of the aforementioned encoders to honor the annotation if present.

velo:
#1184

PR #1184, opened 2020-03-06
Overriding query parameter name

boggard:
Relates to #1183

### DIFF
--- a/core/src/main/java/feign/Param.java
+++ b/core/src/main/java/feign/Param.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2019 The Feign Authors
+ * Copyright 2012-2020 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


core/src/main/java/feign/ParamAlias.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Allows specify query parameter name different from field name, when Object is used like parameters container
+ *
+ * @see feign.querymap.BeanQueryMapEncoder
+ * @see feign.querymap.FieldQueryMapEncoder
+ */
+@Retention(RUNTIME)
+@Target({ElementType.FIELD, ElementType.METHOD})
+public @interface ParamAlias {
```
velo: Any particular reason no to use existing `@Param`

kdavisk6: That would be my recommendation as well.  Consider using our existing `@Param` annotation.

kdavisk6:
Please add a description referencing the issue this PR is for.

boggard:
I used `@Param` following the comments above. Do i need to use `Param.Expander` for parameter value, when `@Param` used?

PR #1195, opened 2020-03-21
Allow empty expression in header values

### LINKED ISSUE #1172: Sending "{}" as header value fails since 10.2.0
The following minimal reproducer in https://github.com/neiser/feign-reproduce-empty-json-header-value-bug using the code

```java
    public static void main(String[] args) {
        TestClient testClient = Feign.builder().target(TestClient.class, "http://google.com");
        testClient.get(""); // works!
        testClient.get("{\"a\":  \"b\"}"); // works!
        testClient.get("{}"); // fails!
    }


    private interface TestClient {
        @Headers({
            "Some-Header-Key: {someHeaderValue}"
        })
        @RequestLine("GET /")
        void get(@Param("someHeaderValue") String someHeaderValue);
    }
```

shows that sending the string `{}` as a header value fails with 

```
Exception in thread "main" java.lang.IllegalArgumentException: an expression is required.
	at feign.template.Expressions.create(Expressions.java:48)
	at feign.template.Template.parseFragment(Template.java:194)
	at feign.template.Template.parseTemplate(Template.java:178)
	at feign.template.Template.<init>(Template.java:60)
	at feign.template.HeaderTemplate.<init>(HeaderTemplate.java:81)
	at feign.template.HeaderTemplate.create(HeaderTemplate.java:57)
	at feign.RequestTemplate.lambda$appendHeader$3(RequestTemplate.java:742)
	at java.util.Map.compute(Map.java:1093)
	at feign.RequestTemplate.appendHeader(RequestTemplate.java:740)
	at feign.RequestTemplate.header(RequestTemplate.java:710)
	at feign.RequestTemplate.header(RequestTemplate.java:692)
	at feign.RequestTemplate.resolve(RequestTemplate.java:236)
	at feign.ReflectiveFeign$BuildTemplateByResolvingArgs.resolve(ReflectiveFeign.java:328)
	at feign.ReflectiveFeign$BuildTemplateByResolvingArgs.create(ReflectiveFeign.java:232)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:75)
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:100)
	at $Proxy4.get(Unknown Source)
	at Main.main(Main.java:12)
```

This bug still appears in version 10.7.4 and is failing since version 10.2.0. In 10.0.1, the above code works.

neiser:
Allow empty expression `{}` as a literal within header template values. In any other template (Uri, Body, Query), they're still not allowed (there the expression should be pct-encoded in those cases). 

Fixes #1172.

### DIFF
--- a/core/src/main/java/feign/template/BodyTemplate.java
+++ b/core/src/main/java/feign/template/BodyTemplate.java
@@ -51,7 +51,7 @@ public final class BodyTemplate extends Template {
   }
 
   private BodyTemplate(String value, Charset charset) {
-    super(value, ExpansionOptions.ALLOW_UNRESOLVED, EncodingOptions.NOT_REQUIRED, false, charset);
+    super(value, ExpansionOptions.ALLOW_UNRESOLVED, EncodingOptions.NOT_REQUIRED, false, false, charset);
     if (value.startsWith(JSON_TOKEN_START_ENCODED) && value.endsWith(JSON_TOKEN_END_ENCODED)) {
       this.json = true;
     }
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -40,12 +40,17 @@ public final class Expressions {
         SimpleExpression.class);
   }
 
-  public static Expression create(final String value) {
+  public static Expression create(final String value, boolean allowEmpty) {
 
     /* remove the start and end braces */
     final String expression = stripBraces(value);
     if (expression == null || expression.isEmpty()) {
-      throw new IllegalArgumentException("an expression is required.");
+      if(allowEmpty) {
+        return null;
+      }
+      else {
+        throw new IllegalArgumentException("an expression is required.");
+      }
     }
 
     Optional<Entry<Pattern, Class<? extends Expression>>> matchedExpressionEntry =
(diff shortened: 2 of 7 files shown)


neiser:
I've investigated it a bit more and I'm a bit puzzled if this is actually the right way to go. The true problem I think lies in `RequestTemplate.java:236`. As far as I understand it, the header template is parsed and resolved, then added again, where it's parsed again. Providing a value which is resolved to something which looks like an expression leads to the error this PR tries to fix. Maybe avoiding parsing expressions again after resolving to values is the actual fix for #1172? 

```java
 if (!this.headers.isEmpty()) {
      /*
       * same as the query string, we only want to keep resolved values, so clear the header map on
       * the resolved instance
       */
      resolved.headers(Collections.emptyMap());
      for (HeaderTemplate headerTemplate : this.headers.values()) {
        /* resolve the header */
        String header = headerTemplate.expand(variables);
        if (!header.isEmpty()) {
          /* split off the header values and add it to the resolved template */
          String headerValues = header.substring(header.indexOf(" ") + 1);
          if (!headerValues.isEmpty()) {
            resolved.header(headerTemplate.getName(), headerValues);
          }
        }
      }
    }
```

kdavisk6:
See my note in #1172 

This will require some significant refactoring to correct.  I'm going to close this PR in favor of a new upcoming PR that will address the issue.

Issue #1197, opened 2020-03-26
@RequestParam empty lists are expanded as singleton lists containing `null`


darrenfoong:
I'm using Spring Cloud OpenFeign, and I have a use case where I need to send a `List` as a `@RequestParam`, for example:

```
@GetMapping("/items")
Page<Item> getItem(@RequestParam List<String> manufacturer);
```

When I send an empty list (e.g. `remoteService.getItem(Collections.emptyList())`), the receiver receives a singleton list with a single `null`.

**Cause**

The issue seems to be in `Expressions.expandIterable()` ([link](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/template/Expressions.java#L132)), which expands an empty list into `null`, which gets returned to `expand()`, whose `StringBuilder` appends a `null`. `StringBuilder` will append a literal `null` to the internal string buffer, resulting in an expansion like `/items?manufacturer=null`.

**Attempted fix**

I forked Feign and checked the above behaviour in `QueryTemplateTest`. In my investigation (https://github.com/OpenFeign/feign/compare/master...darrenfoong:query-template-empty-list), I modified `expandIterable()` and added tests to check the expected behaviour.

EzioL:
I also met this bug. 😂You can set `required = false` to solve the problem first.

Here is appendNull()  code
` private AbstractStringBuilder appendNull() {
        int c = count;
        ensureCapacityInternal(c + 4);
        final char[] value = this.value;
        value[c++] = 'n';
        value[c++] = 'u';
        value[c++] = 'l';
        value[c++] = 'l';
        count = c;
        return this;
    }`

kdavisk6:
@darrenfoong 

This appears to be a bug in our expansion.  The expected behavior here is that the parameter is omitted completely.  Here is a link to our documentation on undefined vs empty

https://github.com/OpenFeign/feign#undefined-vs-empty-values

So, for this to be correct, these are the test cases and their results:

```java
// empty
Collection<String> items = Collections.emptyList();

// /items?manufacturer=

// null collection
Collection<String> items = null;

// /items
```

I'll look over your PR and review.

jebeaudet:
Any update on this? I also think this should be backported to the 10.7 release as this is the release supported by spring boot 2.2/spring cloud hoxton. Thanks

kdavisk6:
Closed via #1200

Issue #1194, opened 2020-03-20
OpenFeign 10.7.4, 10.7.3 and 10.8 break backward compatibility


sofax:
Updating from v10.7.2 to v10.7.3, v10.7.4 or v10.8 breaks existing code, because `Request.requestBody()` was removed. The following line won't compile anymore:

```
      final Request.Body body = request.requestBody();
```

kdavisk6:
@sofax, that move was intentional.  `requestBody()` was an internal method that mistakenly was made public.  Use `Request#body`.

sofax:
That means we will NEVER be able to update OpenFeign, because we deliver artifacts that MUST NOT break existing code with minor, or - as in this case - even patch level version upgrades.

Sorry, guys, this is very bad! :(

velo:
hi @sofax feel free to submit a PR with a fix for this.

Keep in mind the method will be removed for feign 11.

sofax:
We are more interested in bug fixes for the 10.x branch, not a major upgrade. Breaking the code with a patch release violates the idea of semantic versioning, which I thought/was hoping you are adhering to. If the exposition of an internal API in a previous release was an accident you should provide a workaround for it instead of just hiding it again, or you will make all upcoming bugfix releases unusable for most people.

We will (have to) stick with v10.7.2 then.

velo:
> We will (have to) stick with v10.7.2 then.

Or, you can get the souces and submit a PR restoring the missing method.

Also, feel free to also include the necessary tooling to prevent API for ever being broken again.

kdavisk6:
Closed by #1200

PR #1200, opened 2020-03-30
Fix expansion of @RequestParam empty lists

darrenfoong:
This PR addresses #1194.

`Expressions.expandIterable()` ([link](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/template/Expressions.java#L132)), expands an empty list into `null`, which gets returned to `expand()`, whose `StringBuilder` appends a `null`. `StringBuilder` will append a literal `null` to the internal string buffer, resulting in an expansion of an empty list into something like `/items?manufacturer=null`, which will be serialized by the caller as a singleton list with a single `null`.

This PR modifies `expandIterable()` to return an empty string instead of `null` in the case of an empty list.

### DIFF
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2019 The Feign Authors
+ * Copyright 2012-2020 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


kdavisk6:
A note on Travis, the build succeeded however the status did not make it back to GitHub.  This PR is green.

darrenfoong:
I realised that I had been referencing the wrong issue all this time: it should be #1197 (which I created) and not #1194 (which was in the original pull request body).

I've updated the original comment and apologies for any confusion caused. In any case, this should close #1197? @kdavisk6

PR #1219, opened 2020-05-02
Headers overwrite

### LINKED ISSUE #1217: JAXRS contract on API with @HeaderParam("Content-Type") and @Consumes with multiple values cause feign client build fail
Version : 11.0

API :
```
    @POST
    @Path("/{stream:.+}")
    @Consumes({"application/xlsx", "text/csv"})
    @Produces("application/json")
    List<Integer> sourceFile(
            @PathParam("stream") String stream,
            @QueryParam("delimiterChar") @DefaultValue(",") String delimiterChar,
            @HeaderParam("Content-Type") String contentType,
            @RequestBody(required = true, description = "Input file") InputStream file
    ) throws IOException;
```

Calling API with :
```
myCtrl.sourceFile("stream", null, "text/csv", inputStream);
```

Causes the error :
```
java.lang.NullPointerException: null
	at feign.okhttp.OkHttpClient.toOkHttpRequest(OkHttpClient.java:64) ~[feign-okhttp-11.0.jar:na]
	at feign.okhttp.OkHttpClient.execute(OkHttpClient.java:168) ~[feign-okhttp-11.0.jar:na]
	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:119) ~[feign-core-11.0.jar:na]
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:89) ~[feign-core-11.0.jar:na]
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:100) ~[feign-core-11.0.jar:na]
	at com.sun.proxy.$Proxy143.sourceFile(Unknown Source) ~[na:na]
```

Because at OkHttpClient:62 (`mediaType = MediaType.parse(value);`), `value` variable equals `"application/xlsx, text/csv"` instead of `"text/csv"`.
Method `MediaType.parse(value)` returns `null` as `"application/xlsx, text/csv"` is not a valid media type.

velo:
Fixes #1217

Specially when using contracts, there is more than one way to set some headers.

`Content-Type` seems to be the most special one, were client sends one.

This change will guarantee that only one value for this header will survive.

@kdavisk6 do you think there is any situation feign would wan't to send more than one Content-Type header?

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -696,7 +696,7 @@ public final class RequestTemplate implements Serializable {
 
   /**
    * Add a header using the supplied Chunks.
-   * 
+   *
    * @param name of the header.
    * @param chunks to add.
    * @return a RequestTemplate for chaining.
@@ -751,6 +751,14 @@ public final class RequestTemplate implements Serializable {
     if (!values.iterator().hasNext()) {
       /* empty value, clear the existing values */
       this.headers.remove(name);
+      return this;
+    }
+    if (name.equals("Content-Type")) {
+      // a client can only produce content of one single type, so always override Content-Type and
+      // only add a single type
+      this.headers.remove(name);
+      this.headers.put(name,
+          HeaderTemplate.create(name, Collections.singletonList(values.iterator().next())));
       return this;
     }
     this.headers.compute(name, (headerName, headerTemplate) -> {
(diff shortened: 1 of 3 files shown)


kdavisk6:
Content type is not a multi-value header according to the spec. Are you sure you don’t mean Accept here?

velo:
> Content type is not a multi-value header according to the spec. 
Exactly, but feign was adding multiple values on the Content-Type request.

> Are you sure you don’t mean Accept here?
No, content type.

PR #1220, opened 2020-05-05
fixed query map with array parameter bug

### LINKED ISSUE #1170: @SpringQueryMap doesn't support Map<String, String[]> issue
Hi, I read a issue in spring-cloud-openfeign associated with @SpringQueryMap. 

link - https://github.com/spring-cloud/spring-cloud-openfeign/issues/256#issuecomment-582547145

If it is bug, I think some code should be added at "addQueryMapQueryParameters" method in ReflectiveFeign to solve this issue like this.
```java
Object currValue = currEntry.getValue();
  if (currValue instanceof Iterable<?>) {
  	Iterator<?> iter = ((Iterable<?>) currValue).iterator();
  	while (iter.hasNext()) {
            Object nextObject = iter.next();
            values.add(nextObject == null ? null : encoded ? nextObject.toString() : UriUtils.encode(nextObject.toString()));
  	}
  } else if (currValue instanceof Object[]) {
  	for (Object nextObject : (Object[]) currValue) {
            values.add(nextObject == null ? null : encoded ? nextObject.toString() : UriUtils.encode(nextObject.toString()));
  	}
  } else {
  	values.add(currValue == null ? null : encoded ? currValue.toString() : UriUtils.encode(currValue.toString()));
  }
```

thanks.

dharmeshjogadia:
close #1170

### DIFF
--- a/core/src/test/java/feign/FeignTest.java
+++ b/core/src/test/java/feign/FeignTest.java
@@ -24,6 +24,7 @@ import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.SocketPolicy;
 import okio.Buffer;
+import org.assertj.core.util.Maps;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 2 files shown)


core/src/test/java/feign/FeignTest.java
```diff
@@ -947,6 +960,9 @@ void form(
     @RequestLine("GET /?1={1}&2={2}")
     Response queryParams(@Param("1") String one, @Param("2") Iterable<String> twos);
 
+    @RequestLine("GET /")
+    Response queryArrParams(@QueryMap Map<String, String[]> twos);
```
kdavisk6, line 964: nit: use full name, `queryMapWithArrayValues`

Issue #1221, opened 2020-05-05
Suggestion to upgrade okhttp3 from 3.14.6 to 4.6.0


Sam-Kruglov:
Hi, trying to use the Logbook logging interceptor but they are using 4.x version of OkHttp. Maven by default picks the lowest, which goes from `feign-okhttp` jar, which makes Logbook break on non-existing method. Just an inconvenience that I have to redefine okhttp to 4.x manually, fortunately, it seems to work.

velo:
That is doable. Send us a PR and I will make sure to review it.

FWIW, you can pick the version you need by using
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

jonatan-ivanov:
Also, it seems okhttp 3.x only receives security fixes: https://github.com/square/okhttp/issues/6019

velo:
Awesome.

Sam-Kruglov:
Alright

PR #1223, opened 2020-05-06
#1221: bump okhttp3 version from 3.14.6 to 4.6.0

Sam-Kruglov:
See #1221

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,8 +70,7 @@
 
     <main.basedir>${project.basedir}</main.basedir>
 
-    <okhttp3.client.version>3.14.6</okhttp3.client.version>
-    <okhttp3.mockwebserver.version>3.14.6</okhttp3.mockwebserver.version>
+    <okhttp3.version>4.6.0</okhttp3.version>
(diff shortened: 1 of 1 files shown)


okhttp/src/main/java/feign/okhttp/OkHttpClient.java
```diff
@@ -84,7 +84,7 @@ static Request toOkHttpRequest(feign.Request input) {
       }
     }
 
-    RequestBody body = inputBody != null ? RequestBody.create(mediaType, inputBody) : null;
+    RequestBody body = inputBody != null ? RequestBody.create(inputBody, mediaType) : null;
```
Sam-Kruglov, line 87: the previous signature is deprecated

velo: I would say to use the deprecated one, so we keep feign compatible with a wider number or releases

Sam-Kruglov:
Thanks, @velo! Can we merge it?

velo:
> Thanks, @velo! Can we merge it?

Will do, as soon I get to a computer... Still 6am here :)

PR #1228, opened 2020-05-19
update comment Retryer.java

paladim:


core/src/main/java/feign/Retryer.java
```diff
@@ -83,7 +83,7 @@ public void continueOrPropagate(RetryableException e) {
      * The interval increases exponentially with each attempt, at a rate of nextInterval *= 1.5
      * (where 1.5 is the backoff factor), to the maximum interval.
      *
-     * @return time in nanoseconds from now until the next attempt.
+     * @return time in milliseconds from now until the next attempt.
```
vivekmore, line 86: 👍🏼 just wasted 30 mins trying to understand why period would be in nanos and maxPeriod in millis 🤣

PR #1233, opened 2020-06-02
feat: java11 request options timeout.

ChunMengLu:
Config feign-java11 request timeout use feign `Request.options`.

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -59,6 +59,7 @@ public class Http2Client implements Client {
     try {
       httpResponse = client.send(httpRequest, BodyHandlers.ofByteArray());
     } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
       throw new IOException("Invalid uri " + request.url(), e);
     }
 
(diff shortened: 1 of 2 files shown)


ChunMengLu:
java11 javadoc:

```text
HttpRequest.Builder timeout​(Duration duration)
Sets a timeout for this request. If the response is not received within the specified timeout then an HttpTimeoutException is thrown from HttpClient::send or HttpClient::sendAsync completes exceptionally with an HttpTimeoutException. The effect of not setting a timeout is the same as setting an infinite Duration, ie. block forever.
Parameters:
duration - the timeout duration
Returns:
this builder
Throws:
IllegalArgumentException - if the duration is non-positive
```

But I found that it is not very good in the actual test, and running the test directly does not work, but I broke the point to the jdk.internal.net.http.HttpClientImpl#purgeTimeoutsAndReturnNextDeadline method, and found that timeout is actually supported,The jdk.internal.net.http.ResponseTimerEvent#handle method will execute and throw an exception.

ChunMengLu:
> LGTM, can merge once the build is passing.

ok!

gb-klein:
@ChunMengLu , @velo : One thing i noticed is, that you only considered the request timeout, but not the connect timeout. This would have to be set on the HttpClient instance (e.g. via Builder::connectTimeout). One could do this manually by providing a custom built HttpClient to Http2Client. However it would be good to leave this over to Http2Client when it constructs the HttpClient for us. E.g. by replacing the default constructor with a constructor taking Options or (IMHO better) lazily instantiating HttpClient - if no custom was specified via constructor - when the request should be sent.

Issue #1210, opened 2020-04-16
Can't determine if `RequestTemplate` or `Request` has a `null` body


jebeaudet:
Hi

In the process of updating feign from version 10.2.3 to 10.7.4, I noticed some deprecation in my codebase on internal code dealing with `RequestTemplate` and `Request`. The methods `RequestTemplate::requestBody` has been marked as deprecated and `Request::requestBody` have been removed. 

I'm fine with moving on and building something better obviously but it seems there is no non deprecated replacement to detect a `null` body in both classes. For example, this simple code will throw a NPE : 
```
    public static void main(String[] args)
    {
        Request.create(HttpMethod.GET,
                       "http://patate.com",
                       Collections.emptyMap(),
                       (Request.Body) null,
                       (RequestTemplate) null)
               .body();
    }
```

If there is a non nullability constraint in both classes, there should be preconditions in the constructors and clear java documentation for users. 

Thanks

kdavisk6:
@jebeaudet 

Documentation in the latest version does indicate that `body` and `requestTemplate` can be `null`, however the function `Request#body` does not properly check the `body` to ensure that it is present.  This is an oversight on our part.  I've marked this issue as a bug for us to triage and correct.  Thanks!

Dainerx:
Hello, 

Any updates on this?

PR #1236, opened 2020-06-09
#1210

finnetrolle:
Hi all,
This is an attempt to fix an issue. I think if we have null-body for request it is better to provide null back if client ask Request#Body. But in fact we're sending back not Body but body's data. In that case I don't think null is ok for byte array. So I decided to create constant EMPTY body and when we have null body to create Request we can just put that constant instead null.


velo:
🤨

Issue #1268, opened 2020-09-03
Cannot get response headers from a FeignException


pzapasni:
TL;DR: Is it possible to retrieve response headers from a ``FeignException``?

My use case is this: I'm using Feign from microservice A to call microservice B. If the call fails with a 4xx status code (which results in a ``FeignClientException`` being thrown), the response may contain information about what went wrong, which fields were missing in the request, etc, so I'd like to relay the response to the caller of microservice A. I've created a Spring exception handler for this purpose:

```kotlin
@RestControllerAdvice
class FeignExceptionHandler {

    @ExceptionHandler(FeignException.FeignClientException::class)
    fun handleFeignClientException(e: FeignException.FeignClientException): ResponseEntity<String> {
        val headers = HttpHeaders()
        headers.contentType = MediaType.APPLICATION_JSON
        return ResponseEntity(e.contentUTF8(), headers, HttpStatus.valueOf(e.status()))
    }

}
```

The thing is, that the response content type may not always be ``application/json``, so I'd like to fetch the content type from the response itself (if it's available). However, I found no easy way to retrieve the response headers from a ``FeignException``, only the response body is available. Am I missing something obvious here? Or maybe is there a better way to achieve what I'm trying to do?

kdavisk6:
@pzapasni 

`FeignException` does not collect response headers.  We would need to modify it to do so.  Feel free to submit a PR for this.

pzapasni:
@kdavisk6 I've added the implementation in #1277

PR #1277, opened 2020-09-23
Add response headers to FeignException

pzapasni:
Implementation of #1268 

This PR adds the ability to retrieve the response headers from a ``FeignException``.

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -22,6 +22,7 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
@@ -40,18 +41,22 @@ public class FeignException extends RuntimeException {
   private static final long serialVersionUID = 0;
   private int status;
   private byte[] responseBody;
+  private Map<String, Collection<String>> responseHeaders;
   private Request request;
 
   protected FeignException(int status, String message, Throwable cause) {
     super(message, cause);
     this.status = status;
+    this.responseHeaders = Collections.emptyMap();
     this.request = null;
   }
 
-  protected FeignException(int status, String message, Throwable cause, byte[] responseBody) {
+  protected FeignException(int status, String message, Throwable cause, byte[] responseBody,
+      Map<String, Collection<String>> responseHeaders) {
     super(message, cause);
     this.status = status;
     this.responseBody = responseBody;
+    this.responseHeaders = responseHeaders;
     this.request = null;
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/feign/FeignException.java
```diff
@@ -40,58 +41,70 @@
...
     this.request = null;
   }
 
-  protected FeignException(int status, String message, Throwable cause, byte[] responseBody) {
+  protected FeignException(int status, String message, Throwable cause, byte[] responseBody,
+      Map<String, Collection<String>> responseHeaders) {
     super(message, cause);
     this.status = status;
     this.responseBody = responseBody;
+    this.responseHeaders = responseHeaders;
```
kdavisk6, line 59: Despite the fact that Strings are immutable, the Map is not.  Can you change this to create a copy of the map?  Some examples are `Map.of(responseHeaders)` or `new LinkedHashMap<>(responseHeaders)`

core/src/main/java/feign/FeignException.java
```diff
@@ -142,13 +155,24 @@ public String contentUTF8() {
...
   }
 
+  /**
+   * The Response headers.
+   *
+   * @return the headers from the {@link Response}.
+   * @see Response#headers()
+   */
+  public Map<String, Collection<String>> responseHeaders() {
+    return this.responseHeaders;
```
kdavisk6, line 165: We should also wrap this as an immutable collection.  Can we wrap with `Collections.unmodifiableMap`?

kdavisk6 (changes requested):
@pzapasni 

Thank you for taking the time to put this together.  I have two small concurrency notes.  Address these and this PR is good to merge.

core/src/main/java/feign/FeignException.java
pzapasni: Sure, but if we wrap the response headers as an immutable collection inside the constructor, shouldn't that be sufficient?

core/src/main/java/feign/FeignException.java
pzapasni: Will do. I'm just wondering if wrapping the map itself is enough? The values of this map are of type ``Collection<String>``, so although wrapping the map would make it immutable, the values would still be mutable. To make this "truly" immutable, we would need to wrap the values as well. Or do you think that's an overkill?

kdavisk6: Strings are immutable by nature, so to ensure immutability, you must lock down the map.

core/src/main/java/feign/FeignException.java
kdavisk6: I should be, but we don't advertise that to the consumers.  It's also good practice to return a copy of the internal collection just in case.

kdavisk6:
@pzapasni Apologies for the late replies.

core/src/main/java/feign/FeignException.java
virtual-machinist: I'd argue that it's better to do the copy on construction (and optionally wrap it in `Collections.unmodifiableMap()`) instead of returning a new copy each time.

core/src/main/java/feign/FeignException.java
virtual-machinist: True that but having an immutable map still gives you the option of `responseHeaders().get("Something").clear()` unless `Collection<String>` isn't also wrapped in an `Collections.unmodifiableCollection()` at some point before.

virtual-machinist:
Any chance this gets released for X-mas?  :santa:

nakker:
Love this change. +1 for sorting out the remaining issues.

radio-rogal:
As far as I understand the source repository is no longer exists.

Could I re-create PR as new one?

core/src/main/java/feign/FeignException.java
radio-rogal: Response has the method `caseInsensitiveCopyOf`. I think it should be moved to `Utll` and reused here.

Issue #1286, opened 2020-10-13
ApacheHttpClient unreleased socket


austinpio:
the Openfeign client - `ApacheHttpClient` - does not explicitly close the socket resource. 
The recommended approach according to the examples specified in the Apache HttpComponents is to explicitly call `CloseableHttpResponse#close()` from a `finally` clause.
[Example from apache readme](https://hc.apache.org/httpcomponents-client-ga/quickstart.html)

This might lead to resource leaks due to unreleased socket resources in the event of errors and other exception scenarios.

[after reading the response the program should explicitly close the socket](https://github.com/OpenFeign/feign/blob/master/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java#L169)

velo:
I made a PR for this, lemme know what you think of it.

PR #1287, opened 2020-10-15
Fix for #1286

velo:
When feign is done with the response, also invoke close on http response (if closeable)

#1286


httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -233,6 +231,8 @@ public Reader asReader(Charset charset) throws IOException {
       @Override
       public void close() throws IOException {
         EntityUtils.consume(entity);
+        if (httpResponse instanceof CloseableHttpResponse)
+          ((CloseableHttpResponse) httpResponse).close();
```
austinpio, line 235: could this close method call be enclosed in a `finally` block? that would ensure the socket is released even in the event of an exception when consuming the entity.
It is also the approach recommended here - https://hc.apache.org/httpcomponents-client-ga/quickstart.html
quoted text from the document 

> // In order to ensure correct deallocation of system resources
> // the user MUST call CloseableHttpResponse#close() from a finally clause.
>try{
>    EntityUtils.consume(entity1);
>} finally {
>    response1.close();
>}

-  Additionally it would also not show up as a potential issue in static code analyzers like Fortify, which expect to see a finally block which closes the connection.

austinpio (changes requested):

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
austinpio: hi @velo you reckon, this finally block something we could incorporate in here?

velo: good point, lemme me do it

austinpio: That would be really great. Thank you @velo

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -233,6 +231,8 @@ public Reader asReader(Charset charset) throws IOException {
       @Override
       public void close() throws IOException {
         EntityUtils.consume(entity);
+        if (httpResponse instanceof CloseableHttpResponse)
```
austinpio, line 234:
```suggestion
        try {
          EntityUtils.consume(entity);
        } finally {
          if (httpResponse instanceof CloseableHttpResponse)
            ((CloseableHttpResponse) httpResponse).close();
        }
```

austinpio: hi @velo 
Please do have a look at this, when you are free.
I do not have write permissions to apply this change.
Thank you

austinpio (changes requested):

austinpio (changes requested):
Requesting changes

Bangalor:
```java
    final HttpEntity entity = httpResponse.getEntity();
    if (entity == null) {
      return null;
    }
```
apache response won't be closed in case of missing entity.
Is this case should be handled as well?

Issue #1257, opened 2020-07-22
FieldQueryMapEncoder may not be thread-safe


542803359:
![image](https://user-images.githubusercontent.com/46744543/88133164-f9771f80-cc13-11ea-93bc-0378f476c409.png)

建议使用ConcurrentHashMap 
eg:private final Map<Class<?>, ObjectParamMetadata> classToMetadata =
            new ConcurrentHashMap<Class<?>, ObjectParamMetadata>();

![image](https://user-images.githubusercontent.com/46744543/88133085-d0568f00-cc13-11ea-9a69-8a78e740e830.png)
此处s是否可以添加 synchronized ()修饰符,当然,使用ConcurrentHashMap就没有必要了

velo:
Vixi Mano, assim fica complicado o bagulho, tá ligado.

kdavisk6:
For the English speakers here, the original request is asking why this loop, which is present in `FieldMapQueryEncoder` is not synchronized, implying the thread safety of the serialization is suspect.  It is true that the way the `FieldMapQueryEncoder` is currently written, access to the underlying cached map is not thread-safe, however since access to the map is not "leaked", it is still fine to use.

To correct this, we would need to apply atomic get-then-set semantics and switch the cache to a `ConcurrentHashMap`

FedericoAlvarez:
Hello there, 
i would like to help on this as part as the hackoctoberfest!
Any extra information to start?

Cheers,

FedericoAlvarez:
I've created this pr please let me know if is ok :)
https://github.com/OpenFeign/feign/pull/1290

PR #1290, opened 2020-10-17
Make Field query map encoder thread safe

FedericoAlvarez:
#hacktoberfest

### DIFF
--- a/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
+++ b/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
@@ -13,12 +13,18 @@
  */
 package feign.querymap;
 
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
 import feign.Param;
 import feign.QueryMapEncoder;
 import feign.codec.EncodeException;
-import java.lang.reflect.Field;
-import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * the query map will be generated using member variable names as query parameter names.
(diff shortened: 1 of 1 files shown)


core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
```diff
@@ -31,16 +37,20 @@
...
   private final Map<Class<?>, ObjectParamMetadata> classToMetadata =
-      new HashMap<Class<?>, ObjectParamMetadata>();
+      new ConcurrentHashMap<>();
 
   @Override
   public Map<String, Object> encode(Object object) throws EncodeException {
     try {
-      ObjectParamMetadata metadata = getMetadata(object.getClass());
-      Map<String, Object> fieldNameToValue = new HashMap<String, Object>();
+      AtomicReference<Object> atomicReference = new AtomicReference<>();
```
velo, line 45: I don't understand how the `atomicReference` is going to help here...

velo (changes requested):
Can we also get some unit tests related to the change?

core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
kdavisk6: Seconded.  This is not necessary here.  All that is required to make this thread safe is to make `ObjectParamMetadata` immutable, since that's where the information can escape.

FedericoAlvarez: Hello there!
Sorry for the delay. The motivation of this pr is this https://github.com/OpenFeign/feign/issues/1257.

And the use of the atomic reference was because of this comment https://github.com/OpenFeign/feign/issues/1257#issuecomment-697046816 maybe i mis understood.

Cheers,

core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
```diff
@@ -31,16 +37,20 @@
...
+      atomicReference.set(object);
+
+      ObjectParamMetadata metadata = getMetadata(atomicReference.get().getClass());
+
+      Map<String, Object> fieldNameToValue = new ConcurrentHashMap<>();
       for (Field field : metadata.objectFields) {
-        Object value = field.get(object);
-        if (value != null && value != object) {
+        Object value = field.get(atomicReference.get());
+        if (value != null && value != atomicReference.get()) {
```
kdavisk6, line 53: Instead of an `AtomicReference` you can use the concurrent hash maps `putIfAbsent`.  This will provide the atomic get/then/put we need without the reference.

kdavisk6 (commented):
@FedericoAlvarez, see my latest comment.  What was meant in the issue that we need an atomic operation, not an `AtomicReference`

core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
rage-shadowman: `ObjectParamMetadata` is immutable. It uses a `final` "unmodifiable" list. The only issue I can see with the old code is that there could be multiple calls to generate the same metadata object at the same time (although only 1 will end up in the map) which a later PR fixes nicely.

FYI - When I wrote the old `QueryMapEncoder`, I felt that lower sync overhead later was worth a one-time possible multiple initialization slowdown.

Issue #1249, opened 2020-07-12
Should the documentation have an example for how to use multiple decoders?


sshevlyagin:
In my implementation I need to use the default decoder for `byte[]` and GsonDecoder to map to objects and I'm doing the following in Java 11:

`
@Bean
    public Decoder decoder() {
        Decoder decoder = (response, type) -> {
            if (type.getTypeName().equals("java.lang.String") ||
                    type.getTypeName().equals("byte[]")
            ) {
                return new Decoder.Default().decode(response, type);
            } else {
                return new GsonDecoder().decode(response, type);
            }
        };
        return new ResponseEntityDecoder(decoder);
    }
`

Is this worth having an example in the documentation? I was inspired by this discussion https://github.com/spring-cloud/spring-cloud-netflix/issues/2246

kdavisk6:
@sshevlyagin 

The example you provided is something that I think is a little too specific for our general documentation.  We may want to consider it part of our Wiki instead.  Thoughts?

Klapsa2503:
I agree, this should be a part of Wiki, code snippet was very useful for me

sshevlyagin:
@kdavisk6 - happy to make it a wiki article. Can I have permission to edit or would you like to do it? I tried pushing up the attached file but I don't have the rights. [wiki gist](https://gist.github.com/sshevlyagin/e5565a0d4f7537e69c82023e76097e36)

velo:
@sshevlyagin lemme check how to do that, one sec

velo:
@sshevlyagin seems it's a all or nothing situation regarding to the wiki... either is public to anyone to edit or not...

@kdavisk6 what do you think about giving @sshevlyagin write access?

velo:
@sshevlyagin what if we added a new `#decoder(MediaType,Decoder)` to `Feign.Builder`

Invoking `decoder(Decoder)` would overwrite the decoder as is today.
Invoking `decoder(MediaType,Decoder)` would replace `decoder` with a `TypedDecoder`, and would assign the existing `decoder` to media-type `*/*`
Subsequent invocations of `decoder(MediaType,Decoder)` would append a new `Decoder` to the `TypedDecoder`.
If `MediaType` already registered on `TypedDecoder` log warning and replace.

`TypedDecoder` would be a implementation of `Decoder` that checks the `MediaType` first and then delegate the decoding process to `Decoder`

sshevlyagin:
[...]

PR #1293, opened 2020-10-24
Introducing typed decoders

velo:
Possible fix to https://github.com/OpenFeign/feign/issues/1249


kdavisk6:
I wonder if it would be better to make all of the Encoders and Decoders generic, that way, they would be typed by default.  Thoughts?

velo:
> 
> 
> I wonder if it would be better to make all of the Encoders and Decoders generic, that way, they would be typed by default. Thoughts?

Somehow I missed your comment.... 

Yeah, we could totaly do it, it won't break compatibility

Issue #1270, opened 2020-09-10
Feign HeaderTemplates adds an extra space to date headers after commas 


muhamadto:
[HeaderTemplates#expand(Map<String, ?>)](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/template/HeaderTemplate.java#L142) adds space after each comma in headers. This causes an issue when converting date that uses the `EEE, d MMM yyyy HH:mm:ss Z` format.

Example
Client code sends `Wed, 4 Jul 2001 12:08:56 -0700`
Provider expects   `Wed, 4 Jul 2001 12:08:56 -0700`
Provider receives  `Wed,  4 Jul 2001 12:08:56 -0700` please note the extra space between the `,` and `4`

kdavisk6:
@muhamadto Can you please provide a minimal working example that demonstrates this issue?

gtay003:
We've just been hit by this issue too, it seems to affect any expanded header with a comma. Example code attached. The log output shows that the provided date value has an additional space inserted after the comma:

```
12:34:33.537 [main] DEBUG feign.Logger - [SampleApi#run] ---> GET https://reqres.in/api/users?page=2 HTTP/1.1
12:34:33.538 [main] DEBUG feign.Logger - [SampleApi#run] Date: Thu,  24 Sep 2020 10:34:09 GMT
12:34:33.538 [main] DEBUG feign.Logger - [SampleApi#run] ---> END HTTP (0-byte body)
```

The issue was likely caused by this change: https://github.com/OpenFeign/feign/pull/1203

Downgrading to 10.9 (last version before this change was made) has fixed for us for now.

[feign-header-bug.zip](https://github.com/OpenFeign/feign/files/5275672/feign-header-bug.zip)

jjz921024:
hi, i made pr #1295 for fix this issue, but CI build failed because of  `The command "./travis/no-git-changes.sh" exited with 1`

i don't know how to make it build successfully

gtay003:
> hi, i made pr #1295 for fix this issue, but CI build failed because of `The command "./travis/no-git-changes.sh" exited with 1`
> 
> i don't know how to make it build successfully

Looks like it's complaining about code formatting; did you see this in the build output?

`Please run 'mvn clean install' locally to format files.` 

At a guess I'd say it's because you have two blank lines above your new unit test method, and the formatter expects only one. But I guess `mvn clean install` will fix this for you.

jjz921024:
[...]

PR #1298, opened 2020-10-29
Fixed extra space to headers after commas

jjz921024:


jjz921024:
hi all, i fix a bug about issut #1270,  but I'm not sure I'm doing it right. there are anyone can review my code? thank

kdavisk6:
Replaced by #1347

PR #1302, opened 2020-11-02
Use charset from response Content-Type header to decode 

jaimepineiro:
Sometimes charset from response is not UTF-8, then if charset is informed in Content-type header, this charset is used to decode


jackson/src/main/java/feign/jackson/JacksonDecoder.java
```diff
@@ -66,4 +69,23 @@ public Object decode(Response response, Type type) throws IOException {
       throw e;
     }
   }
+
+  private Charset getResponseCharset(Response response) {
```
kdavisk6, line 73: I recommend moving this to `Response#getCharset()`.  This way it can be used by all decoders.

jackson/src/main/java/feign/jackson/JacksonDecoder.java
```diff
@@ -47,7 +49,8 @@ public JacksonDecoder(ObjectMapper mapper) {
   public Object decode(Response response, Type type) throws IOException {
     if (response.body() == null)
       return null;
-    Reader reader = response.body().asReader(Util.UTF_8);
+    Charset charset = getResponseCharset(response);
+    Reader reader = response.body().asReader(charset);
```
kdavisk6, line 53: If you follow my additional advice below, then this could be refactored to `response.body().asReader()` where the implementation would use the charset included in the response header by default.  The result would be two methods `body().asReader()` and `body.asReader(Charset)`

kdavisk6 (changes requested):

jackson/src/main/java/feign/jackson/JacksonDecoder.java
jaimepineiro: I agree, but maybe the method name should be Response#charset(), to follow naming convention in Response (Response#body(), Response#headers(), etc)

jackson/src/main/java/feign/jackson/JacksonDecoder.java
jaimepineiro: Now the method asReader() is deprecated, should i remove deprecation?

kdavisk6 (approved):

PR #1307, opened 2020-11-11
Java11client connect timeout

### LINKED ISSUE #1306: java11 client should consider Options::connectTimeoutMillis
feign.http2client.Http2Client seems not to consider Options::connectTimeoutMillis. In the course of #1233 request timeout was fixed - but not the connect timeout. 

Implementation note: this would have to be set on the HttpClient instance (e.g. via Builder::connectTimeout). One could do this manually by providing a custom built HttpClient to Http2Client. However it would be good to leave this over to Http2Client when it constructs the HttpClient for us. E.g. by replacing the default constructor with a constructor taking Options or (IMHO better) lazily instantiating HttpClient - if no custom was specified via constructor - when the request should be sent.

gb-klein:
Java11 client (feign.http2client.Http2Client) should consider Options::connectTimeoutMillis

Problem: Http2Client does not consider connectTimeoutMillis from Options (and followRedirects as well). 
Solution: 
- on request execution check if the provided Options differ from the current client's timeout and redirect settings. If yes => create a new client from the existing one, but with timeout and redirect settings from Options.
- Additionally provide a constructor taking Options. This will decrease the probability of client configuration mismatches and thus increase performance

fixes #1306

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -47,6 +47,12 @@ public class Http2Client implements Client {
         .build());
   }
 
+  public Http2Client(Options options) {
+    this(newClientBuilder(options)
+        .version(Version.HTTP_2)
+        .build());
+  }
+
   public Http2Client(HttpClient client) {
     this.client = Util.checkNotNull(client, "HttpClient must not be null");
   }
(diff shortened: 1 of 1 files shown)


kdavisk6 (approved):
Thanks for the work on this one.

PR #1309, opened 2020-11-12
Allows @Param value in default contract to be taken from param names.

### LINKED ISSUE #1297: Support JEP 118 parameters
I haven't seen this issue, but it is quite hard to search for this so sorry for duplicate.

Java 8 introduced an additional javac option in ([JEP 118](http://openjdk.java.net/jeps/118)) using `-parameters` to allow retrieval of parameter names during runtime.

This retrieval is supported in e.g. Spring and Jackson to reduce amount of duplication, so I think it would be good to have it in feign also.

E.g. I have following method in interface:
```
    @Body("{content}")
    @Headers(value = {"Content-Type: application/json", "Accept: application/json"})
    @RequestLine("POST /prop/{property}/ver?cont={contact}&addr={address}")
    String createNewVersion(@Param("property") String property, @Param("contact") String contact, @Param("address") String address, @Param("content") Map<String, Integer> content);
```

As you see the same name is both in the parameter name as well as in annotation, if `-parameters` would be supported this could be simplified to:
```
    @Body("{content}")
    @Headers(value = {"Content-Type: application/json", "Accept: application/json"})
    @RequestLine("POST /prop/{property}/ver?cont={contact}&addr={address}")
    String createNewVersion(@Param String property, @Param String contact, @Param String address, Map<String, Integer> content);
```

Currently (in 9.7.0) it is not supported because `value` in `@Param` annotation is required.

krzyk:
JEP 118 introduced javac option `-parameters` that adds to bytecode
method parameter names, so if code is compiled with that option
and @Param value is empty we can get that template parameter name
from method parameter name.

Fixes #1297.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ should work.  Feign's default contract defines the following annotations:
 | Annotation     | Interface Target | Usage |
 |----------------|------------------|-------|
 | `@RequestLine` | Method           | Defines the `HttpMethod` and `UriTemplate` for request.  `Expressions`, values wrapped in curly-braces `{expression}` are resolved using their corresponding `@Param` annotated parameters. |
-| `@Param`       | Parameter        | Defines a template variable, whose value will be used to resolve the corresponding template `Expression`, by name. |
+| `@Param`       | Parameter        | Defines a template variable, whose value will be used to resolve the corresponding template `Expression`, by name provided as annotation value. If value is missing it will try to get the name from bytecode method parameter name (if the code was compiled with `-parameters` flag). |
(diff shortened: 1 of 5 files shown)


kdavisk6 (changes requested):
Simple enough.  Thank you for this.  If you could please update the `Readme` with this documentation?

krzyk:
@kdavisk6 I've updated the README.md

kdavisk6 (approved):

krzyk:
@kdavisk6 Can this be merged? Builds are passing and I would like to use it in my project

PR #1311, opened 2020-11-14
Support metric names customization

### LINKED ISSUE #1300: Support different names for micrometer metrics
### Description
It is not possible to override standard `feign.Client` and similar metric names. In order to have similar metric names not depending on client implementation (e.g. if it's Feign or Spring WebClient for instance) it should be possible to define different metric names for `MeteredClient`, `MeteredEncoder` and similar classes.

```
public MeteredEncoder(Encoder encoder, MeterRegistry meterRegistry) {
    this.encoder = encoder;
    this.meterRegistry = meterRegistry;
    this.metricName = new FeignMetricName(Encoder.class); // basically, hard-coded
  }
```
### Solution
Allow passing third parameter defining metric name. FeignMetricName may become interface with different implementations, e.g. class, string etc.

Kuvaldis:
Currently, it's not possible to use custom names for micrometer metrics without having to almost completely reimplement existing Feign Micrometer classes. 
Different clients (e.g. Spring WebClient) may define different names like `http.client.request`. Same name can be used for Feign Clients.
Intention is to allow developers to define their own implementations of `MetricName` for `MeteredClient` and other related classes.

Fixes #1300


kdavisk6 (approved):
Thanks for this update.

PR #1322, opened 2020-12-04
Micrometer metrics tags extension

### LINKED ISSUE #1301: Support uri and exception tags in micrometer MeteredClient
### Description

In `MeteredClient` implementation there is a try/catch block:
```
try {
      return meterRegistry.timer(
          metricName.name(),
          metricName.tag(template.methodMetadata(), template.feignTarget()))
          .recordCallable(() -> client.execute(request, options));
    } catch (IOException e) {
      throw e;
    } catch (Exception e) {
      throw new IOException(e);
    }
```

A couple of things might be improved here:
* `exception` and `uri` tags would be really helpful as part of `feign.Client` metric.
* IOException wrapping prohibits calling method to catch original exception. Seems to be unnecessary to do.

### Proposal

* Add extensions (e.g. as abstract methods) to `MeteredClient` and similar classes allowing to add custom tags.
* Remove `IOException` wrapping.

Kuvaldis:
Following Micrometer features added:
1. Add timer metrics to requests resulting with exception. E.g. in MeteredClient. Includes exception name as tag.
2. Add uri as default tag in MeteredClient metrics.
3. Support adding custom tags in metered client, encoder, decoder etc.

Fixes #1301


kdavisk6:
@Kuvaldis looks like I may have merged your pull requests out of order.  Can you check this one?

Kuvaldis:
@kdavisk6 All done.

kdavisk6 (approved):
@Kuvaldis

This looks OK to me.  I'd like @velo to weigh in also.  I'll mark it ready and allow for some additional comments before we merge.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -13,15 +13,15 @@
...
-import feign.Request;
+import feign.*;
 import feign.Request.Options;
-import feign.RequestTemplate;
-import feign.Response;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import java.io.IOException;
```
velo, line 22: it would be nice not to change import ordering... but, don't worry too much about it =)

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -45,38 +45,82 @@ public MeteredClient(Client client, MeterRegistry meterRegistry, MetricName metr
...
+    return meterRegistry.timer(metricName.name("exception"), allTags);
+  }
+
+  protected List<Tag> extraSuccessTags(RequestTemplate template, Options options) {
+    final List<Tag> result = new ArrayList<>();
+    result.add(Tag.of("uri", template.path()));
+    return result;
+  }
+
+  protected List<Tag> extraExceptionTags(RequestTemplate template,
```
velo, line 118: I think we should join together the "success" and "exception"  methods...  I think is ok to have the last argument nullable and only add the tag when needed

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -45,38 +45,82 @@ public MeteredClient(Client client, MeterRegistry meterRegistry, MetricName metr
...
     } catch (Exception e) {
+      sample.stop(createExceptionTimer(template, options, e));
       throw new IOException(e);
     }
   }
 
+  private void countSuccessResponseCode(RequestTemplate template,
+                                        Options options,
+                                        int responseStatus) {
+    final List<Tag> tags = extraSuccessTags(template, options);
```
velo, line 70: I think code would be a bit better to read if "extraTags" returned "Tag[]"

```
metricName.tag(template.methodMetadata(), template.feignTarget())
  .and(extraTags(...))
  .and("xyz", 10);
```

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -45,38 +45,82 @@ public MeteredClient(Client client, MeterRegistry meterRegistry, MetricName metr
   @Override
   public Response execute(Request request, Options options) throws IOException {
     final RequestTemplate template = request.requestTemplate();
-
+    final Timer.Sample sample = Timer.start(meterRegistry);
```
velo, line 48: can we also have a `finally` block, just in case?

micrometer/src/main/java/feign/micrometer/MeteredDecoder.java
```diff
@@ -46,41 +49,81 @@ public MeteredDecoder(Decoder decoder, MeterRegistry meterRegistry, MetricName m
...
+
+  protected List<Tag> extraExceptionTags(Response response, Type type, Exception exception) {
+    final RequestTemplate template = response.request().requestTemplate();
+    final List<Tag> result = new ArrayList<>();
+    result.add(Tag.of("uri", template.path()));
+    result.add(Tag.of("exception_name", exception.getClass().getSimpleName()));
+    return result;
+  }
+
+  protected List<Tag> extraSummaryTags(Response response, Type type) {
```
velo, line 125: I assume you are using this for some internal metrics (I would be curious to know what you are going =] )

I would prefer to see code to live in the `MetricName` interface... that would give a single place for all your tag customization needs.

We would have a few methods, one called `requestExtraTags` other one `responseExtraTags`, another like `exceptionExtraTags`

What do you think?

velo (changes requested):
Overall, I think is a good improvement.

Things I would like to see
a) Not sure if make sense or no, but `finally` added to all try/catch blocks.
b) use `Tag[]` or `Tags` instead of `List<Tag>` + `toArray(new Tag[] {})` combo
c) move metric naming to `MetricName` giving people one place for all customization needs.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -45,38 +45,82 @@ public MeteredClient(Client client, MeterRegistry meterRegistry, MetricName metr
...
+    tags.add(Tag.of("status_group", responseStatus / 100 + "xx"));
+    final Tags allTags = metricName
+        .tag(template.methodMetadata(), template.feignTarget(), tags.toArray(new Tag[] {}));
+    meterRegistry.counter(
+        metricName.name("http_response_code"),
+        allTags)
+        .increment();
+  }
+
+  private Timer createSuccessTimer(RequestTemplate template, Options options) {
```
velo, line 96: I really like the idea of moving Timer creation (and all other metrics) creation to methods...

Could you please make them `protected`, if anyone has a need for a really complex customization, they would be able to do it here

Kuvaldis: Done.

micrometer/src/main/java/feign/micrometer/MeteredDecoder.java
Kuvaldis: I extracted tags resolution to a separate extendable class. However, seems like depending on where we create tags (i.e. client, encoder/decoder, method handler) we have really different parameters to build tags upon. I added defaultTags method, which might be helpful for customization in general.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
Kuvaldis: Done.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
Kuvaldis: Done.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
Kuvaldis: Done.

Kuvaldis:
> c) move metric naming to `MetricName` giving people one place for all customization needs.

Basically, there is `String name(String suffix)` should be enough, depending on suffix it can be overriden, for example. Another option is to create a separate method for each metric, but I think it is too specific and not really required.

velo (approved):

r1jsheth:
is there any example depicting this use case?

Issue #1339, opened 2020-12-24
feign.template.HeaderTemplate#expand repaces ',' to ', ' (An extra space after commas)


gdut-yy:
https://github.com/OpenFeign/feign/blob/04bd147fba11bdf1cac26aba1d64d189e0867a3f/core/src/main/java/feign/template/HeaderTemplate.java#L151

[spring-cloud/spring-cloud-openfeign](https://github.com/spring-cloud/spring-cloud-openfeign) >= `2.2.3.RELEASE` was affected.

This will cause some APIs authentication to fail! 

Example 

SimpleDateFormat `EEE, d MMM yyyy HH:mm:ss 'GMT'`

```
"Wed, 23 Dec 2020 02:34:12 GMT"
```

was replaced by

```
"Wed,  23 Dec 2020 02:34:12 GMT"
```

The latest version still exists, is there any plan to fix it?

![image](https://user-images.githubusercontent.com/33390928/103065822-8292e800-45f2-11eb-8e35-978c05b4b6dd.png)

gdut-yy:
@kdavisk6 Hi, Davis. This issue was likely caused by this commit: 2b0bf0060d3734a787882fb8b43b52288b1fe3ca

We need to update the open source component versions regularly for security reasons, but this issue is holding us back.

For more information: [https://github.com/spring-cloud/spring-cloud-openfeign/issues/450](https://github.com/spring-cloud/spring-cloud-openfeign/issues/450)

silkentrance:
@gdut-yy why would you pass in such a datetime string when there are other ISO formats that can be parsed much easier?

silkentrance:
@gdut-yy for my projects I am using the "standard" iso format that is provided by `https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME`

kdavisk6:
The date format here is the HTTP Date format used in the HTTP Spec.  It is different from the ISO specifications.

gdut-yy:
@silkentrance Your suggestion is good. But if I could change the API's signature, I wouldn't be open this issue. In our actual development, some of the APIs come from other teams or companies. For various reasons, they cannot be changed easily.

gdut-yy:
@kdavisk6 When add new features, we should not modify existing code to introduce new "bugs". I have two solutions without modifying the API's signature:

[...]

PR #1344, opened 2020-12-29
Fix for #1339 (An extra space after commas)

gdut-yy:
feign.template.HeaderTemplate#expand repaces ',' to ', ' (An extra space after commas)


kdavisk6:
Replaced by #1347

Issue #1325, opened 2020-12-07
ErrorDecoder cannot get the real response's charset


Linda-pan:
Code FeignException in feign-core-10.10.1

```java
 private static Charset getResponseCharset(Map<String, Collection<String>> headers) {

      Collection<String> strings = headers.get("content-type");
      if (strings == null || strings.size() == 0) {
        return null;
      }

      Pattern pattern = Pattern.compile("charset=([^\\s])");
      Matcher matcher = pattern.matcher(strings.iterator().next());
      if (!matcher.lookingAt()) {
        return null;
      }

      String group = matcher.group(1);
      if (!Charset.isSupported(group)) {
        return null;
      }
      return Charset.forName(group);

    }
```

It seems that matcher.group(1) can only get one charactor, just like "UTF-8" can only get 'U'.

then all charset will become the default value utf-8.

```java
   private static String getBodyAsString(byte[] body, Map<String, Collection<String>> headers) {
      Charset charset = getResponseCharset(headers);
      if (charset == null) {
        charset = Util.UTF_8;
      }
      return getResponseBody(body, charset);
    }
```

Actually, content-type not only contain "charset", it usually like "application/json;charset=UTF-8".
I suggess that pattern could be ".*charset=([^\\s|^;]+).*"。
or use “matcher.find()” with  "charset=([^\\s|^;]+)"

kdavisk6:
@Linda-pan  Feel free to submit a PR to correct this.

Linda-pan:
@kdavisk6 Sorry for my late reply, I will submit a PR as soon as possible.

PR #1345, opened 2020-12-29
Fix bug for FeignException cannot get the correct charset (#1325)

Linda-pan:
Fix bug for FeignException cannot get the correct charset(#1325)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,10 @@ commands:
   nexus-deploy-jdk11:
     steps:
       - run:
+          name: 'Build JDK 11 Release modules locally'
+          command: |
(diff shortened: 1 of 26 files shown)


kdavisk6 (changes requested):
@Linda-pan Please add a unit test for this use case.

kdavisk6:
@Linda-pan Your latest changes are not formatted correctly, please fun `mvn install` locally to format the files, commit the changes and push the updates to fix the build

Linda-pan (commented):
Fix bug for FeignException cannot get the correct charset and add some tests

kdavisk6 (approved):

PR #1351, opened 2020-12-30
connection request timeout setting documentation. For #1318

### LINKED ISSUE #1318: Can't config connection request timeout for http client
There is a field named 'connectionRequestTimeout' in RequestConfig.java of http client, but I can't config it from openfeign.

- I don't find a property to config it from FeignHttpClientProperties.java
- I don't find a property to config it from Request.Options 

When a large number of requests are sent through openfeign, If we don't config this property, the connection pool is exhausted, there will be a large number of requests waiting for connections, and no timeout will occur. This will deplete the tomcat threads. 

Can you tell us how to config this property? Thanks!

youngxinler:
Request timeout configuration is not documented.

Add this documentation into README , example-github and example-wikipedia.

fixes #1318

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,10 @@ with the following alterations:
 ---
 ### Customization
 
-Feign has several aspects that can be customized.  For simple cases, you can use `Feign.builder()` to construct an API interface with your custom components.  For example:
+Feign has several aspects that can be customized.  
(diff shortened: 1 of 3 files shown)


kdavisk6 (approved):

kdavisk6:
@youngxinler Thank you for the work!

Issue #1361, opened 2021-02-02
Duplicated headers with Request Template


efres:
# Problem
When we use the request interceptor feign to set headers on runtime and we set the same header key with different values, the previous header will not be overwritten. This could cause business problems when the token is refreshed and the request persist trying to access the rest endpoint.

# Cause

Call appendHeader RequestTemplate.java:708 
```java

  /**
   * Specify a Header, with the specified values. Values can be literals or template expressions.
   *
   * @param name of the header.
   * @param values for this header.
   * @return a RequestTemplate for chaining.
   */
  public RequestTemplate header(String name, Iterable<String> values) {
    if (name == null || name.isEmpty()) {
      throw new IllegalArgumentException("name is required.");
    }
    if (values == null) {
      values = Collections.emptyList();
    }

    return appendHeader(name, values); // Line 708
  }
```
Then not verify header presence before appending him. RequestTemplate.java:732

```Java

  /**
   * Create a Header Template.
   *
   * @param name of the header
   * @param values for the header, may be expressions.
   * @return a RequestTemplate for chaining.
   */
  private RequestTemplate appendHeader(String name, Iterable<String> values) { // Line 732
    if (!values.iterator().hasNext()) {
      /* empty value, clear the existing values */
      this.headers.remove(name);
      return this;
    }

    if (name.equals("Content-Type")) {
      // a client can only produce content of one single type, so always override Content-Type and
      // only add a single type
      this.headers.remove(name);
      this.headers.put(name,
          HeaderTemplate.create(name, Collections.singletonList(values.iterator().next())));
      return this;
    }
    this.headers.compute(name, (headerName, headerTemplate) -> {
      if (headerTemplate == null) {
        return HeaderTemplate.create(headerName, values);
      } else {
        return HeaderTemplate.append(headerTemplate, values);
      }
    });
    return this;
  }
```

# Proposed Solution
Should be implemented validation according to "Content-Type" validation.  PR Open on this [link](https://github.com/OpenFeign/feign/pull/1362)

kdavisk6:
@efres 

This is intentional as the function is designed to support the building of multi-value headers over multiple steps within the request building process.  From your example, the correct way to replace a header values is to first remove it, then append the new values.

To replace header values, you must use the `removeHeader` method to clear any existing header values, then use the `appendHeader` method.

```java
/* replace an entire header */
template.removeHeader("header");
template.appendHeader("header", values);
```

[...]

PR #1362, opened 2021-02-02
feat: added validation for duplicated keys on header

efres:
# What: 
Problem described on [Issue 1361](https://github.com/OpenFeign/feign/issues/1361)

# Why:
To fix integration and requests problems with the interceptor.

# Action:
Added validation to remove the previous header set.

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -736,7 +736,7 @@ public final class RequestTemplate implements Serializable {
       return this;
     }
 
-    if (this.headers.containsKey(name)) {
+    if (headerIsPresentOnHeaders(name)) {
       // a client can only produce content of one single type, so always override an present header and
       // only add a single type
       this.headers.remove(name);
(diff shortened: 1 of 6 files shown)


evertonsegurzup:
(RFC-2616)[https://tools.ietf.org/html/rfc2616#section-4.2]
```
Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.
```

evertonsegurzup:
[RFC-7230](https://tools.ietf.org/html/rfc7230#section-3.2.2)

```
A sender MUST NOT generate multiple header fields with the same field name in a message unless either the entire field value for that header field is defined as a comma-separated list [i.e., #(values)] or the header field is a well-known exception (as noted below).

A recipient MAY combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value to the combined field value in order, separated by a comma. The order in which header fields with the same field name are received is therefore significant to the interpretation of the combined field value; a proxy MUST NOT change the order of these field values when forwarding a message.
```

kdavisk6:
@efres Please see my comments on this issue.  This is by design and not a bug.

efres:
@kdavisk6 ok. I'll answer there. Thanks

PR #1369, opened 2021-02-24
Make the FieldQueryMapEncoder encoder thread safe

### LINKED ISSUE #1257: FieldQueryMapEncoder may not be thread-safe

Budlee:
Change the caching map to be concurrent and only
insert items that are missing to the map.
Fixes #1257

### DIFF
--- a/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
+++ b/core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2020 The Feign Authors
+ * Copyright 2012-2021 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -13,12 +13,17 @@
  */
 package feign.querymap;
 
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 import feign.Param;
 import feign.QueryMapEncoder;
 import feign.codec.EncodeException;
-import java.lang.reflect.Field;
-import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * the query map will be generated using member variable names as query parameter names.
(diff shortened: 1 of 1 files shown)


core/src/main/java/feign/querymap/FieldQueryMapEncoder.java
```diff
@@ -56,7 +61,7 @@ private ObjectParamMetadata getMetadata(Class<?> objectType) {
     ObjectParamMetadata metadata = classToMetadata.get(objectType);
     if (metadata == null) {
       metadata = ObjectParamMetadata.parseObjectType(objectType);
-      classToMetadata.put(objectType, metadata);
+      classToMetadata.putIfAbsent(objectType, metadata);
```
rage-shadowman, line 64: Could this whole method be replaced with:
```
return classToMetadata.computeIfAbsent(objectType, ObjectParamMetadata::parseObjectType);
```
???

Budlee: I think it could,
I wasn’t sure on the refactor culture. Some like to make as few changes as possible.

happy to add this in

Budlee: Updated, 
the other change I thought was to the core of this loop to be the following, please let me know if its okay as I woulduse a pair:

```java
return metadata.objectFields.stream()
            .map(this::pairFieldAndOptionalValue)
            .filter(pair -> pair.right.isPresent())
            .collect(Collectors.toMap(pair -> createKeyFromField(pair.left), pair -> pair.right));
```

Its much smaller (personally I think it  reads better), however, I  understand  not to everyones  taste as some do not like the use of pair and also pairFieldAndOptionalValue is a hidden side effect where if the field.get fails will throw then EncodeException

Budlee: @rage-shadowman what are thoughts on this refactor?

kdavisk6: I think it all depends on how `field.get` fails.  The current loop is vulnerable to that now.  I don't have any issue with catching there and wrapping in an `EncodeException`, that fits with the `Encoder` interface, so it's not a hidden side-effect, so long as you add some comments around the check.

rage-shadowman: If you're asking my opinion, I think that any time you simplify the code without breaking anything, you make it better. I think this PR looks more simplified.

As far as that stream usage goes, it looks more difficult to read at-a-glance to me, but the original isn't that easy to read at-a-glance either, and I tend to be biased against all but the simplest of streams usage.

As long as your stream doesn't modify anything from the input `metadata`, or from one of the earlier `map()`d records, then there aren't any side effects. Throwing a runtime exception due to an unexpected illegal state sounds reasonable.

Budlee:
@rage-shadowman and @kdavisk6  I have updated this  now,
Personally I think it is more readible, however, can revert it back if required.

thanks

kdavisk6:
@Budlee Done is better than perfect here.

kdavisk6 (approved):

Budlee:
@kdavisk6 , when does this get merged?

PR #1375, opened 2021-03-12
Collect `http_response_code` for successfull and failed requests

velo:
On previous implementation response code would only be collected for errors.

To avoid mixing new and old metrics, gave the one that includes successfull codes a different name

### DIFF
--- a/dropwizard-metrics4/pom.xml
+++ b/dropwizard-metrics4/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2012-2020 The Feign Authors
+    Copyright 2012-2021 The Feign Authors
 
     Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
     in compliance with the License. You may obtain a copy of the License at
@@ -48,6 +48,12 @@
       <groupId>org.hamcrest</groupId>
       <artifactId>java-hamcrest</artifactId>
       <version>2.0.0.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>feign-micrometer</artifactId>
+      <type>test-jar</type>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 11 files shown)


kdavisk6 (commented):
@velo 

Should we consider using Timers instead of Counters?  Timer implicitly count and can provide more information.  Thoughts?

velo:
We already have timers for method handler, client, encoder and decoder.

At the time we start the timers we don't know the response code, so can't
include that piece of information on timers

kdavisk6:
Ahh, but Micrometer you can

```java
long start = System.nanoTime();  // record the start time.
Response response = client.execute(Request); // execute the request
registry.timer(metric.name).tag(getTags(response)).record(System.nanoTime() - start) // create a new timer with the time calculated and use getTags() to read the response and set the Success or Failre
```

This way, you can include a `tag` that indicates if the request was successful or not.  You can use this tag in your metrics system to calculate success and failure rates, without needed a separate counter.  Not sure how to do it with Dropwizard.

velo:
> Ahh, but with Micrometer you can.
> 
> ```java
> long start = System.nanoTime();  // record the start time.
> Response response = client.execute(Request); // execute the request
> registry.timer(metric.name).tag(getTags(response)).record(System.nanoTime() - start, TimeUnit.NANOSECONDS) // create a new timer with the time calculated and use getTags() to read the response and set the Success or Failure
> ```
> 
> This way, you can include a `tag` that indicates if the request was successful or not. You can use this tag in your metrics system to calculate success and failure rates, without needed a separate counter. Not sure how to do it with Dropwizard.

Yeah, you can do "exactly" the same on dropwizard metrics....  but ...

```
      return meterRegistry.timer().recordCallable(() -> {...});
```

Looks much better than

```
long start = System.nanoTime();
try {
   Response response = delegate.execute();
   meterRegistry.timer( tag(response.getStatus()) ).record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
   return response;
} catch (FeignException e) {
   meterRegistry.timer( tag(e.getStatus()) ).record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
   throw e;
}
```

Also, if something outside FeignException happens, you won't get neither the timer nor the status code.

Do you have any specific use case for this or just an improvement?

velo:
@kdavisk6 also, do you mind if I merge as is, kick an release and if need be we rework this later?!

Cheers,

kdavisk6:
Could refactor with a  try/finally instead. Sounds like you need this for something you are working on. I’ll approve and think on it some more.

kdavisk6 (approved):

velo:
Awesome, thanks @kdavisk6 

I will make a new release for this...

shall we go with 11.1 (as I see people using 11.0 out there, or 10.x?

kdavisk6:
Yeah. Might as well fix the 11.x problem.

Issue #1385, opened 2021-04-21
Another pair of decoder and encoder for JSON-java 


radio-rogal:
I propose to add module e.g. feign-json that will provide decoder and encoder for JSONObject and JSONArray objects.

Related project https://github.com/stleary/JSON-java

If this proposal will be approved I could try to implement this module.

velo:
@radio-rogal I think this is very reasonable

FWIW, I join the feign project after having a bunch of small feign modules spread across github.

You can either start with something you fully own and control (release cycles can be much faster) and include on feign later on.


Anyway, feel free to make a PR and ping me when you think is ready to review.

PR #1387, opened 2021-04-22
Add JSON-java decoder and encoder

radio-rogal:
Added module for encoding and decoding using JSON-java.

Implements #1385


json/README.md
```diff
@@ -0,0 +1,15 @@
+JSON-java Codec
+===================
+
+This module adds support for encoding and decoding JSON via [JSON-java][].
```
velo, line 4: could you please add the link to https://www.json.org/json-en.html

json/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.github.openfeign</groupId>
+    <artifactId>parent</artifactId>
+    <version>-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>feign-json</artifactId>
```
velo, line 26: @kdavisk6 any opinion about the package/module names?  Just calling it "json" triggers something on the back of my mind, not that I have a better suggestion.

json/src/main/java/feign/json/JsonEncoder.java
```diff
@@ -0,0 +1,65 @@
...
+ *   github.create("openfeign", "feign", contributor);
+ * </pre>
+ */
+public class JsonEncoder implements Encoder {
+
+  @Override
+  public void encode(Object object, Type bodyType, RequestTemplate template)
+      throws EncodeException {
+    if (JSONArray.class == bodyType || JSONObject.class == bodyType) {
+      if (object == null)
```
velo, line 57: could you move this outside the first `if` block?!

No reason to not "work" on any nulls.

json/src/main/java/feign/json/JsonEncoder.java
```diff
@@ -0,0 +1,65 @@
...
+ *   contributor.put("contributions", 0);
+ *   github.create("openfeign", "feign", contributor);
+ * </pre>
+ */
+public class JsonEncoder implements Encoder {
+
+  @Override
+  public void encode(Object object, Type bodyType, RequestTemplate template)
+      throws EncodeException {
+    if (JSONArray.class == bodyType || JSONObject.class == bodyType) {
```
velo, line 56: wonder if we should instead use `object instanceof JSONArray || ... JSONObject`

also, it's possible to anyone to extend `JSONArray`. Worse case scenario would need to use `isAssignableFrom`

My preference is for instance of on the encoder

json/src/main/java/feign/json/JsonDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+        throw (IOException) jsonException.getCause();
+      }
+      throw new DecodeException(response.status(), jsonException.getMessage(), response.request(),
+          jsonException);
+    }
+  }
+
+  private Object decodeJSON(Response response, Type type, Reader reader) {
+    JSONTokener tokenizer = new JSONTokener(reader);
+    if (JSONObject.class == type)
```
velo, line 77: Anyone can extends JSONObject would be better to use `isAssignableFrom` here to make sure any subtypes would also work.

Please include inheritance on the codec test too.

velo (changes requested):
Excellent work @radio-rogal , I asked for a few changes, lemme know how it goes.

json/pom.xml
kdavisk6: Since the implementation here is `org.json`, I'm good with it.

radio-rogal:
@velo done

json/README.md
radio-rogal: +

json/pom.xml
radio-rogal: I have renamed it to `feign-org.json` and module to `org.json`.

json/src/main/java/feign/json/JsonEncoder.java
radio-rogal: done

json/src/main/java/feign/json/JsonEncoder.java
radio-rogal: replaced with instanceof

json/src/main/java/feign/json/JsonDecoder.java
radio-rogal: +

json/pom.xml
velo: Ow, no, feign-json was fine

radio-rogal: Sorry, misunderstood

I has returned back feign-json as artifact name and json as directory name.

radio-rogal: How about json-java or org.json as folder/mofule name? The artifact name is still feign-json.

I also did not like that the name is too common, but the artifact of the library, like its name (json-java), is also too general and I could not choose another name.

json/src/main/java/feign/json/JsonDecoder.java
velo: @radio-rogal can we get this done? As far I can tell, is the only thing preventing me from merging. I just merged master into your branch, so, it's almost ready

radio-rogal: @velo Sorry for late response, looks like I have missed this point.

I have changed it in 3d14686

velo (approved):

kdavisk6 (approved):

PR #1393, opened 2021-05-01
WIP: Add support for java 9

Sineaggi:
Attempt at fixing #935 via https://github.com/velo/feign/pull/4 and using multi-release jars.


.circleci/config.yml
```diff
@@ -160,45 +160,56 @@ workflows:
...
       - test:
-          jdk: '11'
+          jdk: 'circleci/openjdk:11'
           name: 'jdk 11'
           filters:
             <<: *all-branches
       - test:
-          jdk: '14-buster'
+          jdk: 'circleci/openjdk:14-buster'
           name: 'jdk 14'
```
velo, line 174: I think as we add jdk 16 "support" we could drop 14.

core/src/main/java9/feign/Capability.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Capabilities expose core feign artifacts to implementations so parts of core can be customized
+ * around the time the client being built.
+ *
+ * For instance, capabilities take the {@link Client}, make changes to it and feed the modified
+ * version back to feign.
+ *
+ * @see Metrics5Capability
+ */
+public interface Capability {
```
velo, line 33: What is the difference between regular `Capability.java` and java 9 copy of it?!

velo (changes requested):
Having two copies of the sources is not an option... as nobody should be expected to patch two locations every time something changes.

Why do we need that?

core/src/main/java9/feign/Capability.java
Sineaggi: Same goes for every other duplicate class, yes. I'm still working on the testing, which is the hardest part I'm finding out.

Sineaggi:
> Having two copies of the sources is not an option... as nobody should be expected to patch two locations every time something changes.
> 
> Why do we need that?

Honestly just so I could get to fixing the tests. I'll be rewriting history to remove all the duplicate classes that are unnecessary when I can get maven to stop complaining.

Sineaggi:
Unit tests are run via the compile output classpath, not via jar as per the [maven lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-lifecycle-is-made-up-of-phases) meaning the 'Multi-Release JAR Files' can't be used in tests. (As the multi-release jar doesn't exist until the package goal has run). That means we can't run java 16 unit-tests against this project, since java 16 doesn't like this code. (That's not entirely the truth, we could add `--illegal-access=warn --add-opens java.base/java.lang.invoke=ALL-UNNAMED` as flags to the tests to get them to pass, and then maybe add the failsafe plugin to integration test the jar after the fact, without those flags. Just an idea.) For now, I've removed the dependency on java 16 due to the added complexity around testing. This code really only needs to add the option to build against java9 as an additional supported release. In the future, I'd like to see maven's strategy around testing multi-release jar's improved, having to use integration-tests to test what seem to me to be unit's of code seems strange.

Sineaggi:
Not sure yet how to handle the jdk8 builds. They don't like the `--release`, and can't compile the java 9 source. Maybe a profile, or remove the java 8 build?

Sineaggi:
Looks like the build file currently releases most jars under java 8, and one jar under java 11. Wonder if it would be ok to move all the jars under 11, then use the `--release` to have java 11 cross compile to java 8.

core/src/main/java9/feign/DefaultMethodHandler.java
```diff
@@ -0,0 +1,73 @@
...
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Method;
+
+/**
+ * Handles default methods by directly invoking the default method code on the interface. The bindTo
+ * method must be called on the result before invoke is called.
+ */
+@IgnoreJRERequirement
+final class DefaultMethodHandler implements MethodHandler {
```
velo, line 28: Maybe we could add to https://github.com/OpenFeign/feign/tree/master/java11

I read mixed reviews of multi release support

Sineaggi: Multi release has pretty sparse documentation from what I've found. It seems strange that there's no built-in functionality that would let your run unit-tests against the multi-release classes. The support for running tests against jars only exists in the failsafe plugin.

It might be possible to make a ServiceLocator that can load the different MethodHandlers, but that seems like a lot of complexity to add when using multi-release jars feels like it was built for exactly this purpose.

velo: Yeah, having that on the `java11` module instead, would lead to much better testability as well.

Sineaggi: At least according to the gradle team, they don't recommend multi release jars according to [this blog post](https://blog.gradle.org/mrjars). On the other hand, there's just one class that needs to be updated for java 9+, and we don't need a new api surface to support it, or jdk11 variant jars either. (The java 11 module looks like it just adds support for the new httpclient, and doesn't replace any existing functionality.)

velo: Yeah, right now, java11 module only contain an http2 client.  But nothing prevent from adding a new package with a better `MethodHandler` implementation.

Does java16 work at all with existing `DefaultMethodHandler`?

Sineaggi: Yes, and it will continue to work, but it requires flags like `--illegal-access=warn --add-opens java.base/java.lang.invoke=ALL-UNNAMED` at runtime to be supported. Which can be used to make the tests pass on java 16, as I've tested it.

I'm still leaning more towards multi release modules, but if you think it's better for the project to use some other way to optionally use the new `final Lookup lookup = MethodHandles.privateLookupIn(declaringClass, MethodHandles.lookup());` based code, I'd still be down to do the work.

kdavisk6: You are on the right track by changing the `MethodHandles` to `privateLookupIn`, however, we still need the build to be backward compatible.  It will not be possible to have a special `DefaultMethodHandler` in a specific module.  The core library doesn't accept pluggable method handlers, so we'll need to first check if we have access to `privateLookupIn` and if we do, use it, if not, fall back to the JDK 8 way.  

Check out Lukas's approach to this: https://blog.jooq.org/2018/03/28/correct-reflective-access-to-interface-default-methods-in-java-8-9-10/

Before we can begin to implement this, we'll need to do the following:

1.  Change our CI builds to build and package using the latest supported JDK, targeting our lowest supported JDK.  In this case we would build using JDK 16, and set the compiler target to `-target:1.8`.
2. Change our CI workflow to **test** against JDK 16, 11, and 8.  This requires breaking the build and test jobs out and then running the test job using the multiple JDKs configured.
3. Make the appropriate changes to `DefaultMethodHandler` to try the JDK 9+ way, and if that fails, use the JDK 8 way.

We can make these changes safely because we use Animal Sniffer to force JDK 8 compatibility.  

I can start working on the workflow changes in a separate PR, but I'd like to get a better understanding of the urgency of this.  This is pretty big change, something we should consider for a major release only.

@Sineaggi, can you help me understand the urgency around this?

Sineaggi: I'm not totally sure what you mean by `we still need the build to be backward compatible`. The pr as it is should already be backwards compatible with java 8.

In order to support multi-release jars, we can use java 11 to compile everything; using the --release 8 flag for any non 11 jars, including the core one, the --release 9 flag for the one class in the multi-release jar, and then --release 11 for the java 11 specific jar. This will allow the core jar to run under java 8 just as it had been (it would still pull the old relfective `DefaultMethodHandler`) and under java 9+ the jvm runtime will choose to load the new `DefaultMethodHandler` class instead.

I don't think we need any clever fallback solution for 'trying the jdk 9 way, then falling back' since by virtue of the multi-release jar class only being loaded on jdk 9+, there won't be any need for runtime checks by this library.

Also I do agree the testing strategy will have to change.

Sineaggi: Also as an aside, I tried the ServiceLoader pattern here, https://github.com/OpenFeign/feign/compare/master...Sineaggi:add-privder-support?expand=1 and while it works and is easy enough to test, it also has a few drawbacks. For one, this doesn't solve the problem of being unable to run all the tests against java 16. It also requires users of feign to have the java11 jar on their classpath, which until now wasn't necessary (and arguable still isn't, unless you're running java 16). Also, this function was released in java9, whereas this jar skips support for java9 and 10, which might not matter. 🤷

Sineaggi: Finally, there is no _real_ urgency, it's just we already have a solution for java 9+ (using the new `privateLookupIn` function) and going forward it would be nice if a user of this library didn't have to set scary flags (like `--illegal-access=warn`) just to continue running on newer versions of java.

It's just the matter of figuring out how all the pieces are supposed to fit together in this new java world 😄

kdavisk6: As you are finding out, trying to use the multi-release concept, even in theory, is difficult and wrought with drawbacks.  We strive to keep complexity to a minimum.  The simple `if/else` suggestion solves the problem without requiring us to support a multi-release infrastructure.  This same advice is echoed [by the Maven team themselves](https://maven.apache.org/plugins/maven-compiler-plugin/multirelease.html).

> Libraries should make a decision based on: do I need this new Java feature? Can I make this Java version the new requirement? Would it be acceptable to solve this with an else/if-statement as mentioned in the first paragraph?

Given this change is isolated to the `DefaultMethodHandler`, I'd like to keep the changes small, even if we have, what amounts to a temporary `if` statement.

Sineaggi: Are you thinking of maybe there being a class in the java11 package, similar to my branch above, but instead of going around with a full serviceloader and whatnot we instead use a `Class.forName("DefaultMEthodHandlerJava11")` and then fall back to 8 if unavailable? I don't exactly see the 'if/else statement' as being necessarily easier or less complex, it just shifts where the complexity is.
Have you looked at the other prototype branch above? It's not quite as simple as an if statement, but the locator nonesense could be removed in favor of something simpler. Would still require the java11 jar though.

velo: I made an alternative change, with DefaultMethodHandler configurable
https://github.com/OpenFeign/feign/pull/1394

What do you thing of it?!

Sineaggi: My only issue (which is specific to where I work) is for the most part clients ship their own instantiation of the Feign.builder() call, so end-users would have no ability to alter what method handler to use, making this not a viable solution.

Sineaggi: Also thinking about this further, this doesn't seem something pluggable in the same way jackson or the network library is imo, this is just about 'use this for one version of the jdk, this other thing for another version of the jdk'

core/src/main/java9/feign/DefaultMethodHandler.java
```diff
@@ -0,0 +1,73 @@
...
+  // When Feign upgrades to Java 7, remove the @IgnoreJRERequirement annotation.
+  private final MethodHandle unboundHandle;
+
+  // handle is effectively final after bindTo has been called.
+  private MethodHandle handle;
+
+  public DefaultMethodHandler(Method defaultMethod) {
+    try {
+      final Class<?> declaringClass = defaultMethod.getDeclaringClass();
+      final Lookup lookup = MethodHandles.privateLookupIn(declaringClass, MethodHandles.lookup());
```
velo, line 40: I think I have a solution for this...

can we just use reflection on the existing `DefaultMethodHandler`?

We just need try to invoke the new API using reflection, if that fail, fallback to existing code

kdavisk6: That’s what I’ve bee saying

Sineaggi: Will create a pr for this tonight

core/src/main/java9/feign/DefaultMethodHandler.java
velo: I just did, had some time allocated at work to be around feign =)

core/src/main/java9/feign/DefaultMethodHandler.java
velo: I just did, had some time allocated at work to be around feign =)

core/pom.xml
```diff
@@ -79,4 +79,24 @@
...
+      <activation>
+        <jdk>11</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <argLine>--illegal-access=deny</argLine>
```
Sineaggi, line 95: I like this!

core/src/main/java9/feign/DefaultMethodHandler.java
Sineaggi: The changes look good! Really like the use of `deny` to get java 11 to help test this change.

.circleci/config.yml
Sineaggi: I have an uncommitted change for this, moving to `cimg/openjdk:16.0.0` (since that's where circleci appears to be migrating its images to).

velo (approved):
I'm happy with the current version of the change,

This also matches what @kdavisk6 suggested, thumbs up from me

kdavisk6:
💯

PR #1394, opened 2021-05-03
Safe default methodhandler

velo:
@Sineaggi
inspired by your PR, I made default method handler configurable....

Lemme know what you think


velo:
Read some feedback, and agreed that my proposed fix is awful

Issue #1358, opened 2021-01-24
overview.png has black text on (possibly) black background


C-Otto:
With GitHub dark mode the text in https://github.com/OpenFeign/feign/blob/master/src/docs/overview.png cannot be seen, as it is black on black. Please make sure the text is readable with both black and white backgrounds.

kdavisk6:
At the moment, there is nothing we can do about this.  [Github is tracking an issue related to this however.](https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/46)

ashleyfrieze:
@kdavisk6 the mindmap is a PNG file. Can it be generated with a background, rather than transparent?

ashleyfrieze:
Alternatively, if the mindmap were rendered as PlantUML - stored as PlantUML markdown in the repo, and then embedded into the README like this - https://raw.githubusercontent.com/jonashackt/plantuml-markdown/master/README.md - it might be a longer-term option and easier to maintain?

kdavisk6:
Closed with #1396

PR #1396, opened 2021-05-07
Change MindMap to PlantUML

ashleyfrieze:
This has a couple of benefits:

1. Avoid having to regenerate the .PNG every time the mind map is changed
2. Renders automatically with white background, even in dark mode

See #1358


ashleyfrieze:
Note: the URL of the mindmap is presently set to this branch on this fork. This is just for illustration. It would need to be set to the `master` location before merging.

velo (approved):
That is awesome. Thanks for that

ashleyfrieze:
@velo  thanks for taking it on. I’m a big fan of the open feign library. Delighted to contribute even a tiny amount.

PR #1413, opened 2021-05-15
Add @Shared methods to configure shared parameters

hellproxy:
Add the `@Shared` annotation to the default contract. Annotating a method with `@Shared` turns it into a setter/configurator for the built client. 

```java
interface ClientApi {

  @Shared
  ClientApi setEnv(@Param("env") String env);

  @RequestLine("GET /{resource}/{env}")
  void getResource(@Param("resource") String resource);
}
```

This allows users to perform one-time configuration on their client.

```java
ClientApi clientApi = Feign.builder()
  .target(ClientApi.class, "https://example-host.com")
  .setEnv("qa");

// ...

clientApi.getResource("resource-identifier");
```

### DIFF
--- a/core/src/main/java/feign/SharedParameters.java
+++ b/core/src/main/java/feign/SharedParameters.java
@@ -1,7 +1,19 @@
+/**
+ * Copyright 2012-2021 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
 package feign;
 
 import feign.InvocationHandlerFactory.MethodHandler;
-
 import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.IntStream;
--- a/core/src/main/java/feign/SharedParameters.java
+++ b/core/src/main/java/feign/SharedParameters.java
@@ -22,19 +22,19 @@ public class SharedParameters {
 
   private final Map<String, Object> paramNameToValue = new HashMap<>();
 
-  public ConfigurationMethodHandler newHandler(final MethodMetadata methodMetadata) {
-    return new ConfigurationMethodHandler(methodMetadata);
+  public SharedMethodHandler newHandler(final MethodMetadata methodMetadata) {
+    return new SharedMethodHandler(methodMetadata);
   }
 
   public Map<String, Object> asMap() {
     return paramNameToValue;
   }
 
-  public class ConfigurationMethodHandler implements MethodHandler {
+  public class SharedMethodHandler implements MethodHandler {
 
     private final MethodMetadata methodMetadata;
 
-    public ConfigurationMethodHandler(final MethodMetadata methodMetadata) {
+    public SharedMethodHandler(final MethodMetadata methodMetadata) {
       this.methodMetadata = methodMetadata;
     }
 
--- a/core/src/main/java/feign/SharedParameters.java
+++ b/core/src/main/java/feign/SharedParameters.java
@@ -6,16 +6,16 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.IntStream;
 
-public class ConfigurationParameters {
+public class SharedParameters {
 
-    private final Map<String, Object> configuredParams = new HashMap<>();
+    private final Map<String, Object> paramNameToValue = new HashMap<>();
 
     public ConfigurationMethodHandler newHandler(final MethodMetadata methodMetadata) {
         return new ConfigurationMethodHandler(methodMetadata);
     }
 
     public Map<String, Object> getParameterMap() {
-        return configuredParams;
+        return paramNameToValue;
     }
 
     public class ConfigurationMethodHandler implements MethodHandler {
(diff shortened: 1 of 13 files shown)


core/src/main/java/feign/SharedParameters.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private final Map<String, Object> paramNameToValue = new HashMap<>();
+
+  public ConfigurationMethodHandler newHandler(final MethodMetadata methodMetadata) {
+    return new ConfigurationMethodHandler(methodMetadata);
+  }
+
+  public Map<String, Object> asMap() {
+    return paramNameToValue;
+  }
```
kdavisk6, line 31: Consider returning an immutable map instead.  We don't accidental overwrites of the shared parameter value

```suggestion
  public Map<String, Object> asMap() {
    return Collections.unmodifiableMap(paramNameToValue);
  }
```

Or you can return a copy

```suggestion
  public Map<String, Object> asMap() {
    return new LinkedHashMap<>(this.paramNameToValue);
  }
```

core/src/main/java/feign/SharedParameters.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Object invoke(Object[] argv) throws Throwable {
+      IntStream.range(0, argv.length)
+          .forEach(index -> methodMetadata.indexToName()
+              .get(index)
+              .forEach(name -> paramNameToValue.put(name, argv[index])));
+      return null;
+    }
```
kdavisk6, line 48: This could overwrite a value placed on the method itself.  We'll need to consider what to do if a user defined a shared value explicitly in a method.  What should occur here?  Should the shared value always overwrite or would we want to consider the one explicitly declared?

core/src/test/java/feign/FeignTest.java
```diff
@@ -905,6 +905,44 @@ public void beanQueryMapEncoderWithEmptyParams() throws Exception {
         .hasQueryParams("/");
   }
 
+  @Test
+  public void sharedPathParameterIsIncluded() throws Exception {
+    TestInterface api = new TestInterfaceBuilder().queryMapEndcoder(new BeanQueryMapEncoder())
+        .target("http://localhost:" + server.getPort());
+    api.configure("sharedValue");
```
kdavisk6, line 912: Feign `Target` objects are immutable once created.  This breaks that encapsulation and allows for the underlying `MethodMetadata` objects to be modified after the target has been built.  

In addition, there are thread-safety concerns with the mutability.  Feign Target's must be thread safe.  To keep this mechanism, the `ConfigurationMethod` must be synchronized.  

I recommend looking into alternative ways that don't require invoking a method on the Target, but using the `Contract` parsing on the `Builder` directly.  These approaches are already thread-safe, since Target construction must be done all at once.

kdavisk6 (changes requested):
@hellproxy 

Thank you for submitting this Pull Request.  This is a good start, however; we'll need to address the mutability introduced by using a method on the Target interface to supply the configuration.  Feign eagerly configures the Targets and considers them immutable once built.  I'd like you to consider a different method for applying this configuration.

To get you started, you could move the annotation to the Target itself, where users can specify a static value or an `expander` that references an `Expander` implementation for the value.

```java
// maybe consider multiple?
@Shared(parameters = {
   @Param(name="sharedParam", value="static"),
   @Param(name="anotherParam" expander=Expander.class)
})
interface TestInterface {

}
```

Another option is to configure it on the builder directly

```java
Map<String, Object> shared = 
   Map.of("sharedParam", "static", "another", AnotherExpander.class);

Feign.builder()
   .sharedParameters(shared)
   .target();
```

There are just two more options that provide the same flexibility while addressing the thread-safety and immutability concerns.

core/src/main/java/feign/SharedParameters.java
hellproxy: In the example below, I reckon the intent of the user would be to have the `env` parameter passed in `getResourceWithExplicitEnv` take precedence over any value set by `setEnv`. Thoughts?

```java
interface ClientApi {

  @Shared
  void setEnv(@Param("env") String env);

  @RequestLine("GET /{resource}/{env}")
  void getResource(@Param("resource") String resource);

  @RequestLine("GET /{resource}/{env}")
  void getResourceWithExplicitEnv(@Param("resource") String resource, @Param("env") String env);
}
```

This example deserves a unit test.

core/src/test/java/feign/FeignTest.java
hellproxy: Current implementation does indeed sacrifice immutability for the sake of clean, explicit configuration methods. This isn't great.

The best possible solution (imo) I can think of would be for the configuration methods to generate a new target, i.e.

```java
interface ClientApi {
  @Shared
  ClientApi withEnv(@Param("env") String env);
}

ClientApi api = Feign.builder()
  .target(ClientApi.class, "http://host") // creates Target object A
  .withEnv("prod");                       // creates Target object B, object A is unchanged
```

The above is thread-safe and doesn't mutate the initially created target.

This does make the requirements around configuration methods a bit more strict - their return type always needs to be the interface that declares them - but this isn't a particularly hard rule to enforce or follow.

velo (commented):
I see one big limitation to this proposal: it only works for feign-default contract.

I this this is very valuable and we should expose to any contracts.

That is why I proposed passing this to Feign#Builder.

Current usage goes like this:

```
    TestInterface api = new TestInterfaceBuilder()
        .target("http://localhost:" + server.getPort())
        .configure("sharedValue");

    server.enqueue(new MockResponse());
```

If we do something like this:
```
    TestInterface api = new TestInterfaceBuilder()
        .shared("sharedParam", "sharedValue")
        .target("http://localhost:" + server.getPort());

    server.enqueue(new MockResponse());
```

It would be very similar, but now is generic and benefit even people using spring-contract, which is a very large user base

PR #1424, opened 2021-05-21
Bug: ConcurrentModificationException with threads.

JBAZP:


micrometer/pom.xml
```diff
@@ -50,6 +50,17 @@
       <version>2.0.0.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-openfeign-core</artifactId>
+      <version>2.2.6.RELEASE</version>
+      <scope>test</scope>
+    </dependency>
```
kdavisk6, line 58: Please use our internal contract for testing.  We cannot rely on 3rd party contracts.

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
```diff
@@ -97,5 +110,51 @@ protected Meter getMetric(String suffix, String... tags) {
...
+      final ApiRestTest apiRestTest = threadExecution.get();
+      assertNotNull(apiRestTest);
+    }
+
+  }
+
+  private interface ApiRestTest {
+    @GetMapping(path = "/apiRest")
+    public Object find(@RequestParam(value = "clientId") Long clientId);
+  }
```
kdavisk6, line 157: Use our default contract.

kdavisk6 (changes requested):
Please remove the dependency on Spring Cloud and use our Default Contract.

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
```diff
@@ -97,5 +110,51 @@ protected Meter getMetric(String suffix, String... tags) {
...
+
+    // When
+    final List<Future<ApiRestTest>> threadsExecution = tester.run(feignBuilder -> {
+
+
+      final long t1 = System.currentTimeMillis();
+
+      final ApiRestTest apiRestTest = feignBuilder
+          .addCapability(new MicrometerCapability())
+          .target(ApiRestTest.class, "http://localhost:8080");
```
velo, line 136: I don't think this is a valid test.

Imagine instead it was changing the `decoder`... It would be impossible to know which decoder would be used.


Same for capabilities. It's impossible to know which ones are enable for each client.

On a concurrent scenario like this one, reach client should use it's own builder.

velo (changes requested):
I don't think we should do this,  the use case exposed on the tests seems invalid

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
velo: To be a bit more clear, we can change Feign builder to prevent this error. But, we shouldn't cause the builder is not safe to be modified by multiple threads anyway.

Only way for this to be safe, would be changing the builder to Copy on Write... so each method would return a copy of the builder.

Problem on doing that, this is a major changed on feign behaviour

micrometer/pom.xml
JBAZP: modified!

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
JBAZP: modified

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
velo: @kdavisk6 what do you think?

kdavisk6: I agree.  Target construction is not thread safe, and not expected to be.  You cannot share a builder across threads.  That is usage is unsupported.

velo:
Ok, we discussed this, and seems this is not a valid use case.

>  Target construction is not thread safe, and not expected to be. You cannot share a builder across threads. That is usage is unsupported.

Only way for target build to be thread safe, would be changing the builder to Copy on Write... so each method would return a NEW builder.
This would a major changed on feign behavior.

I would love to see such a change for feign 12, but I don't have the availability to do it myself.

Any volunteers?!

rage-shadowman:
A quick-and-dirty solution would be to add a `copy()` method to the builder that would return a new identical instance of the builder that you could then continue to modify without making any changes to the original.

FYI - I agree that builders are not generally intended to be thread safe. They are generally intended to be used within a single thread to build a thread safe object. IMO, making your builder thread-safe is overkill for the vast majority of applications.

PR #1436, opened 2021-06-12
Modify DefaultMethodHandler to work with Android

xrpdevs:
I was getting errors regarding illegal reflective access here, and also another one saying IMPL_LOOKUP was not defined, depending on the Android version I was running on.

These changes should work across the board, Android or actual JRE's.

A very frustrating way for someone inexperienced with reflection to spend a day :/

HTH someone!


core/src/main/java/feign/DefaultMethodHandler.java
```diff
@@ -1,114 +1,164 @@
...
-    try {
-      Lookup lookup = readLookup(declaringClass);
-      this.unboundHandle = lookup.unreflectSpecial(defaultMethod, declaringClass);
-    } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException
-        | InvocationTargetException ex) {
-      throw new IllegalStateException(ex);
+    // Uses Java 7 MethodHandle based reflection. As default methods will only exist when
+    // run on a Java 8 JVM this will not affect use on legacy JVMs.
+    // When Feign upgrades to Java 7, remove the @IgnoreJRERequirement annotation.
+    public MethodHandle unboundHandle;
```
velo, line 36: please keep `final` modifier on this field

core/src/main/java/feign/DefaultMethodHandler.java
```diff
@@ -1,114 +1,164 @@
...
+                Lookup lookup = readLookup(declaringClass);
+                this.unboundHandle = lookup.unreflectSpecial(defaultMethod, declaringClass);
+            }
+
+        } catch (IllegalAccessException ex) {
+            throw new IllegalStateException(ex);
+        } catch (InstantiationException | InvocationTargetException | NoSuchMethodException e) {
+            e.printStackTrace();
+        } catch (NoSuchFieldException e) {
+            e.printStackTrace();
```
velo, line 70: this is not merge-able, as this is eating the exceptions.  If we are unable to get the `unboundHandle` instance, throw the exception and let it blow.

Since you are going to touch this, please also include an message to this exception

velo (changes requested):
Build is not passing, so, I can't review it in details unless this is building successfully

core/src/main/java/feign/DefaultMethodHandler.java
xrpdevs: Currently I'm building this with Gradle in Android Studio.

Will refactor tomorrow and try to build in Maven against different JDKs.

xrpdevs: See latest commit. Hopefully this fixes problems you mentioned

velo: I will wait until build passes to give this a second look

core/src/main/java/feign/DefaultMethodHandler.java
```diff
@@ -27,125 +27,126 @@
...
-        }
+  }
+
+  private Lookup readLookup(Class<?> declaringClass)
+      throws IllegalAccessException, InvocationTargetException, NoSuchFieldException {
+    try {
+      if (System.getProperty("java.vm.name").equalsIgnoreCase("Dalvik")) {
+        return androidLookup(declaringClass);
+      } else {
+        return safeReadLookup(declaringClass);
```
xrpdevs, line 57: I had "readLookup" instead of "safeReadLookup" here. Building now...

xrpdevs (commented):
This is building OK now?

velo:
For some unknow reason, this is not building at all:
Screen Shot 2021-06-16 at 15 22 56


@kdavisk6 any clue?

xrpdevs:
Buiding locally for me without issue (IDEA/Maven) and passing CircleCi tests.

I did read this regarding merges being blocked due to how CircleCI is configured:

https://circleci.com/docs/2.0/workflows/#workflows-waiting-for-status-in-github 

Marvin Froeder ***@***.***> wrote ..

PR #1441, opened 2021-06-16
Android compat

xrpdevs:
I read somewhere that if I created a new branch this might solve the merge issue. Closing previous PR.


xrpdevs:
Nope, that didn't work either.

Looks like it has something to do with 
https://github.com/OpenFeign/feign/pull/1436#issuecomment-862009191

xrpdevs:
circleci-feign

velo:
Somehow, when I open a new PR it built fine
https://github.com/OpenFeign/feign/pull/1442

velo:
https://app.circleci.com/pipelines/github/OpenFeign/feign/322/workflows/190526ae-13b1-458b-9707-ac46f3df0279/jobs/1039

xrpdevs:
I can't find a sane way to add a test in an android environment with maven in CircleCI. I've reverted the config.yml to it's initial state.

The only realistic way to get tests to run seems to be to reconfigure the entire project as a gradle project.

In any case, all of the code I have comitted is pure java, we're not doing anything with any android-specific dependencies here, it's just a workaround for how Android's restrictions on reflection stopped feign from working on that platform.

Feign is a Java project, not an Android project in any way shape or form - therefore I don't really think testing against an android virtual machine is going to be helpful here, besides, doing this in CircleCI is fairly compute intensive and requires the spinning up of an entire KVM instance rather than just running in Docker.

Given that this is passing, would it not be better to just merge into master and if anybody using Android has issues with it they can open a ticket? I don't mind revisiting this as and when necessary, for example Google might make changes in a future version of Android..

velo:
> I can't find a sane way to add a test in an android environment with maven in CircleCI. I've reverted the config.yml to it's initial state.
> 
> The only realistic way to get tests to run seems to be to reconfigure the entire project as a gradle project.
> 
> In any case, all of the code I have comitted is pure java, we're not doing anything with any android-specific dependencies here, it's just a workaround for how Android's restrictions on reflection stopped feign from working on that platform.
> 
> Feign is a Java project, not an Android project in any way shape or form - therefore I don't really think testing against an android virtual machine is going to be helpful here, besides, doing this in CircleCI is fairly compute intensive and requires the spinning up of an entire KVM instance rather than just running in Docker.
> 
> Given that this is passing, would it not be better to just merge into master and if anybody using Android has issues with it they can open a ticket? I don't mind revisiting this as and when necessary, for example Google might make changes in a future version of Android..

Ok, merged based on my branch, just cause circle CI ran on that one with positive feedback

kdavisk6:
I can handle the gradle/circle issues. We can make them optional

xrpdevs:
Looking at https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/android/images/api-26/Dockerfile
it seems that any testing that the android app test / build / deploy flow the tests (if we were testing an android app) are being run with OpenJDK 11

PR #1442, opened 2021-06-16
Android support

velo:


velo:
Same as https://github.com/OpenFeign/feign/pull/1441, just a alternative way to get circle CI to build

core/src/main/java/feign/DefaultMethodHandler.java
```diff
@@ -50,12 +51,49 @@ public DefaultMethodHandler(Method defaultMethod) {
   private Lookup readLookup(Class<?> declaringClass)
       throws IllegalAccessException, InvocationTargetException, NoSuchFieldException {
     try {
-      return safeReadLookup(declaringClass);
-    } catch (NoSuchMethodException e) {
+      if (System.getProperty("java.vm.name").equalsIgnoreCase("Dalvik")) {
+        return androidLookup(declaringClass);
```
velo, line 55: this entire flow is not been tested at all, as we only build feign with openjdk.

Is that something we could include on circleCI?

velo (commented):
LGTM

core/src/main/java/feign/DefaultMethodHandler.java
xrpdevs: Yes CircleCI allows to run under Dalvik/ART.

Build of my project was OK in CCI with the changes to feign-core

I'll look into if/how the flow can tested against Android's JVM's, for this module or perhaps for the whole project if it works out more easy to just add a line to the .yml specifying Dalvik/ART.

xrpdevs: Hmm my most recent changes seem to have gone missing in the android-compat branch. Will update it later.

xrpdevs:
Please review changes in the other branch - my new commits are not showing up here.

The entire flow can now be tested against openjdk as we're no longer checking for Dalvik, just running the methods that work on Android when the safeReadLookup method fails, falling back to legacyReadLookup which was originally the default behaviour.

velo:
Ok, I merged your latest changes here.

Looks nice I would really like to see android VMs been tested on circle CI, IMHO, that is all that is missing

PR #1444, opened 2021-06-16
Android support

xrpdevs:
This should sort out any problems related to flow not being tested.

Tested and working in Java 8, 11, 14 and Android. 

Whilst we could use CircleCI to build/test specifically against Dalvik, the amount of incompatibility between Androids JRE and a normal Java one is fairly minimal, generally speaking anything that builds agains java 8 will run just fine on Android, with the exception of certain types of reflective access - this isn't a JRE issue as much as it is a "security feature".


xrpdevs:
Could this be merged into master now so I can sort out the mess that my project's dependencies have become in the course of debugging this?

PR #1452, opened 2021-07-06
Add response headers to FeignException

radio-rogal:
Implementation of #1268

This PR adds the ability to retrieve the response headers from a `FeignException`.

This is replacement for #1277 as the source branch is gone.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -29,6 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
+    <hamcrest.version>1.3</hamcrest.version>
   </properties>
 
   <dependencies>
@@ -61,6 +62,13 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-core</artifactId>
+      <version>${hamcrest.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -22,12 +22,12 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import static feign.Util.UTF_8;
-import static feign.Util.checkNotNull;
+import static feign.Util.*;
 import static java.lang.String.format;
 
 /**
@@ -40,6 +40,7 @@ public class FeignException extends RuntimeException {
   private static final long serialVersionUID = 0;
   private int status;
   private byte[] responseBody;
+  private Map<String, Collection<String>> responseHeaders;
   private Request request;
 
   protected FeignException(int status, String message, Throwable cause) {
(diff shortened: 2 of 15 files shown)


core/src/main/java/feign/Util.java
```diff
@@ -360,4 +352,33 @@ public static boolean isNotBlank(String value) {
...
+
+  /**
+   * Copy entire map of string collection.
+   *
+   * The copy is unmodified map of unmodified collections.
+   *
+   * @param map string collection map
+   * @return copy of the map or an empty map if the map is null.
+   */
+  public static Map<String, Collection<String>> caseInsensitiveCopyOf(Map<String, Collection<String>> map) {
```
radio-rogal, line 364: Unit-tests will be provided later.

radio-rogal (commented):
I think it would be great if @pzapasni and @virtual-machinist left some comment.

core/src/main/java/feign/Util.java
```diff
@@ -360,4 +352,33 @@ public static boolean isNotBlank(String value) {
   public static boolean isBlank(String value) {
     return value == null || value.isEmpty();
   }
+
+  /**
+   * Copy entire map of string collection.
+   *
+   * The copy is unmodified map of unmodified collections.
```
virtual-machinist, line 359: Typo: should be _unmodifiable_ (those that you cannot modify), not _unmodified_ (those that you didn't modify)?

virtual-machinist:
Can you please also add headers to `feign.codec.DecodeException`? Doing this would be trivial with your existing changes - you just have to add a couple of constructors that accept `feign.Response` instead of `feign.Request` and then calling another `super()` that accepts both `feign.Request` and headers.

radio-rogal:
> Can you please also add headers to `feign.codec.DecodeException`? 

I am OK with this but looks like it takes more time: there are a lot of calls of the method `Util.toByteArray(response.body().asInputStream())` that could throw `IOException`.
In my opinion it should be refactored: at least to make `responseBodyToByteArray` that hides work with `InputStream` or pass to constructors whole response object instead of pair body as array and headers.

virtual-machinist:
Ah, I see what you mean. However no `feign.codec.DecodeException` constructor currently even passes the body object to `super()`, so I'd argue we can safely pass a `null` instead of the body array.

But Perhaps you're right and it's better to do this addition as a follow-up in a separate issue.

radio-rogal:
@velo && @kdavisk6 how about remove some constructors and do not use the parameter response body as prepared byte array but use the response?

**Proposal constructors of FeignException:**

- status, message // EncodeException
- status, message, cause // EncodeException
- status, message, request // RetryableException but we do not use it - so it could be removed too
- status, message, request, cause // RetryableException when a client could not read response at all.
- status, message, response // DecodeException, errorStatus() -> ErrorDecoder, FeignClientException, FeignServerException
- status, message, response, cause // AsyncJoinException, DecodeException, errorReading() -> AsyncResponseHandler

The reading of response body move to an utility method that hides possible `IOException` and returns `Nullable` byte array.

core/pom.xml
```diff
@@ -29,6 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
+    <hamcrest.version>1.3</hamcrest.version>
```
velo, line 32: can we move this to the parent pom?!

velo: also, there is a hamcrest 2.0 somewhere, would prefer to go with that one

core/src/main/java/feign/FeignException.java
```diff
@@ -61,10 +64,12 @@ protected FeignException(int status, String message) {
     this.request = null;
   }
 
-  protected FeignException(int status, String message, byte[] responseBody) {
+  protected FeignException(int status, String message, byte[] responseBody,
+      Map<String, Collection<String>> responseHeaders) {
     super(message);
     this.status = status;
     this.responseBody = responseBody;
+    this.responseHeaders = caseInsensitiveCopyOf(responseHeaders);
```
velo, line 72: I don't think `FeignException` should re-order response headers.  It may make sense to sort it, sure, but not here.

core/src/main/java/feign/FeignException.java
```diff
@@ -40,6 +40,7 @@
   private static final long serialVersionUID = 0;
   private int status;
   private byte[] responseBody;
+  private Map<String, Collection<String>> responseHeaders;
   private Request request;
 
   protected FeignException(int status, String message, Throwable cause) {
```
velo, line 46: this exception is getting more and more complicated.... wondering if we should have one constructor that takes `Request` as argument and the other one that takes `Response` 

Anyway, just concerned that this class is getting more and more complicated.

velo (approved):
I think is a good improvement overall.

Could you please just file a ticket to simplify `FeignException`?

Ideally it should have one constructor for `Request` and one for `Response`. I'm sure it's more complicated than that.

Issue #1448, opened 2021-06-30
Add support for encoders even when no or multiple body parameters are presented


fabiocarvalho777:
Currently Feign will not resolve methods with `BuildEncodedTemplateFromArgs` if they have 0 or more than 1 body parameters, which prevents the usage of custom Feign encoders. There use cases though when an application (or library) wants to a client to provide a custom encoder and wants that encoder triggered even if the client method has 0 or more than 1 "body parameters" (parameters not annotated with Feign annotations or custom annotations registered in Feign using `DeclarativeContract.registerParameterAnnotation`).

A few examples of this use case:

1. An application needs a Feign client whose request payload is defined entirely by a custom Feign encoder even if no parameter is declared at the client method. In this case, even with no presence of body parameter the encoder will know how to define the request payload (based on the method name, method return type, and other optional metadata provided by annotations).
2. An application needs a Feign client whose request payload is defined by a custom Feign encoder plus one or more parameters, all of them annotated with custom annotation(s) registered in Feign using `DeclarativeContract.registerParameterAnnotation`. In this case the encoder will use all the parameters, plus metadata in their respective annotations, to properly define the request message payload.

To be more specific, this issue is required to support certain use cases in [Mocca](https://github.com/paypal/mocca). This improvement would help to simplify Mocca's API a lot.

We are also open to contributing a PR to implement this feature, if Feign team first accesses this issue and let us know we should do so.

By the, we tried first to address this use case by creating a custom Feign Contract implementation, but that was not possible because overwriting a `parseAndValidateMetadata` requires instantiating `MethodMetadata` and that class has a package-private constructor (besides being final too).

Please let us know your thoughts on this and, if you believe this is an useful enhancement to have, if we can go ahead and provide a PR addressing this (or if you prefer to do so yourselves, please let us know the timeline to provide it).

Thanks.

fabiocarvalho777:
By the way, regarding how to implement this, I was thinking about providing a subclass of `DeclarativeContract`. That would be also nice in terms of backwards compatibility. Let me know if that makes sense or if you have another idea to recommend. Thanks.

kdavisk6:
@fabiocarvalho777 Looks like we work together. Look me up at work and we can chat.

velo:
HI Fabio,

0 bodies should be fine. Let's see if I got it correctly. By ZERO body you mean a client like this:

```
interface Bank {
  @RequestLine("POST /account/{id}")
[...]

PR #1459, opened 2021-07-13
Modifying Contract to support passing all parameters to encoders #1448

fabiocarvalho777:
This PR addresses issue #1448 

@velo @kdavisk6 please just let me know if you have any requests for changes, or if this PR should be opened against a different branch and I will do it. Thank you very much.


fabiocarvalho777:
I noticed CircleCI build failed. However, the error is not related to my changes. Here is the error:

```
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project feign-core: Some files do not have the expected license header -> [Help 1]
```

When building from my laptop I skip that license header check by adding `-Dlicense.skip=true` to my Maven command. I am surprised though the build in CircleCI is failing because of that.

What should be done to avoid this build error in CircleCI?

velo:
Run `license:format` locally and you should manage to fix whatever needs fixing.

fabiocarvalho777:
@velo build passed.

velo:
Sweet, I gave it a very initial pass. I see no tests for it, could you please include some?!

fabiocarvalho777:
@velo added tests. Builds passed. Thanks.

fabiocarvalho777:
@velo hello. Checking if you had a chance to review the PR. Thanks.

velo:
> @velo hello. Checking if you had a chance to review the PR. Thanks.

thanks Fabio, I downloaded a copy of your changes here, and I'm tinkering around, trying to avoid this boolean, in an attempt to create a more generalized solution. Hopefully I will have an answer over the weekend

kdavisk6 (approved):
Meets the need and fits as a minimum viable release.  I'm going to approve it, pending additional comments from @velo

fabiocarvalho777:
> > @velo hello. Checking if you had a chance to review the PR. Thanks.
> 
> thanks Fabio, I downloaded a copy of your changes here, and I'm tinkering around, trying to avoid this boolean, in an attempt to create a more generalized solution. Hopefully I will have an answer over the weekend

Thanks. Let me know if you have any question, or need any change.

velo:
I still feel this should be a contract-only change.


Contract holds the magic on how to prepare payloads for encoders, so, it alone should be able to "translate" method arguments into an Object[] or Map<> or even a different object assembled method arguments.  Just haven't figure HOW yet

fabiocarvalho777:
> I still feel this should be a contract-only change.

Sorry, I thought this is the Contact change we have been discussing. Isn't it? As opposed to changing the Encoder API.

> it alone should be able to "translate" method arguments into an Object[] or Map<> or even a different object assembled method arguments

But that is exactly what this solution does, it delivers all method arguments into a `Object[]`. And all of that preserving the current APIs (except for the addition of `alwaysEncodeBody`).

> I just don't wanna introduce a "fabio" specific change now, to only figure that we should have done something else later.

I hear you, no one wants specific solutions, including myself, trust me, otherwise the project becomes unmaintainable and nobody wants their projects to rely on unmainatable libraries. I have maintained 3 different open source projects, so I know what you are talking about. I am just not sure that is the case here, as (please correct me if I am wrong) the PR meets what has been asked in the discussion in this thread, including the points in your last comment.

Anyways, please let me know if you have any other idea, or if there is anything else I can do to help, I would be glad to change anything and retest.

Thanks.

velo:
> Sorry, I thought this is the Contact change we have been discussing. Isn't it? As opposed to changing the Encoder API

My concern, is this boolean is behaving more like an "body no op contract", and then we need to check the boolean everywhere to ensure correct behavior.

Let's say we have a client like this:

```
@RequestLine("GET /{id}/something")
@Headers("Authorization: {login}")
public String readSomething(@Param("id") String id, @Param("login") String login, Object body);
```

Feign behave in very distinct ways with `alwaysEncodeBody`, and I'm sure this will be generating plenty of support, specially for those using third party encoders that will inherit this "for free"

I was not able to find a different solution for your problem over the weekend, and I won't have the time to get a different approach.

Let's compromise. Instead of having `alwaysEncodeBody` on BaseContract, that will spread around like wildfire, let's have a new Contract base class with this functionality and let's leave existing feign and third party contracts intact.

I don't exactly have a name suggestion for that class, maybe something like `BodyNoOpContract`, `Body?Contract`, `AlwaysEncodeBodyBaseContract`

If you happy to proceed with that, I would merge and release feign 11.5 today.

core/src/main/java/feign/MethodMetadata.java
```diff
@@ -118,6 +119,15 @@ public MethodMetadata queryMapEncoded(boolean queryMapEncoded) {
     return this;
   }
 
+  public boolean alwaysEncodeBody() {
+    return alwaysEncodeBody;
+  }
+
+  MethodMetadata alwaysEncodeBody(boolean alwaysEncodeBody) {
```
velo, line 126: Please annotate both methods with `@Experimental`, just in case we change our minds later on.

fabiocarvalho777:
> > Sorry, I thought this is the Contact change we have been discussing. Isn't it? As opposed to changing the Encoder API
> 
> My concern, is this boolean is behaving more like an "body no op contract", and then we need to check the boolean everywhere to ensure correct behavior.
> 
> Let's say we have a client like this:
> 
> ```
> @RequestLine("GET /{id}/something")
> @Headers("Authorization: {login}")
> public String readSomething(@Param("id") String id, @Param("login") String login, Object body);
> ```
> 
> Feign behave in very distinct ways with `alwaysEncodeBody`, and I'm sure this will be generating plenty of support, specially for those using third party encoders that will inherit this "for free"
> 
> I was not able to find a different solution for your problem over the weekend, and I won't have the time to get a different approach.
> 
> Let's compromise. Instead of having `alwaysEncodeBody` on BaseContract, that will spread around like wildfire, let's have a new Contract base class with this functionality and let's leave existing feign and third party contracts intact.
> 
> I don't exactly have a name suggestion for that class, maybe something like `BodyNoOpContract`, `Body?Contract`, `AlwaysEncodeBodyBaseContract`
> 
> If you happy to proceed with that, I would merge and release feign 11.5 today.

Thank you very much. I will implement these changes now.

core/src/main/java/feign/MethodMetadata.java
fabiocarvalho777: Ok, I will do that. Thanks.

core/src/main/java/feign/AlwaysEncodeBodyContract.java
```diff
@@ -0,0 +1,31 @@
...
+ * This type of contract is useful when an application needs a Feign client whose request payload is
+ * defined entirely by a custom Feign encoder regardless of how many parameters are declared at the
+ * client method. In this case, even with no presence of body parameter the provided encoder will
+ * have to know how to define the request payload (for example, based on the method name, method
+ * return type, and other metadata provided by custom annotations, all available via the provided
+ * {@link RequestTemplate} object).
+ *
+ * @author fabiocarvalho777@gmail.com
+ */
+public abstract class AlwaysEncodeBodyContract extends DeclarativeContract {
```
fabiocarvalho777, line 30: Since the purpose of this class is to allow the body always to be defined by the encoder, there was no need to add `alwaysEncodeBody` field here.

core/src/main/java/feign/Contract.java
```diff
@@ -111,7 +84,9 @@ protected MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method me
       data.method(method);
       data.returnType(Types.resolve(targetType, targetType, method.getGenericReturnType()));
       data.configKey(Feign.configKey(targetType, method));
-      data.alwaysEncodeBody(this.alwaysEncodeBody);
+      if (AlwaysEncodeBodyContract.class.isAssignableFrom(this.getClass())) {
```
velo, line 87: can't you override this method on the subclass?

core/src/main/java/feign/AlwaysEncodeBodyContract.java
fabiocarvalho777: Notice this class extends `DeclarativeContract`, instead of `BaseContract`. That is necessary to take advantage of the annotations processors, which is required for extensions of `AlwaysEncodeBodyContract`.

velo (approved):
Looking good

core/src/main/java/feign/MethodMetadata.java
fabiocarvalho777: Done

core/src/main/java/feign/AlwaysEncodeBodyContract.java
velo: makes sense.

core/src/main/java/feign/Contract.java
velo: na, I don't think it's possible.
ok, nvm

core/src/main/java/feign/Contract.java
```diff
@@ -84,6 +84,9 @@ protected MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method me
       data.method(method);
       data.returnType(Types.resolve(targetType, targetType, method.getGenericReturnType()));
       data.configKey(Feign.configKey(targetType, method));
+      if (AlwaysEncodeBodyContract.class.isAssignableFrom(this.getClass())) {
```
fabiocarvalho777, line 87: This setting is necessary here to avoid having to override this whole method (a copy plus this setting) under `AlwaysEncodeBodyContract`.

velo:
Here you go

https://repo1.maven.org/maven2/io/github/openfeign/feign-core/11.5/

fabiocarvalho777:
@velo @kdavisk6 thank you very much, we really appreciate your time and work!!

velo:
All good, sorry for holding this for so long.

FWIW, the way I would love to see this done (as in the future) is something like this:

```
@RequestLine("GET /{id}/something")
@Headers("Authorization: {login}")
@Body(expression = "java( new Object[] {id, login, name, age} )")
public String readSomething(@Param("id") String id, @Param("login") String login, String name, int age);
```

The main advantage of this `Body#expression` is the flexibility.

Prefer a Map instead, sure:
```
@Body("java( Map.of("id", id, "name": name, ... ) )")
```

Do you have your own object, not a problem
```
@Body("java( new InternalRecord( id, login, new UserIdentifiableInformation(name, age) ) )")
```

HOW you may ask?! No idea, I just have a dream!

PR #1467, opened 2021-07-23
FeignException uses request and response but doesn't use a response body

radio-rogal:
Simplifying of FeignException: it uses request and response objects but does not use a response body as byte array.

It is breaking changes if any extension use the methods _content_. _contentUTF8_, _responseBody_ or _responseHeaders_.
FeignException and DecodeException have new constructors.

These changers start from the discussion #1452

P.S.
I like [the idea to make FeignException more simplier](https://github.com/OpenFeign/feign/pull/1452#issuecomment-875538393). But I have not found how to do it without breaking changes.

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -37,108 +37,71 @@ public class FeignException extends RuntimeException {
 
   private static final String EXCEPTION_MESSAGE_TEMPLATE_NULL_REQUEST =
       "request should not be null";
+  private static final String EXCEPTION_MESSAGE_TEMPLATE_NULL_RESPONSE =
+      "response should not be null";
   private static final long serialVersionUID = 0;
-  private int status;
-  private byte[] responseBody;
-  private Map<String, Collection<String>> responseHeaders;
+  private final int status;
   private Request request;
+  private Response response;
 
-  protected FeignException(int status, String message, Throwable cause) {
-    super(message, cause);
-    this.status = status;
-    this.request = null;
-  }
-
-  protected FeignException(int status, String message, Throwable cause, byte[] responseBody,
-      Map<String, Collection<String>> responseHeaders) {
-    super(message, cause);
-    this.status = status;
-    this.responseBody = responseBody;
-    this.responseHeaders = caseInsensitiveCopyOf(responseHeaders);
-    this.request = null;
-  }
-
+  // EncodeException
   protected FeignException(int status, String message) {
     super(message);
     this.status = status;
     this.request = null;
   }
 
-  protected FeignException(int status, String message, byte[] responseBody,
-      Map<String, Collection<String>> responseHeaders) {
-    super(message);
+  // EncodeException
+  protected FeignException(int status, String message, Throwable cause) {
+    super(message, cause);
     this.status = status;
-    this.responseBody = responseBody;
-    this.responseHeaders = caseInsensitiveCopyOf(responseHeaders);
     this.request = null;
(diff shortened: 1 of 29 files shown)


core/src/main/java/feign/FeignException.java
```diff
@@ -37,108 +37,61 @@
...
+  private static final String EXCEPTION_MESSAGE_TEMPLATE_NULL_RESPONSE =
+      "response should not be null";
   private static final long serialVersionUID = 0;
-  private int status;
-  private byte[] responseBody;
-  private Map<String, Collection<String>> responseHeaders;
+  private final int status;
   private Request request;
-
-  protected FeignException(int status, String message, Throwable cause) {
```
velo, line 46: I HATE this class!

That said, it's not safe to remove the extra constructors, as people may be using it.

Bring them back, and feel free to deprecate them all

core/src/main/java/feign/FeignException.java
```diff
@@ -37,108 +37,61 @@
...
-  public byte[] content() {
-    return this.responseBody;
-  }
-
-  /**
-   * The Response body.
-   *
-   * @return an Optional wrapping the response body.
-   */
-  public Optional<ByteBuffer> responseBody() {
```
velo, line 130: as with the original constructors... we need to keep the old methods for compatibility.

Please bring them all back, not sure if they got deleted or just sorted, keeping them on their original position makes this more reviewable.

core/src/main/java/feign/FeignException.java
```diff
@@ -271,149 +203,130 @@ static FeignException errorExecuting(Request request, IOException cause) {
   }
 
   public static class FeignClientException extends FeignException {
-    public FeignClientException(int status, String message, Request request, byte[] body,
```
velo, line 274: anything public, needs to be kept, add `@deprected` to them, but keep them in place

velo (changes requested):
Overall I liked the change, but, due to feign nature, we CAN'T just remove public/protected methods/constructors.  At least not in minor releases.

Please bring them back, mark as deprecated and we may nuke on feign 12...

velo:
> I like [the idea to make FeignException more simplier](https://github.com/OpenFeign/feign/pull/1452#issuecomment-875538393). But I have not found how to do it without breaking changes.

I feel ya

core/src/main/java/feign/FeignException.java
radio-rogal: done in the commit 2523ef1

core/src/main/java/feign/FeignException.java
radio-rogal: done in the commit 2523ef1

core/src/main/java/feign/FeignException.java
radio-rogal: done in the commit 2523ef1

radio-rogal:
The Snyk upgraded jackson's version in c88b65a and the build was successful. But then I update my branch and it fell for jdk14 only.
Unfortunatelly I can see what is wrong.

radio-rogal:
Today I tried to return unit-tests for response body methods. I thought that all these methods should be deprecated. But I found that they are required as a response could have a non-repeatable body so:

-  we could not made an exception message without touching a response body
-  we could not get a response body from exception if a response has non-repeatable body as we just read it when created new exception.

radio-rogal:
It requires a different approach, perhaps you need to transfer everything to factory methods.

PR #1476, opened 2021-08-03
Move hamcrest dependency to the parent, upgrade it to 2.2

radio-rogal:
These changes was discussed in #1452
https://github.com/OpenFeign/feign/pull/1452#discussion_r665717776
https://github.com/OpenFeign/feign/pull/1452#discussion_r665718036

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -29,7 +29,6 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <hamcrest.version>1.3</hamcrest.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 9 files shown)


velo (approved):
Outstanding cleanup, having 2 versions was so amateur 🤦‍♂️ 

Is it possible to use a single dependency? Or need both `hamcrest` and `hamcrest-library`? Any chance that one pulls the other automatically?

radio-rogal:
> Is it possible to use a single dependency? Or need both `hamcrest` and `hamcrest-library`? Any chance that one pulls the other automatically?

It is a good question: `hamcrest-library`-2.2 contains only `HamcrestLibraryIsDeprecated` class :)

velo (approved):

Issue #873, opened 2019-01-10
Is there a way to retrieve @RequestLine information in an interceptor ?


pilak:
Hi there,

I'm looking for a way to access the request pattern url called in a logging interceptor.

Precisely, I need to get the raw url filled in @RequestLine annotation on the client class...

Is that possible ?
Best regards

kdavisk6:
Are you looking for the uri template used or the resolved value?

starlight36:
I have a question like this. Is there some way to access uri template from RequestTemplate? Do you have any suggestions for me?

flisky:
@kdavisk6 The raw template uri (before resolved).

We also want to record `requestTemplate.url()` as a tag in a micrometer's Timer.

Currently, we have to do in the following way:

1. wireup a interceptor to put the url into headers:
`template -> template.header(ORIGINAL_URI_HEADER, template.url())`

2. wireup a client to access the value and then remove it from headers:
```java
String originalUrl = null;
Map<String, Collection<String>> headers = request.headers();
if (headers.containsKey(ORIGINAL_URI_HEADER)) {
    Collection<String> originalUrlHeaders = headers.get(ORIGINAL_URI_HEADER);
    if (!originalUrlHeaders.isEmpty()) {
        originalUrl = originalUrlHeaders.iterator().next();
        // recreate request because of immutable headers
        val newHeaders = new HashMap<String, Collection<String>>(headers);
        newHeaders.remove(ORIGINAL_URI_HEADER);
        request = Request.create(request.method(), request.url(), newHeaders, request.body(), request.charset());
    }
}
```

Any suggestions? Thanks!

kdavisk6:
@flisky 

By the time the `RequestTemplate` makes it to the interceptor, it resolved.  I'm unclear what value the unresolved uri template brings.  Can you give me context?

flisky:
Quoted from @whiskeysierra

> Feign doesn't expose the URI template which significantly reduces the usefulness of the Micrometer and OpenTracing integration since both use the URI template as tags

The request would be aggregated by the unresolved uri template. And the time-series database could be exploded due to a high-cardinality tag.

kdavisk6:
@flisky 

Thank you for that clarification.  Is there another issue where that quote is from?  I'd like to review it further to get a better understanding of what the desired outcome is.

flisky:
It's from zalando/riptide#618.

[...]

PR #1493, opened 2021-08-26
Aggregate Micrometer "uri" label values under uri template

### LINKED ISSUE #1481: Feign Micrometer processes slugs in URIs
When I'm looking at requests sent to prometheus, this is what I expect to see:
`http_server_requests_seconds_count{exception="ExternalException",method="GET",outcome="SERVER_ERROR",status="500",uri="/api/v1/userdetails/{userId}",} 2.0`
Notice the unexpended slug.

However, with feign client, this is what I receive:
`feign_Client_seconds_count{client="net.corp.integration.MyClient",host="anotherServer.net",method="findCustomerById",uri="https://anotherServer.net/api/v2/customers/12347",} 4.0`

Notice the expanded slug. With every different customerId I'll get a different log entry rather than a grouping.

I've done some digging and it seems that the MeteredClient works with an expanded requestTemplate rather than the original. I've also noticed that at some point in the execution the original requet template is replaced by the resolved template. I'm not sure where that's happpening and I'd gladly submit a MR if somebody would present a solution to this.

Feign micrometer version: 11.5

martinacat:
# Overview
Use path expression rather than explicit value for Micrometer `uri` labels

This new default means that the value of the `uri` metric label becomes `/get/{id}` rather than `/get/123`

* Changes the default value of the `uri` label
* Adds a bunch of new unit tests around that feature

# Motivation
This change provides better aggregation, protecting time series databases such as Prometheus from a large influx of time series (high-cardinality metrics).

Currently - as noted in #1481 - enabling feign client metrics from config on clients with a parametrized urls creates a scenario where a new time series is produced for each value of the provided parameter. This provides extremely high-cardinality metrics at a fast rate endangering the underlying time series database. (Resolves #1481, related to #873)

When traffic is significant, collecting client metrics with an uri expression such as `reservation/{reservationId}` can generate tens of thousands of time series (one for each value of the label, a potentially infinite set), putting Prometheus at risk. Enabling OOTB metrics without being aware of the high cardinality of some metrics is quite easy to do, so we thought we could protect some users by changing the default behavior 

# Testing
* Unit tests written
* Approach was also tested on one of my existing applications as described in this [comment](https://github.com/OpenFeign/feign/issues/1481#issuecomment-906349520)

Screen Shot 2021-08-26 at 12 38 08 PM

# Questions
* Is there any integration or regression tests that are worth expanding to cover this feature?
* Which CHANGELOG can I add this feature to? 

Thank you for your time! 😊

### DIFF
--- a/micrometer/src/main/java/feign/micrometer/MeteredClient.java
+++ b/micrometer/src/main/java/feign/micrometer/MeteredClient.java
@@ -83,9 +83,9 @@ public class MeteredClient implements Client {
     final RequestTemplate template = request.requestTemplate();
     final Tags allTags = metricTagResolver
         .tag(template.methodMetadata(), template.feignTarget(), e,
-            new Tag[] {Tag.of("http_status", String.valueOf(responseStatus)),
-                Tag.of("status_group", responseStatus / 100 + "xx"),
-                Tag.of("uri", template.path())})
+            Tag.of("http_status", String.valueOf(responseStatus)),
+            Tag.of("status_group", responseStatus / 100 + "xx"),
+            Tag.of("uri", template.methodMetadata().template().path()))
         .and(extraTags);
     meterRegistry.counter(
         metricName.name("http_response_code"),
@@ -99,7 +99,8 @@ public class MeteredClient implements Client {
                               Exception e) {
     final RequestTemplate template = request.requestTemplate();
     final Tags allTags = metricTagResolver
-        .tag(template.methodMetadata(), template.feignTarget(), e, Tag.of("uri", template.path()))
+        .tag(template.methodMetadata(), template.feignTarget(), e,
+            Tag.of("uri", template.methodMetadata().template().path()))
         .and(extraTags(request, response, options, e));
     return meterRegistry.timer(metricName.name(e), allTags);
   }
--- a/micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
+++ b/micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
@@ -13,100 +13,24 @@
  */
 package feign.micrometer;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import feign.*;
-import feign.mock.HttpMethod;
-import feign.mock.MockClient;
-import feign.mock.MockTarget;
+import feign.Capability;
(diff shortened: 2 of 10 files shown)


micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -83,9 +83,9 @@ protected void countResponseCode(Request request,
     final RequestTemplate template = request.requestTemplate();
     final Tags allTags = metricTagResolver
         .tag(template.methodMetadata(), template.feignTarget(), e,
-            new Tag[] {Tag.of("http_status", String.valueOf(responseStatus)),
-                Tag.of("status_group", responseStatus / 100 + "xx"),
-                Tag.of("uri", template.path())})
+            Tag.of("http_status", String.valueOf(responseStatus)),
+            Tag.of("status_group", responseStatus / 100 + "xx"),
+            Tag.of("uri", template.methodMetadata().template().path()))
```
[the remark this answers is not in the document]
velo: I think NPE is only likely for a manually created RequestTemplate.

velo (approved):

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
```diff
@@ -18,19 +18,96 @@
...
+
+  public interface SourceWithPathExpressions {
+
+    @RequestLine("GET /get/{id}")
+    String get(@Param("id") String id, String body);
+
+  }
+
+  @Test
+  public void clientMetricsHaveUriLabel() {
```
velo, line 51: Can we move this tests to the `AbstractMetricsTestBase`, likely this will fail for `dropwizard-metrics` which would need to be made consistent.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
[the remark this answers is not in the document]
martinacat: Thank you!

micrometer/src/test/java/feign/micrometer/MicrometerCapabilityTest.java
martinacat: I see - I originally placed it in `AbstractMetricsTestBase`, however I realised that `dropwizard-metrics` don't attach `uri` and thought that was a design decision. If necessary I can add `uri` on dropwizard 4 and 5 - is that what we would like? Thanks!

martinacat: I have moved the tests in `AbstractMetricsTestBase` and made `dropwizard-metrics` 4 & 5 consistent :)

velo (approved):

PR #1503, opened 2021-09-15
SOAPEncoder: Add support to modify soap message manually

Laess3r:
This small extension adds the possibility to manually set soap headers.
An example is given in the javadoc.

### DIFF
--- a/soap/src/main/java/feign/soap/SOAPEncoder.java
+++ b/soap/src/main/java/feign/soap/SOAPEncoder.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2020 The Feign Authors
+ * Copyright 2012-2021 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


velo:
Ok, I will take a look once build is fixed.

Laess3r:
Thanks @velo ! I fixed the build 😄

soap/src/main/java/feign/soap/SOAPEncoder.java
```diff
@@ -124,6 +124,9 @@ public void encode(Object object, Type bodyType, RequestTemplate template) {
           Boolean.toString(writeXmlDeclaration));
       soapMessage.setProperty(SOAPMessage.CHARACTER_SET_ENCODING, charsetEncoding.displayName());
       soapMessage.getSOAPBody().addDocument(document);
+
+      modifySOAPMessage(soapMessage);
```
velo, line 128: can we change this to 
`soapMessage = modifySOAPMessage(soapMessage);`

Assuming that would be help in case someone want to build a `SOAPMessage` from scratch

velo (approved):
Overall looks good.

I suggested one change, but not sure how helpful it would be, please lemme know.

Wondering why you had to change headers in so many classes....

soap/src/main/java/feign/soap/SOAPEncoder.java
Laess3r: Great idea! I like it.

Regarding the changed headers: I updated them using the maven plugin. Maybe some of them were not required to be updated. Should I revert them?

velo:
No need, no big deal

velo (approved):

PR #1509, opened 2021-09-20
support setting request factory in Google http client

doumdoum:
Hi,
This is a very minor change to support injecting the request factory into the client.


velo (approved):
Lgtm

Issue #1513, opened 2021-09-27
Cache support


Sam-Kruglov:
Hi, I've been battling with getting Spring's `@Cacheable` annotation to work and actually won!
I saw your roadmap to implement this feature in "short term" (updated 2 years ago).

Here's my workaround, maybe it will be helpful to someone with a similar issue or give an idea to contributors:

```java
import feign.Capability;
import feign.InvocationHandlerFactory;
import feign.InvocationHandlerFactory.MethodHandler;
import feign.Target;
import org.aopalliance.intercept.MethodInvocation;
import org.springframework.cache.interceptor.CacheInterceptor;
import org.springframework.stereotype.Component;

import java.lang.reflect.AccessibleObject;
import java.lang.reflect.Method;
import java.util.Map;

import static java.util.stream.Collectors.toMap;

/**
 * Allows Spring's @Cache* annotations to be declared on the feign client's methods
 */
@Component
public class CachingCapability implements Capability {
    private final CacheInterceptor cacheInterceptor;

    public CachingCapability(CacheInterceptor cacheInterceptor) {
        this.cacheInterceptor = cacheInterceptor;
    }

    @Override
    public InvocationHandlerFactory enrich(InvocationHandlerFactory invocationHandlerFactory) {
        return (target, dispatch) -> invocationHandlerFactory.create(target, withCacheProxy(target, dispatch));
    }

    private Map<Method, MethodHandler> withCacheProxy(Target<?> target, Map<Method, MethodHandler> dispatch) {
        return dispatch.entrySet().stream()
                .map(e -> Map.entry(e.getKey(), (MethodHandler) argv -> cacheInterceptor.invoke(new MethodInvocation() {
                    @Override
                    public Method getMethod() {
                        return e.getKey();
                    }

                    @Override
                    public Object[] getArguments() {
                        return argv;
                    }

                    @Override
                    public Object proceed() throws Throwable {
                        return e.getValue().invoke(argv);
                    }

                    @Override
                    public Object getThis() {
                        return target;
                    }

                    @Override
                    public AccessibleObject getStaticPart() {
                        return e.getKey();
                    }
                }))).collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
    }
}
```

As I was searching for a solution, apparently annotations in java interfaces don't get propagated onto the implementations, but looks like feign takes care of that with a proxy since it needs to read the annotations anyway.
[...]

PR #1515, opened 2021-09-28
Add InvocationHandlerFactory.MethodHandlerCustomizer

Sam-Kruglov:
This should make it easier for intergrations. Currently have to provide a capability to enrich the InvocationHandlerFactory itself to rebuild the "dispatch" map.

see #1513


core/src/main/java/feign/Feign.java
```diff
@@ -289,7 +310,8 @@ public Feign build() {
       ParseHandlersByName handlersByName =
           new ParseHandlersByName(contract, options, encoder, decoder, queryMapEncoder,
               errorDecoder, synchronousMethodHandlerFactory);
-      return new ReflectiveFeign(handlersByName, invocationHandlerFactory, queryMapEncoder);
+      return new ReflectiveFeign(handlersByName, invocationHandlerFactory,
+              queryMapEncoder, methodHandlerCustomizers);
```
Sam-Kruglov, line 314: this customizer basically acts just like one of `enrich` methods of `Capability` but I think it's too low level to be in there. Or is it?

core/src/main/java/feign/ReflectiveFeign.java
```diff
@@ -47,18 +50,23 @@
...
+      if (Util.isDefault(method)) {
+        DefaultMethodHandler handler = new DefaultMethodHandler(method, target, methodHandlerCustomizers);
         defaultMethodHandlers.add(handler);
         methodToHandler.put(method, handler);
       } else {
-        methodToHandler.put(method, nameToHandler.get(Feign.configKey(target.type(), method)));
+        MethodHandler handler = nameToHandler.get(Feign.configKey(target.type(), method));
+        for (MethodHandlerCustomizer customizer : methodHandlerCustomizers) {
+          handler = customizer.customize(target, method, handler);
+        }
```
Sam-Kruglov, line 68: this for loop is repeated twice, I could add a static method `MethodHandlerCustomizer#customizeAll(List< MethodHandlerCustomizer>, Target, Method, MethodHandler)` to make it a one-liner.

velo (changes requested):
This is duplicating the efforts of capabilities, but in a different format.

Let's refactor this as part of capability API

Sam-Kruglov:
> This is duplicating the efforts of capabilities, but in a different format.
> 
> Let's refactor this as part of capability API

Are you referring to [this](https://github.com/OpenFeign/feign/pull/1515#discussion_r718199903)?
So, I should convert MethodHandlerCustomizer into a method `Capability#enrich`?

Sam-Kruglov:
If it was an `enrich` method, it would be different. Every other method contains only one argument, but this one would contain more, as it also needs `target` and `method`

Sam-Kruglov:
@kdavisk6 if you look at #1513, I did it using capability but I thought we could add a more specific customizer just for the method handlers. Also, the capability solution does not cover `DefaultMethodHandler` as it's being used outside the factory (see lines 72 and 77 here):

https://github.com/OpenFeign/feign/blob/63069fd5a8fd37537d7a24bd459a30a864014c84/core/src/main/java/feign/ReflectiveFeign.java#L72-L78

kdavisk6:
Default Method handler proxies `default` implementations and cannot be enriched.  This is by design.

velo:
Hi Sam,

You probably trying to achieve the same as we have for metrics, don't you?
https://github.com/OpenFeign/feign/blob/master/dropwizard-metrics5/src/main/java/feign/metrics5/MeteredInvocationHandleFactory.java#L63

Sounds like there is no need to change feign at all

PR #1529, opened 2021-10-26
Fix standard decoders trying to decode body on 404 with decode404 set

### LINKED ISSUE #1528: Standard Decoders of OpenFeign don't return null if decode404 is set and a body is present
With version 10.6 of OpenFeign, the behavior of the standard Decoders (GsonDecoder, JacksonJaxbJsonDecoder, JacksonDecoder,  JacksonIteratorDecoder,  JAXBDecoder, SAXDecoder) was changed such that they return null (instead of, in case of arrays or Collections, empty arrays or Collections) on a 404 Not Found return code. See [OpenFeign PR 1012](https://github.com/OpenFeign/feign/pull/1012).

This was achieved by removing checking for 404 in the if condition that checks for 204 (no content) and returns an "Empty" version of the required type (empty arrays/collections/..., or null for others).

After that if condition, there's a check for an empty (null) body, and if so, null is returned.

This ignores that 404 does not mean "no content", instead, a descriptive error message (pretty HTML for HTML pages, or some detailed error message such as "user xy not found", written as a problem.json or an appropriate xml structure) can be there and will thus pass that "body == null" check, meaning that error message is then decoded with whatever means the decoder at hand provides - most likely resulting in a DecodeException (with, in case of JAXBDecoder, an UnmarshalException as its cause).

I propose (PR will follow) to add a new check for 404, and return null in these cases.

Note that without the decode404 flag, the regular decoders are not used if a 404 is returned, but instead the error decoder will be used.

ugrepel:
If decode404 flag is set, according to documentation decoders are supposed to return null
on 404 (not found) return codes. Version 10.6 of OpenFeign changed behavior such that
non-empty bodies of 404 responses are decoded by the standard decoders (GsonDecoder,
JacksonJaxbJsonDecoder, JacksonDecoder, JacksonIteratorDecoder, JAXBDecoder, SAXDecoder),
eventually causing DecodeExceptions if the body does not honor the expected data format.
Only on empty bodies, 404 always was honored with a null return value.

Fixes #1528

### DIFF
--- a/gson/src/main/java/feign/gson/GsonDecoder.java
+++ b/gson/src/main/java/feign/gson/GsonDecoder.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2012-2020 The Feign Authors
+ * Copyright 2012-2021 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
--- a/gson/src/main/java/feign/gson/GsonDecoder.java
+++ b/gson/src/main/java/feign/gson/GsonDecoder.java
@@ -43,7 +43,7 @@ public class GsonDecoder implements Decoder {
 
   @Override
   public Object decode(Response response, Type type) throws IOException {
-    if (response.body() == null)
+    if (response.status() == 404 || response.body() == null)
       return null;
     Reader reader = response.body().asReader(UTF_8);
     try {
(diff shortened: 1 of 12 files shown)


radio-rogal:
Can we mark `decode404` as `deprecated` but add `dismiss404` option?

`Dismiss404` means that decoder return empty/null value as the documentation says.

kdavisk6:
`decode404` has been misunderstood for a long time.  The intent is to allow users to force `404` results into whatever decoder has been declared on the pipeline instead of routing to the error decoder, as is the default behavior for all requests not in `1xx`, `2xx`, and `3xx` ranges.

Folks over the years have overloaded this intent.  I'm OK with what's being discussed, however, we should reconsider this option entirely and come up with a more obvious and explicit way for users to decide, by status, how a response should be decoded.

PR #1537, opened 2021-11-08
[GH-1319] Add Support for Path Style Parameter Expansion

### LINKED ISSUE #1319: Support unencoded reserved characters in path segments
Some time ago, we had support for  [MatrixVariables](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-ann-matrix-variables) added to Spring Cloud OpenFeign, using this [processor](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/annotation/MatrixVariableParameterProcessor.java).

This implementation works ok while sending requests to  Spring REST controllers that can handle incorrectly encoded `;` and `=` characters for matrix variables. However, an issue has been reported [here](https://github.com/spring-cloud/spring-cloud-openfeign/issues/395) that causes problems for servers that will not handle incorrect encoding of matrix variable reserved characters. 

Basically, since matrix variables can appear in any path segment, they are handled by us as path params, with values stored in `indexToExpander` and get fully pct-encoded by feign-core, including their special characters, so we get `/server/matrixParams%3Baccount%3Da%3Bname%3D` instead of `/server/matrixParams;account=a;name=n`.

It looks like an issue that can only be addressed within feign-core and not as part of our integration. I could work on a fix if that's fine with you. An idea that comes to mind is to introduce some kind of markers, say `{{}}` that would surround characters that should not be encoded within expanded path param chunks, but I'm also open to work on any other solution that the team might prefer. Let me know what you think about it.

kdavisk6:
Fixes #1319

This change adds limited Path Style support to Feign URI template-style
templates.  Variable expressions that start with a semi-colon `;`
are now expanded in accordance to [RFC 6570 Section 3.2.7](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.7)
with the following modifications:

* Maps and Lists are expanded by default.
* Only Single variable templates are supported.

Examples:
```
{;who}             ;who=fred
{;half}            ;half=50%25
{;empty}           ;empty
{;list}            ;list=red;list=green;list=blue
{;keys}            ;semi=%3B;dot=.;comma=%2C
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,38 @@ resolved values.  *Example* `owner` must be alphabetic. `{owner:[a-zA-Z]*}`
 * Unresolved expressions are omitted.
 * All literals and variable values are pct-encoded, if not already encoded or marked `encoded` via a `@Param` annotation.
 
+We also have limited support for Level 3, Path Style Expressions, with the following restrictions:
+
+* Maps and Lists are expanded by default.
+* Only Single variable templates are supported.
+
+*Examples:*
+
+```
+{;who}             ;who=fred
+{;half}            ;half=50%25
+{;empty}           ;empty
+{;list}            ;list=red;list=green;list=blue
+{;map}             ;semi=%3B;dot=.;comma=%2C
+```
+
+```java
+public interface MatrixService {
+
+  @RequestLine("GET /repos{;owners}")
+  List<Contributor> contributors(@Param("owners") List<String> owners);
+
+  class Contributor {
+    String login;
+    int contributions;
+  }
+}
+```
+
+If `owners` in the above example is defined as `Matt, Jeff, Susan`, the uri will expand to `/repos;owners=Matt;owners=Jeff;owners=Susan` 
+
+For more information see [RFC 6570, Section 3.2.7](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.7)
+
 #### Undefined vs. Empty Values ####
 
 Undefined expressions are expressions where the value for the expression is an explicit `null` or no value is provided.
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -13,32 +13,17 @@
  */
 package feign.template;
 
+import feign.Util;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import feign.Util;
 
 public final class Expressions {
-  private static Map<Pattern, Class<? extends Expression>> expressions;
 
-  static {
-    expressions = new LinkedHashMap<>();
-
-    /*
-     * basic pattern for variable names. this is compliant with RFC 6570 Simple Expressions ONLY
-     * with the following additional values allowed without required pct-encoding:
-     *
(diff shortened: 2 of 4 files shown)


velo (approved):
LGTM, it probably would be nice to have a Feign client example using this param expansion

kdavisk6:
Agreed.  I'll update.  I'm also going to export this expression as an `Expander` so it can be used directly be extending contracts.

dpodder:
Hi @kdavisk6 - thank you for working on this support!

One project I'm working on has a use case where the server expects an optional Map of `;key=value` pairs - such that the values are optional. E.g., assuming

```
/path/from{;map}/values
```

Here are some expected expansions:

* Case 1 - `map := {}` (empty)

  ```
  /path/from/values
  ```

* Case 2 - `map := {"a": "1", "b": "", "c": "3"}`

  ```
  /path/from;a=1;b=;c=3/values
  ```

  or:

  ```
  /path/from;a=1;b;c=3/values
  ```

Would these use cases work, or could the current solution be extended to support them?

kdavisk6:
@dpodder 

We'll adhere to the URI template specification:
https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.7

the spec defines these use cases:

```
Example Template     Expansion

       {;who}             ;who=fred
       {;half}            ;half=50%25
       {;empty}           ;empty
       {;v,empty,who}     ;v=6;empty;who=fred
       {;v,bar,who}       ;v=6;who=fred
       {;x,y}             ;x=1024;y=768
       {;x,y,empty}       ;x=1024;y=768;empty
       {;x,y,undef}       ;x=1024;y=768
       {;hello:5}         ;hello=Hello
       {;list}            ;list=red,green,blue
       {;list*}           ;list=red;list=green;list=blue
       {;keys}            ;keys=semi,%3B,dot,.,comma,%2C
       {;keys*}           ;semi=%3B;dot=.;comma=%2C
```

In the spec ,`empty` would result in the name being included.  To achieve the behavior you are looking for, the map would need to be `undefined`, which in our world would be `null`.  There are more hints in the spec that explain this behavior:

```
` else if an explode modifier is given, then

      *  if named is true, then for each defined list member or array
         (name, value) pair with a defined value, do:

         +  if this is not the first defined member/value, append the
            sep string to the result string;

         +  if this is a list, append the varname to the result string
            using the same encoding process as for literals;

         +  if this is a pair, append the name to the result string
            using the same encoding process as for literals;

         +  if the member/value is empty, append the ifemp string to the
            result string; otherwise, append "=" and the member/value to
            the result string, after pct-encoding any member/value
            characters that are not in the allow set.

      *  else if named is false, then

         +  if this is a list, append each defined list member to the
            result string, after pct-encoding any characters that are
            not in the allow set, with the sep string appended to the
            result between each defined list member.

         +  if this is an array of (name, value) pairs, append each pair
            with a defined value to the result string as "name=value",
            after pct-encoding any characters that are not in the allow
            set, with the sep string appended to the result between each
            defined pair.
```

dpodder:
@kdavisk6 - thank you for the clarification and details pointing out the relevant details from the spec! It looks like the `{;keys*}` example supported by the spec is exactly the behavior we're looking for, and would expand correctly for our cases.

PR #1543, opened 2021-11-27
HTTP protocol version on the logging

### LINKED ISSUE #1533: Feign logs HTTP/1.1 when OkHttpClient is using HTTP/2
We are using OkHttpClient with http2client but the logs always show HTTP/1.1

https://github.com/OpenFeign/feign/blob/3a351fa0716c51b5b4d7caa0c5f72cd61f80445f/core/src/main/java/feign/Logger.java#L64

radio-rogal:
Fixes #1533

Changes:

- `logRequest` does not write protocol version as we do not know what a client will use 
- `logAndRebufferResponse` tries to write a protocol version if a response has it or writes _HTTP/unknown_ otherwise

A Feign client can add a protocol version when it creates a response.

Next Feign clients have hardcoded value of HTTP/1.1 or try to read it from HTTP client:

0. Default (hardcoded value)
1. Apache HTTP Client (hardcoded value)
2. HC5
3. Google HTTP Client (hardcoded value)
4. Java11
5. OkHttpClient

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -59,13 +59,13 @@ public interface Client {
 
   class Default implements Client {
 
-    private static final String HTTP_PROTOCOL = "1.1";
+    private static final String HTTP_PROTOCOL = "HTTP/1.1";
     private final SSLSocketFactory sslContextFactory;
     private final HostnameVerifier hostnameVerifier;
 
     /**
      * Disable the request body internal buffering for {@code HttpURLConnection}.
-     * 
+     *
      * @see HttpURLConnection#setFixedLengthStreamingMode(int)
      * @see HttpURLConnection#setFixedLengthStreamingMode(long)
      * @see HttpURLConnection#setChunkedStreamingMode(int)
@@ -74,7 +74,7 @@ public interface Client {
 
     /**
      * Create a new client, which disable request buffering by default.
-     * 
+     *
      * @param sslContextFactory SSLSocketFactory for secure https URL connections.
      * @param hostnameVerifier the host name verifier.
      */
@@ -86,7 +86,7 @@ public interface Client {
 
     /**
      * Create a new client.
-     * 
+     *
      * @param sslContextFactory SSLSocketFactory for secure https URL connections.
      * @param hostnameVerifier the host name verifier.
      * @param disableRequestBuffering Disable the request body internal buffering for
@@ -140,7 +140,7 @@ public interface Client {
         }
       }
       return Response.builder()
-          .protocol(HTTP_PROTOCOL)
+          .protocolVersion(HTTP_PROTOCOL)
           .status(status)
           .reason(reason)
           .headers(headers)
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -59,6 +59,7 @@ public interface Client {
 
   class Default implements Client {
 
+    private static final String HTTP_PROTOCOL = "1.1";
     private final SSLSocketFactory sslContextFactory;
     private final HostnameVerifier hostnameVerifier;
 
@@ -139,6 +140,7 @@ public interface Client {
         }
       }
       return Response.builder()
+          .protocol(HTTP_PROTOCOL)
           .status(status)
           .reason(reason)
           .headers(headers)
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -59,8 +59,6 @@ public interface Client {
 
   class Default implements Client {
 
-    private static final String HTTP_PROTOCOL_VERSION = "HTTP/1.1";
-
     private final SSLSocketFactory sslContextFactory;
     private final HostnameVerifier hostnameVerifier;
 
@@ -141,7 +139,6 @@ public interface Client {
         }
       }
       return Response.builder()
-          .protocolVersion(HTTP_PROTOCOL_VERSION)
           .status(status)
           .reason(reason)
           .headers(headers)
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -59,13 +59,14 @@ public interface Client {
 
   class Default implements Client {
 
-    private static final String HTTP_PROTOCOL = "HTTP/1.1";
+    private static final String HTTP_PROTOCOL_VERSION = "HTTP/1.1";
+
     private final SSLSocketFactory sslContextFactory;
     private final HostnameVerifier hostnameVerifier;
 
     /**
      * Disable the request body internal buffering for {@code HttpURLConnection}.
-     * 
+     *
      * @see HttpURLConnection#setFixedLengthStreamingMode(int)
      * @see HttpURLConnection#setFixedLengthStreamingMode(long)
      * @see HttpURLConnection#setChunkedStreamingMode(int)
@@ -74,7 +75,7 @@ public interface Client {
 
     /**
      * Create a new client, which disable request buffering by default.
-     * 
+     *
      * @param sslContextFactory SSLSocketFactory for secure https URL connections.
      * @param hostnameVerifier the host name verifier.
      */
@@ -86,7 +87,7 @@ public interface Client {
 
     /**
      * Create a new client.
-     * 
+     *
      * @param sslContextFactory SSLSocketFactory for secure https URL connections.
      * @param hostnameVerifier the host name verifier.
      * @param disableRequestBuffering Disable the request body internal buffering for
@@ -140,7 +141,7 @@ public interface Client {
         }
       }
       return Response.builder()
-          .protocolVersion(HTTP_PROTOCOL)
+          .protocolVersion(HTTP_PROTOCOL_VERSION)
           .status(status)
           .reason(reason)
           .headers(headers)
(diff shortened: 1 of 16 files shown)


core/src/main/java/feign/Client.java
```diff
@@ -59,6 +59,8 @@
 
   class Default implements Client {
 
+    private static final String HTTP_PROTOCOL_VERSION = "HTTP/1.1";
```
velo, line 62: To avoid/reduce impact on any clients out there, we should move this to response.... and make the default value for `protocolVersion` when unset.

So it's up for HTTP2 clients to update it.

core/src/main/java/feign/Response.java
```diff
@@ -29,6 +29,7 @@
   private final Map<String, Collection<String>> headers;
   private final Body body;
   private final Request request;
+  private final String protocolVersion;
```
velo, line 32: Wondering if want to make this an enum or we wan't to leave it open so people can have `FTPES` or any other protocols they want to

core/src/test/java/feign/LoggerTest.java
```diff
@@ -72,18 +74,18 @@ public LogLevelEmitsTest(Level logLevel, List<String> expectedMessages) {
       return Arrays.asList(new Object[][] {
           {Level.NONE, Collections.emptyList()},
           {Level.BASIC, Arrays.asList(
-              "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP/1.1",
+              "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP",
               "\\[SendsStuff#login\\] <--- HTTP/1.1 200 OK \\([0-9]+ms\\)")},
           {Level.HEADERS, Arrays.asList(
-              "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP/1.1",
```
velo, line 78: can we set the protocol so it matches the original log?

core/src/test/java/feign/LoggerTest.java
```diff
@@ -157,16 +201,16 @@ public ReadTimeoutEmitsTest(Level logLevel, List<String> expectedMessages) {
       return Arrays.asList(new Object[][] {
           {Level.NONE, Collections.emptyList()},
           {Level.BASIC, Arrays.asList(
-              "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP/1.1",
+              "\\[SendsStuff#login\\] ---> POST http://localhost:[0-9]+/ HTTP",
```
velo, line 204: same, can we make it match the original `HTTP/1.1`?

googlehttpclient/src/main/java/feign/googlehttpclient/GoogleHttpClient.java
```diff
@@ -112,6 +114,7 @@ private final Response convertResponse(final Request inputRequest,
       contentLength = inputResponse.getHeaders().getContentLength().intValue();
     }
     return Response.builder()
+        .protocolVersion(HTTP_PROTOCOL_VERSION)
```
velo, line 117: remove, let Response default value deal with this one.

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
```diff
@@ -186,6 +187,7 @@ Response toFeignResponse(HttpResponse httpResponse, Request request) throws IOEx
     }
 
     return Response.builder()
+        .protocolVersion(HTTP_PROTOCOL_VERSION)
```
velo, line 190: remove, let Response default value deal with this one.

mock/src/main/java/feign/mock/MockClient.java
```diff
@@ -68,6 +68,7 @@ public synchronized Response execute(Request request, Request.Options options)
     } else {
       responseBuilder = executeAny(request, requestKey);
     }
+    responseBuilder.protocolVersion("mock");
```
velo, line 71: really liked this idea of setting a different protocol for mock

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -124,6 +128,9 @@ protected Response toFeignResponse(Request request, HttpResponse<byte[]> httpRes
     final OptionalLong length = httpResponse.headers().firstValueAsLong("Content-Length");
 
     return Response.builder()
+        .protocolVersion(httpResponse.version().name()
```
velo, line 131: this seems to fragile.

BTW, can you add a comment showing that is the expect value for `version()`

Also, I don't think it would be too back to have protocol = `HTTP_1_1`.

If you reallt think this name transformation is important, can we change this to a `switch` and hardcoded responses, just to make it easier to read.

mock/src/main/java/feign/mock/MockClient.java
velo: BTW, call it `MOCK` to match HTTP case?!

okhttp/src/main/java/feign/okhttp/OkHttpClient.java
```diff
@@ -92,6 +96,9 @@ static Request toOkHttpRequest(feign.Request input) {
   private static feign.Response toFeignResponse(Response response, feign.Request request)
       throws IOException {
     return feign.Response.builder()
+        .protocolVersion(response.protocol().name()
```
velo, line 99: same as java11

slf4j/src/test/java/feign/slf4j/Slf4jLoggerTest.java
```diff
@@ -101,9 +101,9 @@ public void logRequestsAndResponses() throws Exception {
     slf4j.logLevel("debug");
     slf4j.expectMessages("DEBUG feign.Logger - [someMethod] A message with 2 formatting tokens."
         + System.lineSeparator() +
-        "DEBUG feign.Logger - [someMethod] ---> GET http://api.example.com HTTP/1.1"
+        "DEBUG feign.Logger - [someMethod] ---> GET http://api.example.com HTTP"
         + System.lineSeparator() +
-        "DEBUG feign.Logger - [someMethod] <--- HTTP/1.1 200 OK (273ms)"
+        "DEBUG feign.Logger - [someMethod] <--- HTTP/unknown 200 OK (273ms)"
```
velo, line 106: where did this unknown came from?  Can we make previous messages?

velo (changes requested):
Overall looks good. Main concern would be to revert all changes to expected log messages and make sure we would match what was there before.

I expecte with a default `HTTP/1.1` on the Response, all should work as before, reducing backward compatibility problems.

Thanks for this work, looks really nice

slf4j/src/test/java/feign/slf4j/Slf4jLoggerTest.java
radio-rogal: done

mock/src/main/java/feign/mock/MockClient.java
radio-rogal: done

httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java
radio-rogal: done

core/src/main/java/feign/Client.java
radio-rogal: done

core/src/test/java/feign/LoggerTest.java
radio-rogal: done

I have restored old log messages for request line.

core/src/test/java/feign/LoggerTest.java
radio-rogal: done

googlehttpclient/src/main/java/feign/googlehttpclient/GoogleHttpClient.java
radio-rogal: done

java11/src/main/java/feign/http2client/Http2Client.java
radio-rogal: I agree.
I have replaced the string regexp-building by looking for proper enum value by its `name()` and `toString()` values. This solution covers all available in **java11** values: `HTTP_1_1` and `HTTP_2`.

radio-rogal:
There is another idea. After changes `Request` has constant-like field protocolVersion with value HTTP_1_1.

What about to add before logging of a request [SynchronousMethodHandler.executeAndDecode](https://github.com/OpenFeign/feign/blob/79ba51019920d49a6a295dcc56785c59360fba4a/core/src/main/java/feign/SynchronousMethodHandler.java#L112) another call `client.prepareRequest`: the client can upgrade request's version if its wants.

core/src/main/java/feign/Response.java
radio-rogal: I think that we can use enum: there are limited list of protocol version - HTTP/1.0, HTTP/1.1, HTTP/2.0 (HTTP/2). In the future we can add new values if they are needed.

I have added `Request.ProtocolVersion` enum and `Util.enumForName` method that helps to find enum value by any of strings: _HTTP_1_0_, _HTTP/1.0_, _HTTP_2_, _HTTP/2.0_ etc.

velo (approved):
I liked, will leave to simmer for a few days, in case @kdavisk6 has anything to say about it

PR #1548, opened 2021-12-11
The counter indicator increases by 2 each time (on happy path)

### LINKED ISSUE #1480: MeteredClient.execute()， The counter indicator increases by 2 each time
企业微信截图_02a28ec3-5b59-4210-9dbe-a472c0ed07ae

"sample.stop(timer)" in the finally code block should be placed in "if(time == null)"

radio-rogal:
Fixes #1480 

We call `sample.stop(timer)` in the `finally` block and therefore the call in the `try-catch` block is unnecessary.


micrometer/src/test/java/feign/micrometer/MeteredClientTest.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.net.ssl.HttpsURLConnection;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MeteredClientTest {
```
velo, line 38: I know the existing tests are a bit painfull to read, but, could we move this to https://github.com/OpenFeign/feign/blob/master/micrometer/src/test/java/feign/micrometer/AbstractMetricsTestBase.java ?

That would also test any problems on dropwizard metric implementation

radio-rogal: Yeah, sure. I will do

radio-rogal: I have moved test to old abstract class. It works for mictometer: old client breaks the test and new one does not.

Dropwizard clients have another meter implementation and I cannot find how to repeat this case (two calls of meter/counter) there.

velo: Maybe the bug was limited to micrometer implementation, that is great

velo (approved):

PR #1555, opened 2021-12-23
Simplify SSL configuration within Feign builder

Hakky54:
This is a small refactoring to simplify the ssl configuration for the end-user. It takes away the verbosity to declare a new custom client when using ssl with the builder. So without these changes the following is required:

```java
Feign.builder()
    .client(new Client.Default(sslSocketFactory, hostnameVerifier))
    .build()
```
With the refactored client builder, the below would be possible:

```java
Feign.builder()
    .ssl(sslSocketFactory, hostnameVerifier)
    .build();
```

To make this possible I needed to include the property `requestBuffering` in the builder which can also passed on as an option:

```java
Feign.builder()
    .requestBuffering(false)
    .build();
```

### DIFF
--- a/core/src/main/java/feign/Feign.java
+++ b/core/src/main/java/feign/Feign.java
@@ -21,6 +21,9 @@ import feign.codec.Decoder;
 import feign.codec.Encoder;
 import feign.codec.ErrorDecoder;
 import feign.querymap.FieldQueryMapEncoder;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSocketFactory;
 import java.io.IOException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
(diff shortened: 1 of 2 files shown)


radio-rogal:
But what if a developer uses both `.ssl()` and `.client()`?

The `.ssl()` overwrites changes that `.client()` mades earlier.

velo (changes requested):
Sadly, this won't work with any of the other existing clients for feign.

Can't do it

Hakky54:
I didn't analysed the code base well enough. With my code changes a new default client would be created when using the sslsocketfactory and the other will be overridden.

Next to that I noticed that there are multiple implementations of the Client interface. Every implementation might have a different SSL configuration and therefor it would not make sense to have these methods in the builder.

So I propose to close this PR 😅

PR #1557, opened 2021-12-24
Fixed dangling javadoc comment warning for license header

Hakky54:
I was scrolling through the source code and I noticed a warning that the license header is marked as `dangling javadoc comment` I did some research and found out that the better way is to have a license header in the following format:

```
/*
 *
 */
```

Instead of:
```
/**
 *
 */
```

I don't think the existing header is incorrect formatted but my idea gave warnings and this was a way to resolve it. Not quite sure if you guys want this change. Because I change the format the copyright date needed to be updated to 2021 because of the maven license plugin scan. Looking to hear from you soon.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2020 The Feign Authors
+# Copyright 2012-2021 The Feign Authors
 #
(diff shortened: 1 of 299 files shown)


velo:
The PR ia massive and risk introducing unwanted changes.

If you want, just make the `pom.xml` change, and I will checkout the branch, apply `license:format` and push

PR #1571, opened 2022-01-29
Use charset of request to create the String content.

### LINKED ISSUE #1570: ApacheHttp5Client does not use charset from request to create the String content
ApacheHttp5Client does not use charset from request to create the String content.

akuhtz:
This PR fixes #1570


hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
```diff
@@ -135,7 +135,12 @@ ClassicHttpRequest toClassicHttpRequest(Request request, Request.Options options
         entity = new ByteArrayEntity(data, null);
       } else {
         final ContentType contentType = getContentType(request);
-        entity = new StringEntity(new String(data), contentType);
+        if (request.charset() != null) {
+          String content = new String(request.body(), request.charset());
```
velo, line 139: can we use `data` instead?
if not, need a comment explaining why need to use `request.body()`

hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
```diff
@@ -135,7 +135,12 @@ ClassicHttpRequest toClassicHttpRequest(Request request, Request.Options options
...
+          entity = new StringEntity(content, contentType);
```
velo, line 140: can we move this line outside the if, and let the if block just handle the `String content` creation?!

velo (changes requested):
Looking good, still need unti test for the change

hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
akuhtz: Done.

hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
akuhtz: Done.

velo:
Any chance of we having a unit test around this change?

akuhtz:
> Any chance of we having a unit test around this change?

I don't know how to manipulate the request.charset() to return a changed encoding, and I don't know how to change the default for a single test. The issue happens if e.g. the client has default charset ISO_8859_1 and the server has UTF_8.
The existing test already rely on UTF_8.

PR #1579, opened 2022-02-28
1497 : spring contract annotation alias handled

### LINKED ISSUE #1497: Spring 5 AliasFor handling
Spring 5 introduced aliases for annotations: annotation attributes can be declared as aliases for the default `value()` attribute, in order to be more expressive.
Details can be found here:
https://github.com/spring-projects/spring-framework/issues/16137

It means, that the following annotations are properly fine from a Spring point of view:
```java
package foo;
import org.springframework.web.bind.annotation.PathVariable;
import org.springframework.web.bind.annotation.GetMapping;

interface Foo {
   @GetMapping("/foo/{id}/path")
   FooModel getById(@PathVariable(name="id") String id);
}
```

Spring will correctly use `id` as a path variable here, without the `value()` attribute specified.

However, Feign would not, because the SpringContract class does not use Spring 5 semantics to handle aliased attributes, it always assumes, that the `value()` attribute is set.

This means, that properly annotated Spring 5 interfaces cannot be proxied correctly by Feign.

pavansharma36:
Fixes #1497

### DIFF
--- a/spring4/src/main/java/feign/spring/SpringContract.java
+++ b/spring4/src/main/java/feign/spring/SpringContract.java
@@ -17,6 +17,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Parameter;
 import java.util.ArrayList;
 import java.util.Collection;
+import feign.Util;
 import org.springframework.web.bind.annotation.*;
 import feign.DeclarativeContract;
 import feign.MethodMetadata;
@@ -101,17 +102,16 @@ public class SpringContract extends DeclarativeContract {
 
   private String parameterName(String firstPriority,
                                String secondPriority,
-                               Parameter parameter,
-                               String fallback) {
-    if (isNotEmpty(firstPriority)) {
+                               Parameter parameter) {
+    if (Util.isNotBlank(firstPriority)) {
       return firstPriority;
-    } else if (isNotEmpty(secondPriority)) {
+    } else if (Util.isNotBlank(secondPriority)) {
       return secondPriority;
     } else {
       if (parameter.isNamePresent()) {
         return parameter.getName();
       } else {
-        return fallback;
+        return firstPriority;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


spring4/src/main/java/feign/spring/SpringContract.java
```diff
@@ -37,44 +39,79 @@ public SpringContract() {
...
 
       if (requestMapping.method().length == 1)
         data.template().method(Request.HttpMethod.valueOf(requestMapping.method()[0].name()));
     });
 
 
     registerMethodAnnotation(GetMapping.class, (mapping, data) -> {
-      appendMappings(data, mapping.value());
+      String[] paths;
+      if (mapping.path().length > 0) {
```
velo, line 57: can we do something like this:

```
appendMappings(data, fallback(mapping.path(), mapping.value()));

private <E> E fallback(E value, E alternative) {
if(value == null) {
return alternative;
}
if(value instanceof String[] && value.lenght == 0) {
return alternative;
}

return value;
}
```

velo (changes requested):

spring4/src/main/java/feign/spring/SpringContract.java
```diff
@@ -86,21 +87,57 @@ public SpringContract() {
...
+  }
+
+  private String[] mapping(String[] firstPriority, String[] fallback) {
+    return firstPriority.length > 0 ? firstPriority : fallback;
+  }
+
+  private String parameterName(String firstPriority,
+                               String secondPriority,
+                               Parameter parameter,
+                               String fallback) {
```
velo, line 105: `falback` and `secondPriority` seems to be always the same value

spring4/src/main/java/feign/spring/SpringContract.java
```diff
@@ -86,21 +87,57 @@ public SpringContract() {
...
+      String name = parameterName(parameterAnnotation.name(), parameterAnnotation.value(),
+          parameter, parameterAnnotation.value());
       Collection<String> query = addTemplatedParam(data.template().queries().get(name), name);
       data.template().query(name, query);
       nameParam(data, name, paramIndex);
-    });
+    };
+  }
 
+  private boolean isNotEmpty(String s) {
```
velo, line 139: please use `Util.isNotBlank` instead

velo (changes requested):
I think we are almost there

velo (approved):

PR #1586, opened 2022-03-04
Add ability use case insensitive headers for default Client

selectee:


velo (approved):
It looks fine. In which situations this is used?

selectee:
Some of proxy like as HAproxy started reference to HTTP protocol where headers are case insensitive 

https://discourse.haproxy.org/t/haproxy-2-0-0-header/3930/4

PR #1590, opened 2022-03-07
Stream decoder optimize

MrocCyen:
This PR has the following optimizations:
1. StreamDecoder has a strong validation for 'Stream' type that it is not conducive to scaling, so removal can be considered.
2. Added a default implementation of the iterator encoder, providing the user with a choice, although users can also use their own iterator encoder.


core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -76,8 +79,16 @@ public Object decode(Response response, Type type)
         });
   }
 
-  public static StreamDecoder create(Decoder iteratorDecoder) {
```
velo, line 79: since feign is widely used, we need to keep backwards compatibility

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -35,35 +37,36 @@
...
     this.iteratorDecoder = iteratorDecoder;
   }
 
   @Override
   public Object decode(Response response, Type type)
       throws IOException, FeignException {
-    if (!(type instanceof ParameterizedType)) {
-      throw new IllegalArgumentException("StreamDecoder supports only stream: unknown " + type);
+    if (!isStream(type)) {
+      return delegate.decode(response, type);
```
velo, line 65: can't/shouldn't we use the one decoder to handle both?!

Seems confusing to me to have 2 decoders

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -103,4 +114,13 @@ public Type getOwnerType() {
       return null;
     }
   }
+
+  public static class DefaultIteratorDecoder implements Decoder {
```
velo, line 118: What if we call it `LineToIteratorDecoder`?

velo (changes requested):

core/src/main/java/feign/stream/StreamDecoder.java
MrocCyen: This looks pretty good.

core/src/main/java/feign/stream/StreamDecoder.java
MrocCyen: Thanks, I will try to modify it.

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -35,35 +35,41 @@
...
       throws IOException, FeignException {
-    if (!(type instanceof ParameterizedType)) {
-      throw new IllegalArgumentException("StreamDecoder supports only stream: unknown " + type);
+    if (!isStream(type)) {
+      return delegate.decode(response, type);
     }
-    ParameterizedType streamType = (ParameterizedType) type;
-    if (!Stream.class.equals(streamType.getRawType())) {
-      throw new IllegalArgumentException("StreamDecoder supports only stream: unknown " + type);
+    IteratorDecoder iteratorDecoder;
```
velo, line 64: Let's not do this.

If anyone want to use same decoder for both, they can pass both on the constructor.

core/src/main/java/feign/stream/StreamDecoder.java
velo: still breaking previous behaviour

velo (changes requested):

core/src/main/java/feign/stream/StreamDecoder.java
MrocCyen: Can that provide two constructors? one uses same decoder for both, the other uses two decoder for delegate and iterator?

MrocCyen: Another scenario, according to the return value of the 'iteratorDecoder.decode' to determine whether it is a 'Stream' type, and then go through different processing.

velo: Sure thing, as long you keep the original behavior

MrocCyen: Thanks, I get it.

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -43,27 +45,36 @@
...
  * </pre>
+ * 
+ * @author mroccyen
  */
 public final class StreamDecoder implements Decoder {
 
   private final Decoder iteratorDecoder;
 
-  StreamDecoder(Decoder iteratorDecoder) {
+  public StreamDecoder(Decoder iteratorDecoder) {
```
velo, line 55: just to make myself very clear, I'm not against your idea of having 2 decoders, I think that was really smart and adds tons of flexibility that this class don't have right now.

My only "issue" was changing the signature for `public static StreamDecoder create(Decoder iteratorDecoder) {`

The create, with single arg, needs to take the `iteratorDecoder` not the delegate decoder.

What if we did something like this:

```
private final Decoder iteratorDecoder;
private final Optional<Decoder> delegateDecoder;
```

```
if (!isStream(type)) {
  if(delegateDecoder.isEmpty()) {
    throw new IllegalArgumentException("StreamDecoder supports only stream: unknown " + type);
  } else {
    return delegateDecoder.get().decode(response, type);
  }
}
```

velo (approved):
I will merge as is if you are tired of the back and forward, but if not =)

core/src/main/java/feign/stream/StreamDecoder.java
MrocCyen: If there is only one constructor parameter like iteratorDecoder, then how can the deleteDecoder parameter be passed to StreamDecoder, via the setter method?

velo: You can have multiple `create()` methods.... I just want the original one to keep current functionality

since constructor is not public, you could change it to have 2 arguments and have one create with empty and other with a delegate decoder.

MrocCyen: Thanks, I will modify it like this.

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -43,27 +45,33 @@
  *   Stream<Contributor> contributors(@Param("owner") String owner, @Param("repo") String repo);
  * }</code>
  * </pre>
+ * 
+ * @author mroccyen
```
kdavisk6, line 49: We don't include author attributes.  Please remove this comment.

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -103,4 +123,17 @@ public Type getOwnerType() {
...
+   */
+  public static class LineToIteratorDecoder implements Decoder {
+
+    @Override
+    public Object decode(Response response, Type type) throws IOException {
+      BufferedReader bufferedReader = new BufferedReader(response.body().asReader(UTF_8));
+      return bufferedReader.lines().iterator();
+    }
+
+  }
```
kdavisk6, line 138: For these types of delegate style decoders, we don't typically include a default implementation of a delegate.  I recommend moving this into your test class to act as an example to users on how to provide your own delegate.

The risk here is if a User uses this implementation, we'll have to support it long term.

kdavisk6 (changes requested):
@MrocCyen I have some maintenance related feedback.

kdavisk6:
Thank you for updating.  If you could please add a section to the README with some documentation, this PR looks ready to be merged.

kdavisk6 (dismissed):
Please add some documentation to the Readme

velo (approved):

core/src/main/java/feign/stream/StreamDecoder.java
```diff
@@ -47,23 +52,27 @@
 public final class StreamDecoder implements Decoder {
 
   private final Decoder iteratorDecoder;
+  private final Decoder delegateDecoder;
```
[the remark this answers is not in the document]
MrocCyen: sure，i will updating it.

MrocCyen:
Excuse me, how is the situation of this pr review now?

velo (approved):

PR #1597, opened 2022-03-19
 All decoders follow rule: if a status is 404 it returns empty or null value

### LINKED ISSUE #1595: Feign client returns null for non-nullable return value
Hello,
I'm using a feign client that looks something like this:
```kotlin
@FeignClient("api", url = "\${api-url}", decode404 = true)
interface Client {

    @GetMapping("/api/article/{id}")
    fun getLatestRevision(
        @PathVariable("id") id: String,
    ): Optional<ArticleRevision>


    @GetMapping("/api/video/{id}/revision/{revision}")
    fun getRevision(
        @PathVariable("id") id: String,
        @PathVariable("revision") revision: Int
    ): ArticleRevision
}
```

The assumption that I'm making is that when the first method gets a 404 response, it should return `Optional.empty`, which it does. But since the return type of the second method is not nullable, I would assume that an exception would be thrown if the response is 404. However, this does not happen and the function just returns `null`. Is this expected behavior? Can I not use decode404 and expect a non-nullable return value to not be null?

radio-rogal:
Now some decoders follow this rule but other don't.

After 11.6 the flag `decode404` doesn't work as it is described in the documentation. In new implementation a response body will be ignored, not decoded. I propose to replace `decode404` with `dismiss404` following the comment https://github.com/OpenFeign/feign/pull/1529#issuecomment-952232211

Fixes #1595

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -68,7 +68,7 @@ public abstract class AsyncFeign<C> extends Feign {
 
     private Decoder decoder = new Decoder.Default();
     private ErrorDecoder errorDecoder = new ErrorDecoder.Default();
-    private boolean decode404;
+    private boolean dismiss404;
     private boolean closeAfterDecode = true;
 
     public AsyncBuilder() {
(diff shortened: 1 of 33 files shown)


radio-rogal:
Should I update the changelog?
This change is important but the changelog hasn't updated since 10.9.

velo:
> Should I update the changelog? This change is important but the changelog hasn't updated since 10.9.

Since we started doing GitHub releases, we pretty much stop updating the changelog

May be needs a final entry saying so

velo (approved):

ugrepel:
Cool, thanks for fixing this. I've got a question however: is there any reason why the decoders distinguish between 204/404 decoding to (if defined) an empty version of the required type (e.g. an empty List), and, with other status codes, a null body, decoding to null?

radio-rogal:
Good note.

I suppose that we should use `emptyValueOf` for any cases when a response body is `null` or cannot be decoded: 204, 205, 404 (optional) and just body is `null` with any other HTTP status code.

PR #1619, opened 2022-05-10
mvn license:format

cuishuang:


velo (changes requested):
Need to update headers in a few files so the build passes

`mvn license:format`

PR #1626, opened 2022-06-05
Support of capabilities for AsyncFeign

### LINKED ISSUE #1374: capabilities for async client
Am I misreading it or is it not possible to add capabilities for the async builder? Is it on the roadmap?

edudar:
An initial take of supporting capabilities for async clients. 
- All capabilities added to `AsyncBuilder` are propagated to `Builder`.
- It extends the `Capability` interface with enrichment for the `AsyncClient` interface. 
- It adds `SyncBased` (not a perfect name) for `AsyncClient` implementations that used `Client` inside.

This PR does not change existing, DropWizard and Micrometer, capabilities yet. I can work on that if the overall direction is ok.

Fixes #1374


core/src/main/java/feign/AsyncFeign.java
```diff
@@ -255,15 +267,18 @@ protected AsyncFeign(AsyncBuilder<C> asyncBuilder) {
     this.activeContext = new ThreadLocal<>();
 
     this.defaultContextSupplier = asyncBuilder.defaultContextSupplier;
-    this.client = asyncBuilder.client;
 
-    this.logLevel = asyncBuilder.logLevel;
-    this.logger = asyncBuilder.logger;
+    this.client = asyncBuilder.client instanceof AsyncClient.SyncBased
```
velo, line 271: Why not enrich a `PseudoAsync` client?!

velo (changes requested):
I'm unsure why `SyncBased` is needed and would prefer to avoid it, as is the kind of secret sauce people need to know to get feign working as expect...

core/src/main/java/feign/AsyncFeign.java
edudar: So far, I have not found a nicer way to distinguish between "true" async clients (Java11, HC5, OkHttp) and those that wrap client + executor. And that's why I added this `SyncBased` interface. Users don't need to worry about it unless they implement their own clients.

velo: But how can you `enrich` the underlying client?

edudar: Makes sense... From the builder perspective, a user sets `Client` for sync and `AsyncClient` for async. Then feign applies capabilities when the whole thing is being built. It does not know any details of async implementation, meaning there should be no enrichments before that. I'll return to the drawing board. Thanks!

velo: Sweet, if  you could just remove the `SyncBased`, I would love to include this on todays release

edudar: I removed it locally, but I don't think this change is ready for prime time. I've been looking at why I was thinking about double-enrichment in the first place...

In `AsyncFeign`, we keep some functionality in the class, but we also delegate some of it to `Feign.Builder`. Like 

```
public AsyncBuilder<C> invocationHandlerFactory(InvocationHandlerFactory invocationHandlerFactory) {
  builder.invocationHandlerFactory(invocationHandlerFactory);
  return this;
}
```

So when I added this:
```
public AsyncBuilder<C> addCapability(Capability capability) {
  builder.addCapability(capability);
  this.capabilities.add(capability);
  return this;
}
```
It provides capabilities to both Async and delegate (Sync).

When `AsyncFeign` is constructed, we call:
```
  asyncBuilder.builder.client(this::stageExecution);
```
It replaces `client` in the delegate(!) with the keep-this-future function. 

When we call `this.feign = asyncBuilder.builder.build();` a few lines later, we enrich `this::stageExecution` like it's a client because delegate capabilities are invoked in `.build()`. That does not make sense, though, as `this::stageExecution` is not a client but a future-keeper. And that's how we get double-enrichment.

velo: shouldn't we let the Capability implementation handle that?

Another alternative, we could change the regular `Builder` to no enrich the client if it is `Async`

micrometer/src/main/java/feign/micrometer/MeteredClient.java
```diff
@@ -40,7 +43,21 @@ public MeteredClient(Client client,
...
       MetricName metricName,
       MetricTagResolver metricTagResolver) {
-    this.client = client;
+    this.delegate = client;
+    this.meterRegistry = meterRegistry;
+    this.metricName = metricName;
+    this.metricTagResolver = metricTagResolver;
+  }
+
+  public MeteredClient(AsyncClient<Object> asyncClient, MeterRegistry meterRegistry) {
```
velo, line 52: Could you also do the same for dropwizard implementations?!

core/src/main/java/feign/AsyncFeign.java
edudar: > Another alternative, we could change the regular Builder to no enrich the client if it is Async

That's what I attempted in the latest commit. Looks similar to `forceDecoding` flag.

micrometer/src/main/java/feign/micrometer/MeteredClient.java
edudar: Yep, I was testing the waters with micrometer one.

velo (approved):

edudar:
@velo I'm working on the tests, and there's an issue with clients that are both sync and async. `Capability.invoke()` can't differentiate between them, so it finds the first `enrich()` that matches, which may be either sync or async, and goes with it.

core/src/main/java/feign/AsyncFeign.java
```diff
@@ -255,24 +267,27 @@ protected AsyncFeign(AsyncBuilder<C> asyncBuilder) {
...
         asyncBuilder.closeAfterDecode);
 
-    asyncBuilder.builder.client(this::stageExecution);
-    asyncBuilder.builder.decoder(this::stageDecode);
-    asyncBuilder.builder.forceDecoding(); // force all handling through stageDecode
-
-    this.feign = asyncBuilder.builder.build();
+    this.feign = asyncBuilder.builder
+        .client(this::stageExecution)
+        .skipClientEnrichment() // don't enrich the wrapper
```
velo, line 286: I don't like this trend...  what if we only enrich here, not on the other builder?

edudar: That's what these flag functions do - they tell that other builder not to enrich because we did it here. These three flags essentially do the same - that `AsyncFeign` will work differently. The problem is that I don't want to tie them to "forceDecoding" as it's shady logic. So I need something else. I might combine both "skip" flags into one.

core/src/main/java/feign/DelegatingClient.java
```diff
@@ -0,0 +1,62 @@
...
+package feign;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A class that combines both sync and async clients and is capable of executing requests for both
+ */
+public abstract class DelegatingClient<C> implements Client, AsyncClient<C> {
```
velo, line 23: I see much effort is been spent on trying to solve the "what if" scenario a client implements both interfaces....

I think for v1 we shouldn't worry about that at all.

First, cause the method invoked wont be the same.  So, I think we could change `Capability#invoke` to call all matching methods, not just the first. 

What do you think?

edudar: But we do have such clients already. `Http2Client` implements both. `OkHttpClient` does both as well. That's why I'm trying this... If I change `invoke` to call all methods, that will result in `new MeteredClient(new MeteredClient(okHttpClient))` because both `enrich(Client)` and `enrich(AsyncClient)` will do the same thing, and that won't be expected I think.

We can get away without this `DelegatingClient`. I extracted it because I need the same logic for all three metered clients with no other commonalities other than `feign-core`.

velo:
@edudar lemme know what you think of my change based on your work.  Feel free to undo it if you don't like it

edudar:
@velo Sorry, didn't have time to get back to this PR in the last few days, got swamped. BaseBuilder definitely makes sense as there are way too many similarities between the two to keep them separated.

velo:
As long you are happy with the change, all good =)

PR #1634, opened 2022-06-21
Alternative ResponseInterceptor attempt

velo:


fanthos:
Seems that I can have only single ResponseInterceptor for Feign client. It would be great if the ResponseInterceptor could be chained.

velo:
> Seems that I can have only single ResponseInterceptor for Feign client. It would be great if the ResponseInterceptor could be chained.

that indeed would be a nice improvement, feel free to start a PR for it

XiangdongHe:
Has the ResponseInterceptor been deleted? If it is deleted, are there any other methods to replace it

XiangdongHe:
I want to determine whether to throw a custom business exception based on the return result of the Feign call. Throwing a custom business exception in the Decoder will be swallowed by the AsyncResponseHandler. I want to use the ResponseInterceptor to discover that it has been deleted

Issue #1464, opened 2021-07-21
Truncated JSON header value with @Headers


daczczcz1:
Observed on versions: 11.2, 11.4
Not reproducible on version: 10.1.0

When passing a JSON string as `@Param` to `@Headers` the string gets truncated.

Example: 
```
{"some": "value", "other": "value"}
```

Gets truncated to:
```
{"some": "value", "other"}
```

Code to reproduce: https://github.com/daczczcz1/feign-bug-reproduction

vinz486:
The annotation based `@Header` should be different from the `.header(...)` method. A solution could be a `.headerRaw(...)` method.

dagerber:
This breaks our code. We're adding a custom JSON structure to a HEADER which now gets truncated.
Please add another method as suggested above which adds a plain string as value without parsing

JKomoroski:
Ran into this today after upgrading from an older version to 11.8

skpandey91:
Any solution to this problem?
We are blocked.

janson653:
any update?

suryamurali1991:
Blocked by the same problem. Any update or a workaround to resolve this please.

velo:
Hi @suryamurali1991 looking for volunteers to help fixing this issue.

SikoraAdam:
@velo I've seen the PR was close, any chance the issue gonna be attended soon?

velo:
@SikoraAdam it's still waiting for someone to make the necessary changes.  I'm willing to review it and get the release out of the door.  Feel free to take a look on the comments we left at the PR and make the necessary changes

Samylots:
I confirm that we are also impacted by this issue. 
I have temporarely forced our project to get the 10.12 version to make it work.

RussellTaylor83:
Same problem for us. To sidestep it temporarily we're pulling in feign-core 11.0

`<dependency>
<groupId>io.github.openfeign</groupId>
<artifactId>feign-core</artifactId>
<version>11.0</version>
</dependency>`

JKomoroski:
I ran a git bisect on this issue and narrowed it down to the following commit: [f8ad8670e3dc73862f52eec475464a301e8c41cc](https://github.com/OpenFeign/feign/commit/f8ad8670e3dc73862f52eec475464a301e8c41cc)

[...]

PR #1638, opened 2022-06-24
Attempt to fix truncated json literals in header values

sm0217:
Fix for #1464

Attempted a simple fix for the problem where If the value of header, body, queryparam etc starts with {, check if its a json literal before attempting to consider it as an expression. 

Assuming unit tests cover all the cases, this fix works as expected for json literals. Need help identifying if this will impact any other expressions.  

(Too many file changes because of license header. Changes only in feign core pom, Template and HeaderTemplateTest)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -33,12 +33,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.jvnet</groupId>
-      <artifactId>animal-sniffer-annotation</artifactId>
-      <optional>true</optional>
-    </dependency>
-
-    <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
     </dependency>
@@ -92,6 +86,33 @@
             <goals>
               <goal>test-jar</goal>
             </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.1.0</version>
+        <executions>
+          <execution>
+            <id>enforce-banned-dependencies</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <message>feign-core should never include any dependencies, this is a design choice to keep core light and extend functionality using modules</message>
+                  <excludes>
+                    <exclude>*:*:*:*:*:*</exclude>
+                  </excludes>
+                  <includes>
+                    <include>*:*:*:*:test:*</include>
+                  </includes>
+                </bannedDependencies>
+              </rules>
+              <fail>true</fail>
+            </configuration>
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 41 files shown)


sm0217:
@velo For https://github.com/OpenFeign/feign/issues/1464

core/pom.xml
```diff
@@ -38,6 +38,11 @@
       <optional>true</optional>
     </dependency>
 
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
```
velo, line 43: sorry mate, this is not going to happen.

feign core shall have no dependencies

velo (changes requested):
Sorry mate, we can't have `org.json` on core.... if that is really need, need to be extracted into a new API and an implementation outside core must be created using `org.json`, as for core, it must continue clean of dependencies.

core/pom.xml
sm0217: Alternate would be use regex to validate the json string without any external dependencies. I can already see a feign-json library which uses org.json to encode/decode json. I assume its the right place to add the json literal validation. But we would still need to add the feign-json library into feign-core to use the implementation.

velo: all modules have multiple dependecies.... `feign-core` does not

core/src/main/java/feign/template/Template.java
```diff
@@ -214,7 +216,7 @@ private void parseFragment(String fragment) {
       /* check to see if we have an expression or a literal */
       String chunk = tokenizer.next();
 
-      if (chunk.startsWith("{")) {
+      if (chunk.startsWith("{") && !isValidJsonLiteral(fragment)) {
```
sm0217, line 219: @velo This is the fix I think we need to resolve this issue. Check if the string is a json literal before parsing it as a template. So if you agree with this solution and I have to add this check here, what would be the best approach please? Open for suggestions.

velo: What about the regex option you proposed?

sm0217: I can try. The regex looks a bit complicated and yet to confirm if it breaks the template string.  Let me add that with a few tests and update the PR.

velo: feel free to pick any other name besides `expression`

kdavisk6: I prefer this approach. Let’s add a property to the header annotation indicating that it should be treated like a literal. What about naming the property “literal” and be a Boolean?

kdavisk6: It will avoid any regular expressions and simplify the logic to be if “literal” is true, just render otherwise, resolve the expression. There should be a “Literal” type I think in the template package. I think.

SikoraAdam: Please remember annotation isn't the only way of setting the header

sm0217: I agree. This approach is better and avoid any assumption about the expression/literal.

sm0217: Although adding a variable like literal in Headers annotation wouldn't always be right since headers values is a string array, it's possible one of header has an expression and one is a literal. A variable would only work out if we can tailor the new variable for every header added to the array. That will be an expensive change.

PR #1700, opened 2022-07-21
chore(deps): upgrades JAXB dependencies for feign-jaxb on Java17

smlgbl:


velo (approved):
Looking good

sandalovsergey:
@smlgbl all right, one more question

You have added `jakarta.xml.bind-api` if jdk >=17. See https://github.com/OpenFeign/feign/pull/1700/files#diff-849baa0d8ff067821d75f011846fa2b18607333f551aa5080d5fe28d62081539R85

This artifact contains `jakarta.xml.bind`-package only. My question is: why do we need this dependency?

If we look at the code, there is no `jakarta.xml.bind`-package usage. But there is `javax.xml.bind`-package usage. F.e. see
https://github.com/OpenFeign/feign/blob/master/jaxb/src/main/java/feign/jaxb/JAXBEncoder.java#L19
https://github.com/OpenFeign/feign/blob/master/soap/src/main/java/feign/soap/SOAPEncoder.java#L22

Did you add it in order to introduce a transitive dependency in clients code? AFAIK it is a bad practise to depend on transitive ones.

I ask you in case of this issue: https://github.com/OpenFeign/feign/issues/1918. It seems that it needs to add a separate module(f.e. feign-jaxb-jakarta) for support jakarta in feign-jaxb. So before this I want to understand the reason of this PR. What it had to change? Why can't we do without this dependency?

smlgbl:
I think you are right. It would be better to do it the way you suggest in #1918 
You can go ahead and undo the change.

Kudos for making sure your artifact stays clean. Seriously.

Issue #1565, opened 2022-01-08
OpenFeign doesn't work with kotlin suspend method


XhstormR:
I have an API interface where the methods are kotlin suspend methods:

```kotlin
@FeignClient(path = "/api/masker", contextId = "MaskerApi", name = Const.SERVICE, configuration = [FeignConfig::class])
interface MaskerApi {

    @PostMapping("/mask")
    suspend fun mask(
        @Valid @RequestBody maskRequest: MaskRequest,
    ): RestResponse<List<CharSequence>>

    @PostMapping("/matches")
    suspend fun matches(
        @Valid @RequestBody maskRequest: MaskRequest,
    ): RestResponse<List<Boolean>>
}
```

But when I run the program, the program throws an exception:
```
Caused by: java.lang.IllegalStateException: Method has too many Body parameters: public abstract java.lang.Object io.github.xhstormr.masker.web.api.masker.MaskerApi.matches(io.github.xhstormr.masker.model.masker.MaskRequest,kotlin.coroutines.Continuation)
Warnings:
- 
	at feign.Util.checkState(Util.java:121) ~[feign-core-11.7.jar:na]
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:142) ~[feign-core-11.7.jar:na]
	at org.springframework.cloud.openfeign.support.SpringMvcContract.parseAndValidateMetadata(SpringMvcContract.java:193) ~[spring-cloud-openfeign-core-3.1.0.jar:3.1.0]
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:65) ~[feign-core-11.7.jar:na]
	at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:151) ~[feign-core-11.7.jar:na]
	at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:49) ~[feign-core-11.7.jar:na]
	at feign.Feign$Builder.target(Feign.java:268) ~[feign-core-11.7.jar:na]
	at org.springframework.cloud.openfeign.DefaultTargeter.target(DefaultTargeter.java:30) ~[spring-cloud-openfeign-core-3.1.0.jar:3.1.0]
	at org.springframework.cloud.openfeign.FeignClientFactoryBean.loadBalance(FeignClientFactoryBean.java:373) ~[spring-cloud-openfeign-core-3.1.0.jar:3.1.0]
[...]

PR #1706, opened 2022-07-31
Support kotlin coroutines

wplong11:
Resolves: #1565

!! This is working PoC

Inspired by https://github.com/PlaytikaOSS/feign-reactive/pull/486

## TODO

- [ ] Separate Kotlin support module
- [ ] Enhance test case
- [ ] Refactoring
- [ ] Clean up pom.xml


velo:
This looks really nice.  It makes total sense to make this available on the `AsyncFeign` only, really good plan!

snyk is asking to "Upgrade org.jetbrains.kotlin:kotlin-stdlib to version 1.6.0 or higher"

Sounds like a solid plan, once you are ready to move to a new module, fee free to ping me for any assistance needed

wplong11:
@velo I pushed some changes

1. Refactor some code
   - Use `FutureKt.await` instead of [KotlinExtensions.awaitResponse](https://github.com/OpenFeign/feign/blob/80f960e32c4ce08760ecb52e9ae8f9c4499f15cc/core/src/main/java/feign/KotlinExtensions.kt#L12-L13)
   - Renamed KotlinExtensions to MethodKt
2. Update the Kotlin version
  - If the Feign library is using the latest version, and the Kotlin version of the project using the Feign library is lower, is it safe? I'm going to do some research on this, but I'm asking to check if you have any knowledge to share with me.

wplong11:
@velo I'm looking into other libraries to isolate the Kotlin support module so that the core module doesn't reference the Kotlin library, but Spring is using Kotlin in the core module. How important do you think the necessity is with regard to module separation?

kdavisk6:
I so happy someone picked this up. @velo could I have a day or two to review this as well?

velo:
> I’m so happy someone picked this up. @velo could I have a day or two to review this as well?

You absolutely can sir!

But, until this is moved to a module, I wouldn't be comfortable to merge, as core needs to be lean and have no dependencies.

kdavisk6:
> > I’m so happy someone picked this up. @velo could I have a day or two to review this as well?
> 
> You absolutely can sir!
> 
> But, until this is moved to a module, I wouldn't be comfortable to merge, as core needs to be lean and have no dependencies.

Exactly why I wanted to review. Like minds

wplong11:
@velo 

>  But, until this is moved to a module, I wouldn't be comfortable to merge, as core needs to be lean and have no dependencies.

I'm considering splitting modules using Optional Dependencies. What do you think about this approach? After a brief investigation, it seems that Spring Core also supports Kotlin using Optional Dependencies. I'm going to try implementing it today using optional dependencies, and investigate further if there are other approaches. (I live in the KST time zone (UTC+9), and I can write code after work.)

[Optional Dependencies](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html)

> Optional dependencies are used when it's not possible (for whatever reason) to split a project into sub-modules. The idea is that some of the dependencies are only used for certain features in the project and will not be needed if that feature isn't used. Ideally, such a feature would be split into a sub-module that depends on the core functionality project. This new subproject would have only non-optional dependencies, since you'd need them all if you decided to use the subproject's functionality.
> 
> However, since the project cannot be split up (again, for whatever reason), these dependencies are declared optional. If a user wants to use functionality related to an optional dependency, they have to redeclare that optional dependency in their own project. This is not the clearest way to handle this situation, but both optional dependencies and dependency exclusions are stop-gap solutions.

velo:
I don't think `optional` is going to cut.  But tell you what, get your changes to a place where you are happy with, and I can jump in and help refactoring the core to make a module possible

wplong11:
@velo I pushed some changes to share idea.

- Because the optional dependency is used, Kotlin support related libraries will not be referenced in a Java-only project.
- Assert.java / ClasUtils.java / KotlinDetector.java is from spring core project
- As an additional idea, I thought of separating the code related to Kotlin support into a separate module and referencing that module as an Optional Dependency from the core module.

core/src/main/java/feign/kotlinSupport/ClassUtils.java
```diff
@@ -0,0 +1,423 @@
...
+ * Miscellaneous {@code java.lang.Class} utility methods.
+ * Mainly for internal use within the framework.
+ *
+ * @author Juergen Hoeller
+ * @author Keith Donald
+ * @author Rob Harrop
+ * @author Sam Brannen
+ * @since 1.1
+ */
+public abstract class ClassUtils {
```
velo, line 35: This class (and others on the PR) look like copy and paste from somewhere.... please use the original library

wplong11: https://github.com/OpenFeign/feign/pull/1706#issuecomment-1202975619

> Assert.java / ClasUtils.java / KotlinDetector.java is from spring core project -> I will refactor when the implementation approach is agreed upon.

If the approach is agreed upon, Assert.java / ClasUtils.java used in KotlinDetector and KotlinDetector implementation will be refactored. These classes are from [Spring Core](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/ClassUtils.java). If I use the original library, I must use the Spring Core library.

How about reviewing where the feign code branches off using KotlinDetector for Kotlin support? I wrote the code only to demonstrate the approach.

velo:
Hi @wplong11 , so, as is, we can't merge, as we (Myself and @kdavisk6 ) won't approve any changes that introduce dependencies to feign core.... optional=>true is still a dependency, and that is not going to fly. Core needs ZERO dependencies. 

My plan is to refactor `feign-core` to allow for the new bits you created to be attached to core like we do with everything else.
Like, feign supports "jackson" but there is no jackson dependency on core. Could you imagine the caos it would bring to everybody if overnight feign-core included jackson, json, http commons, apache http5 and more and more

So, once you get your changes to a point that you are happy with, ping me and I will help you out.

core/src/main/java/feign/kotlinSupport/ClassUtils.java
wplong11: Assert.java -> Removed
ClasUtils.java / KotlinDetector.java -> Refactored

wplong11:
@velo  Looking at the final result, feign-core depends on feign-kotlin as an optional dependency. I think it's a bit better than relying on kotlin-related dependencies directly, but as you said, I'd also like to have zero dependencies if possible. But I have no idea how to implement it. Do you have any ideas?

Considerations
- People using feign only depend on `feign-kotlin` and do not require additional code.

velo:
>  Do you have any ideas?

Not yet.  But, I will get something done

> * People using feign only depend on `feign-kotlin` and do not require additional code.

Not really, it will bring in kotlin and whatever else it depends on.  As is, you can't use feign if kotlin library is not included.

wplong11:
@velo Looking at the Github Example project, it seems that feign-core is used with zero dependency. Am I misunderstanding something?

스크린샷 2022-08-03 오후 12 00 26

velo:
either that dependency tree wrong, or a runtime exception will happen when performing an async operation

wplong11:
@velo 

> either that dependency tree wrong, or a runtime exception will happen when performing an async operation

I checked that Github Example in Feign project runs correctly without kotlin related modules.

According to [Maven document](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html#how-do-optional-dependencies-work),

> Project-A -> Project-B
> The diagram above says that Project-A depends on Project-B. When A declares B as an optional dependency in its POM, this relationship remains unchanged. It's just like a normal build where Project-B will be added in Project-A's classpath.
>
> Project-X -> Project-A
>When another project (Project-X) declares Project-A as a dependency in its POM, the optional nature of the dependency takes effect. Project-B is not included in the classpath of Project-X. You need to declare it directly in the POM of Project X for B to be included in X's classpath.

Kotlin related modules is not included in the classpath of Github Example. Is there a difference between not being included on the classpath and excluded from the dependency tree?

velo:
@wplong11 I moved kotlin to a module, but couldn't get the tests working.... could you take a look, I must admit I don't know where to start

kotlin/src/test/kotlin/feign/kotlin/CoroutineFeignTest.kt
```diff
@@ -0,0 +1,184 @@
...
+
+        // Act
+        val firstOrder = client.findOrder3(orderId = 1)
+
+        // Assert
+        assertThat(firstOrder).isNull()
+    }
+
+    @Test
+    fun shouldRun4(): Unit = runBlocking {
```
velo, line 85: could we get meaningful names for tests?

wplong11:
Thanks for the additional work. I've been too busy lately. I'll be able to start working again in two weeks. I'll finish this PR soon.

velo:
Awesome, and thanks for this massive contribution.

I have o my mind now, that I need to find a way to have simpler `Feign#builder` classes..

wplong11:
Branch updated to latest state

```shell
$ git rebase master && git push --force
```

kotlin/src/test/kotlin/feign/kotlin/CoroutineFeignTest.kt
wplong11: Corrected. Please provide feedback on the naming convention.

example-github-with-coroutine/src/main/java/example/github/GitHubExample.kt
```diff
@@ -0,0 +1,133 @@
...
+                        "Authorization",
+                        "token 383f1c1b474d8f05a21e7964976ab0d403fee071");
+                }
+                .options(Request.Options(10, TimeUnit.SECONDS, 60, TimeUnit.SECONDS, true))
+                .target(GitHub::class.java, "https://api.github.com")
+        }
+    }
+}
+
+/** Lists all contributors for all repos owned by a user.  */
```
wplong11, line 108: Currently, when this extension function is declared inside the GitHub interface, a Contracts parsing error occurs.

```
Exception in thread "main" java.lang.IllegalStateException: Method GitHub#contributors(String,Continuation) not annotated with HTTP method type (ex. GET, POST)
Warnings:
- Class GitHub has annotations [Metadata] that are not used by contract Default
	at feign.Util.checkState(Util.java:136)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:115)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:65)
	at feign.DeclarativeContract.parseAndValidateMetadata(DeclarativeContract.java:38)
```


This is because it is recognized as a non-default method, so contracts are required. (The default method of the kotlin interface is compiled as an abstract method of the java interface.)

So, I should use `ReflectJvmMapping.getKotlinFunction(method).isAbstract() == false` to check if it is the default method.

However, it seems that many code modifications may be required to put this logic in the code below without using optional dependency.

https://github.com/OpenFeign/feign/blob/75a3c1cf6f0e57c2d1eaa5f8be2538489e605650/core/src/main/java/feign/Util.java#L143-L153

apt-test-generator/pom.xml
```diff
@@ -45,6 +45,12 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
```
velo, line 48: I don't think this is necessary

velo (approved):
@kdavisk6 last call =)

example-github-with-coroutine/src/main/java/example/github/GitHubExample.kt
velo: Yeah, forget the optional dependency, that ain't optional and ain't a option for feign core =)

The feign approach for this kind of problems is abstraction.  So would need an interface for `DefaultMethodsDetection/Handling/Operator`, a default implementation as above and a kotlin impl that does what kotlin needs

wplong11:
1. This Branch updated to latest master branch

    ```shell
    $ git rebase master && git push --force
    ```

2. Add a single commit [78589d5](https://github.com/OpenFeign/feign/pull/1706/commits/78589d5f4f9fa93e1388b6c90a0b03aefcc11501)

apt-test-generator/pom.xml
wplong11: Removed

example-github-with-coroutine/src/main/java/example/github/GitHubExample.kt
wplong11: I totally agree with the approach you are talking about. Let's consider improving it with a separate PR.

velo (approved):

PR #1713, opened 2022-08-07
Proof that clients support gzip and deflate compression

### LINKED ISSUE #1697: Only default client supports GZIP response
The issue exists in OpenFeign 11.8, but it was working in 10.x.

Client.java
-----------
    
    
    @Override
    public Response execute(Request request, Options options) throws IOException {
      HttpURLConnection connection = convertAndSend(request, options);
      return convertResponse(connection, request);
    }

    Response convertResponse(HttpURLConnection connection, Request request) throws IOException {
      int status = connection.getResponseCode();
      String reason = connection.getResponseMessage();
      ...
      InputStream stream;
      if (status >= 400) {
        stream = connection.getErrorStream();
      } else {
        if (this.isGzip(connection.getHeaderFields().get(CONTENT_ENCODING))) {
          stream = new GZIPInputStream(connection.getInputStream());
        } else if (this.isDeflate(connection.getHeaderFields().get(CONTENT_ENCODING))) {
          stream = new InflaterInputStream(connection.getInputStream());
        } else {
          stream = connection.getInputStream();
        }
      }
     ...
    }

ApacheHttpClient.java
-----------

```

public final class ApacheHttpClient implements Client {
  private static final String ACCEPT_HEADER_NAME = "Accept";

  private final HttpClient client;
  ...

  @Override
  public Response execute(Request request, Request.Options options) throws IOException {
    HttpUriRequest httpUriRequest;
    try {
      httpUriRequest = toHttpUriRequest(request, options);
    } catch (URISyntaxException e) {
      throw new IOException("URL '" + request.url() + "' couldn't be parsed into a URI", e);
    }
    HttpResponse httpResponse = client.execute(httpUriRequest);
    return toFeignResponse(httpResponse, request);
  }

```

If any client is used other than the `default` client, the GZIP content can't be decoded and results in an exception, because GZIP is read as simple InputStream, not GZIPInputStream

```

feign.codec.DecodeException: Error while extracting response for type [java.util.List<ProductDto>] and content type [application/json]; nested exception is org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens; nested exception is com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
```

radio-rogal:
Bunch of clients support gzip and deflate compression except of JAX-RS2, OkHTTP and Google HTTP client.

- OkHTTP client with an interceptor can proccess GZIP response, e.g. https://stackoverflow.com/questions/51901333/okhttp-3-how-to-decompress-gzip-deflate-response-manually-using-java-android
- Google HTTP client with NetHttpTransport does not support GZIP response at all but we can use Apache HTTP client as its transport
- JAX-RS2 client does not support GZIP responses out-of-the-box but we can register a class that implies it like Jersey implementation does it, see https://stackoverflow.com/questions/22519642/how-to-enable-gzip-compression-for-content-encoding-with-jersey-jax-rs-2-0-cli

Closes #1697

### DIFF
--- a/core/src/test/java/feign/client/AbstractClientTest.java
+++ b/core/src/test/java/feign/client/AbstractClientTest.java
@@ -14,7 +14,8 @@
 package feign.client;
 
 import static feign.Util.UTF_8;
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
 import static org.junit.Assert.assertEquals;
 import feign.Client;
 import feign.CollectionFormat;
@@ -28,13 +29,18 @@ import feign.Response;
 import feign.Util;
 import feign.assertj.MockWebServerAssertions;
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.zip.DeflaterOutputStream;
+import java.util.zip.GZIPOutputStream;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
+import okio.Buffer;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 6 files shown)


radio-rogal:
I think this pull request could be declined and the issues is closed as it is just demo to proof all works as expected.

velo (approved):

velo:
> I think this pull request could be declined and the issue is closed as it is just demo to proof all works as expected.

More testing always better

PR #1726, opened 2022-08-15
feat: fix default retryer bug where initial wait period is the provid…

FpMarinov:
…ed period multiplied by 1.5, do a refactor on default retryer

### DIFF
--- a/core/src/main/java/feign/Retryer.java
+++ b/core/src/main/java/feign/Retryer.java
@@ -15,6 +15,8 @@ package feign;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
+import java.util.Date;
+
 /**
  * Cloned for each invocation to {@link Client#execute(Request, feign.Request.Options)}.
  * Implementations may keep state to determine if retry operations should continue or not.
@@ -47,35 +49,38 @@ public interface Retryer extends Cloneable {
       this.attempt = 1;
     }
 
-    // visible for testing;
-    protected long currentTimeMillis() {
-      return System.currentTimeMillis();
-    }
-
+    @Override
     public void continueOrPropagate(RetryableException e) {
-      if (attempt++ >= maxAttempts) {
+      if (attempt >= maxAttempts) {
         throw e;
       }
 
       long interval;
-      if (e.retryAfter() != null) {
-        interval = e.retryAfter().getTime() - currentTimeMillis();
-        if (interval > maxPeriod) {
-          interval = maxPeriod;
-        }
+      if (e.retryAfter() == null) {
+        interval = nextMaxInterval();
+      } else {
+        interval = intervalFromRetryAfter(e.retryAfter());
+
         if (interval < 0) {
+          attempt++;
           return;
         }
-      } else {
-        interval = nextMaxInterval();
       }
+
       try {
         Thread.sleep(interval);
       } catch (InterruptedException ignored) {
         Thread.currentThread().interrupt();
         throw e;
       }
+
       sleptForMillis += interval;
+      attempt++;
+    }
+
+    @Override
+    public Retryer clone() {
+      return new Default(period, maxPeriod, maxAttempts);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


velo (commented):
Change makes sense, just needs to get build passing

FpMarinov:
> Change makes sense, just needs to get build passing

This seems to be the cause of the problem, the weird thing is the test passes fine locally:

[ERROR] loadBalancingTargetPath(feign.ribbon.LoadBalancingTargetTest)  Time elapsed: 0.825 s  <<< ERROR!
feign.RetryableException: Connection refused executing GET http://localhost:53935/context-path/servers

at feign.ribbon.LoadBalancingTargetTest.loadBalancingTargetPath(LoadBalancingTargetTest.java:83)

Caused by: java.net.ConnectException: Connection refused


Do you have any ideas what could be going wrong?

kdavisk6:
Per #1727

core/src/main/java/feign/Retryer.java
```diff
@@ -47,35 +48,38 @@ public Default(long period, long maxPeriod, int maxAttempts) {
...
 
-    // visible for testing;
-    protected long currentTimeMillis() {
-      return System.currentTimeMillis();
-    }
-
+    @Override
     public void continueOrPropagate(RetryableException e) {
-      if (attempt++ >= maxAttempts) {
+      if (attempt >= maxAttempts) {
```
kdavisk6, line 53: This change is not backward compatible.  We count the attempt before sleeping intentionally.  This change will reduce the attempts by 1, please revert this change.

core/src/main/java/feign/Retryer.java
```diff
@@ -47,35 +48,38 @@ public Default(long period, long maxPeriod, int maxAttempts) {
...
-        if (interval > maxPeriod) {
-          interval = maxPeriod;
-        }
+      if (e.retryAfter() == null) {
+        interval = nextMaxInterval();
+      } else {
+        interval = intervalFromRetryAfter(e.retryAfter());
+
         if (interval < 0) {
+          attempt++;
```
kdavisk6, line 64: See my previous comment about counting attempts.

core/src/main/java/feign/Retryer.java
```diff
@@ -47,35 +48,38 @@ public Default(long period, long maxPeriod, int maxAttempts) {
...
+
       try {
         Thread.sleep(interval);
       } catch (InterruptedException ignored) {
         Thread.currentThread().interrupt();
         throw e;
       }
+
       sleptForMillis += interval;
+      attempt++;
```
kdavisk6, line 77: One more for the attempt count.

ribbon/src/test/java/feign/ribbon/RibbonClientTest.java
```diff
@@ -339,7 +340,7 @@ public void testFeignOptionsNoFollowRedirect() {
     try {
       Response response = api.get();
       assertEquals(200, response.status());
-      assertEquals("Hello", response.body().toString());
+      assertEquals("Hello", new Scanner(response.body().asInputStream()).nextLine());
```
kdavisk6, line 343: There should be no need for a `Scanner` here.

kdavisk6: Also, this test is ignored.

core/src/test/java/feign/RetryerTest.java
```diff
@@ -91,7 +91,7 @@ public void defaultRetryerFailsOnInterruptedException() {
       Assert.fail("Retryer continued despite interruption");
     } catch (RetryableException e) {
       Assert.assertTrue("Interrupted status not reset", Thread.interrupted());
-      Assert.assertEquals("Retry attempt not registered as expected", 2, retryer.attempt);
+      Assert.assertEquals("Retry attempt not registered as expected", 1, retryer.attempt);
```
kdavisk6, line 94: Last comment related to attempts.

kdavisk6 (changes requested):
Changing how we count the number of attempts will break backward compatibility.  The original implementation is intentional.

PR #1737, opened 2022-09-01
Only allow a single content-type header to be applied while using googlehttpclient

skrzepto:
## Impact
Fixed issue where multiple content-type headers were being added to the request while using googlehttpclient.

## Research
Multiple `Content-Type` headers goes against the spec

> Content-Type is defined as a singleton field.... 
https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3-7

## Implementation Summary
Ignore `Content-Type` from being set on the request because it'll be added via ByteArrayContent


velo (approved):

skrzepto:
@velo Thank you so much for merging in this pr! I was wondering when I can expect to see a build with this bug fix and if there's anything I can do to help.

Thanks!

PR #1754, opened 2022-09-18
Refactor coroutine feign

wplong11:

Eliminates the possibility of breaking changes following AsyncFeign's refactoring in advance

## TODO

- [ ] Modify CoroutineFeign.CoroutineBuilder to extend Feign.Builder
    - It is intended to be used when [customizing](https://docs.spring.io/spring-cloud-openfeign/docs/current/reference/html/#spring-cloud-feign-circuitbreaker) Feign.Builder in Spring Cloud OpenFeign.
    - I think it's okay to work after merged, because CoroutineFeign is an experimental feature right now.

### DIFF
--- a/kotlin/src/main/java/feign/kotlin/CoroutineFeign.java
+++ b/kotlin/src/main/java/feign/kotlin/CoroutineFeign.java
@@ -33,8 +33,16 @@ import kotlinx.coroutines.future.FutureKt;
 
 @Experimental
 public class CoroutineFeign<C> {
-  public static <C> CoroutineBuilder<C> coBuilder() {
+  public static <C> CoroutineBuilder<C> builder() {
     return new CoroutineBuilder<>();
+  }
+
+  /**
+   * @deprecated use {@link #builder()} instead.
+   */
+  @Deprecated()
+  public static <C> CoroutineBuilder<C> coBuilder() {
+    return builder();
   }
 
   private static class LazyInitializedExecutorService {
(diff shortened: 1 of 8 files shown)


kotlin/src/main/java/feign/kotlin/CoroutineFeign.java
```diff
@@ -181,178 +149,51 @@ public <T> T target(Target<T> target, C context) {
...
+          .retryer(retryer)
+          .logger(logger)
+          .encoder(encoder)
+          .queryMapEncoder(queryMapEncoder)
+          .options(options)
+          .requestInterceptors(requestInterceptors)
+          .responseInterceptor(responseInterceptor)
+          .invocationHandlerFactory(invocationHandlerFactory)
+          .defaultContextSupplier((AsyncContextSupplier<Object>) defaultContextSupplier)
+          .methodInfoResolver(KotlinMethodInfo::createInstance)
```
velo, line 170: shouldn't this be extracted to an attribute on the builder and let people override if needed?

Also, that is necessary to allow coroutine `methodInfoResolver` to be enriched

velo (changes requested):
Outstanding job.

Just a small change to add `methodInfoResolver` attribute to Corotine builder.  Same as AsyncFeign builder

kotlin/src/main/java/feign/kotlin/CoroutineFeign.java
wplong11: Do you mean to allow the CoroutineBuilder to set the methodInfoResolver, and also provide an option to enrich the methodInfoResolver using Capability?

wplong11:
This PR will be closed, as PR #1755 containing this PR is merged.

kotlin/src/main/java/feign/kotlin/CoroutineFeign.java
wplong11: I created a new PR #1762

velo: Yes! Perfect

PR #1755, opened 2022-09-18
Refactor async feign

wplong11:
Contains #1754 PR commits

## TODO

- [ ] Modify AsyncFeign.asyncBuilder to extend Feign.Builder or Modify Feign to support the features of AsyncFeign
    - It is intended to be used when [customizing](https://docs.spring.io/spring-cloud-openfeign/docs/current/reference/html/#spring-cloud-feign-circuitbreaker) Feign.Builder in Spring Cloud OpenFeign.
    - Please feedback on Feign supporting the features of AsyncFeign. Are there any concerns regarding the library experience?
    - I think it's okay to work after merged, because CoroutineFeign is an experimental feature right now.

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -45,9 +45,16 @@ import java.util.function.Supplier;
  */
 @Experimental
 public abstract class AsyncFeign<C> {
-
-  public static <C> AsyncBuilder<C> asyncBuilder() {
+  public static <C> AsyncBuilder<C> builder() {
     return new AsyncBuilder<>();
+  }
+
+  /**
+   * @deprecated use {@link #builder()} instead.
+   */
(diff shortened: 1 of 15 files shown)


core/src/main/java/feign/AsyncFeign.java
```diff
@@ -44,10 +45,17 @@
  * be done (for example, creating and submitting a task to an {@link ExecutorService}).
  */
 @Experimental
-public abstract class AsyncFeign<C> extends Feign {
```
velo, line 47: Loved this.

I was hating myself for having `asyncBuilder` but couldn't take the time to fix it, amazing job, thanks

velo (approved):
I will merge as is, cause I really liked this change.

There is a chance it will break binary for people out there, but, it's an experimental API.

I think I will bump feign version as well to make clear it's breaking something

PR #1756, opened 2022-09-18
Extract asynchronous method handler

wplong11:
Contains #1755 PR commits

### DIFF
--- a/core/src/main/java/feign/AsynchronousMethodHandler.java
+++ b/core/src/main/java/feign/AsynchronousMethodHandler.java
@@ -13,9 +13,6 @@
  */
 package feign;
 
-import static feign.ExceptionPropagationPolicy.UNWRAP;
-import static feign.FeignException.errorExecuting;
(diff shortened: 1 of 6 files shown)


core/src/main/java/feign/AsynchronousMethodHandler.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+import static feign.ExceptionPropagationPolicy.UNWRAP;
+import static feign.FeignException.errorExecuting;
+import static feign.Util.checkNotNull;
+
+final class AsynchronousMethodHandler implements MethodHandler {
```
wplong11, line 30: In this PR, The AsynchronousMethodHandler is just copy and paste of SynchronousMethodHandler

PR #1757, opened 2022-09-18
Add retry feature to async feign

### LINKED ISSUE #1758: Add retry feature to async feign
- related: #361
- pre-work
  - #1754
    - Eliminates the possibility of breaking changes following AsyncFeign's refactoring in advance
  - #1755
  - #1756

wplong11:
Resolves #1758
Contains #1756 PR commits


velo (approved):

velo (approved):
@kdavisk6 do you want to take a look too?

I think this will break binary compatibility, so, major release?

wplong11:
@velo What classes are you concerned about breaking binary compatibility?(I know a little about Java binary compatibility) If you're concerned about breaking binary compatibility for the Async feature, isn't it ok because it's experimental?

velo:
> @velo , isn't it ok because it's experimental?

Yeah, it's ok. But, there will be complaints:/

wplong11:
@velo What changes break AsyncFeign's Binary Compatability? If the change from abstract to final is a problem, how about removing the final keyword? After that, how about adding the final keyword as a separate PR?

Issue #1734, opened 2022-08-27
Add support for Micrometer's new Observation API


jonatan-ivanov:
Micrometer introduces a new set of APIs in its next minor version (`1.10.0`).
From the high level perspective this API is somewhat similar to a `Timer` in the sense that it measures elapsed time and you can start/stop your measurements.

The advantage of making measurements using the Observation API over the `Timer` is that you can attach arbitrary handlers (listeners) who can react to events during your observations, e.g.: an observation was started/stopped and error happened, etc. This makes possible not only recording metrics (delegating to a Timer) but also creating Spans (distributed tracing), attaching logs (audit, trace, perf, etc.) or doing any arbitrary thing that you want.

Basically you instrument your code once and you can get multiple things out of this instrumentation, you don't need multiple instrumentations.

We have a migration guide that can give more information and also [examples](https://github.com/micrometer-metrics/micrometer/wiki/Migrating-to-new-1.10.0-Observation-API#processes-yellow-boxes) for migrating from a Timer-based instrumentation to the Observation API.

What do you think?

OlgaMaciaszek:
@velo @kdavisk6 We are making some major changes when it comes to tracing and observability for Spring 6. In order to maintain functionality, we would like to propose the change described above. We could provide the contribution. Please let us know what you think.

kdavisk6:
Apologies, I think this is a very good idea.  Feel free to submit any proposals.

OlgaMaciaszek:
Thanks :) . We'll do.

marcingrzejszczak:
I've submitted a draft PR that we can discuss - https://github.com/OpenFeign/feign/pull/1760

marcingrzejszczak:
Yes!

jonatan-ivanov:
Fixed by https://github.com/OpenFeign/feign/pull/1760

PR #1760, opened 2022-09-19
Micrometer Observations

marcingrzejszczak:
# Introduction

Since I was the one responsible for the code in Spring Cloud Sleuth responsible for integration with Feign, this time I've decided to suggest a little change in Feign so that it will allow easier code instrumentation.

In order to achieve proper tracing support, what we need to do is to 

- start a span before the request is made
- mutate the request headers (via `RequestTemplate`) before the request is made
- make the call
- parse the response to update span information
- close the span

With Micrometer Observation the span information is hidden since tracing is an opt-in feature. So what we need to do is the following

- create an observation before the request is made
- pass the mutable request to a `Observation.Context` object
- start the observation (here we are beginning to measure time / we create spans)
- make the call
- pass the response to the `Observation.Context` object
- end the observation

## Suggested changes

To achieve that in the least intrusive way (that seems to work - of course I will need to test it more thoroughly if you accept the idea behind the PR) I'm suggesting to add the following bits.

- A `ClientInterceptor` to work with mutable headers and easily pass objects before and after the HTTP request / response got procesed
- A way where a `Capability` would allow to mutate the `BaseBuilder` itself so that we can automatically add e.g. `ClientInterceptor`s when a capability is used (now we can only wrap existing objects).

### ClientInterceptor

a `ClientInterceptor` interface. Why not use the `RequestInterceptor` and `ResponseInterceptor`s ? Because `RI` requires an **immutable** `Request` and we can't mutate it's headers :shrug:. With `ClientInterceptor` you have access before `Client` is called (so you have still access to `RequestTemplate`) AND you can also fill a holder (I used a simple `Map` for that) that you can then read after the method execution. That way you don't need to play around with `ThreadLocal`s or anything like that to pass state between methods.

### BaseBuilder and Capabilities

There's already a `MicrometerCapability` that sets things up on a `FeignBuilder` and that's great. But what if we wanted to pre-set the builder with a `ClientInterceptor`? That's not possible (at least I don't know how to do it) - that's why this small change would allow to enrich the builder itself not only the builder's components.

## Things to consider

- Currently when using the capability and observations you will get both, old and new metrics. Most likely we should opt out of part of old metrics (potentially timers & samples) if `ObservationRegistry` is used

Related issue https://github.com/OpenFeign/feign/issues/1734


micrometer/pom.xml
```diff
@@ -27,6 +27,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
+    <micrometer.version>1.10.0-SNAPSHOT</micrometer.version>
```
velo, line 30: That is going to be a problem... will only be able to merge after 1.10 is released

velo: Is 1.10 really needed?

micrometer/pom.xml
```diff
@@ -72,4 +85,22 @@
     </plugins>
   </build>
 
+  <repositories>
```
velo, line 88: yeah... this can't be

core/src/main/java/feign/ClientInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Zero or One {@code ClientInterceptor} may be configured for purposes such as tracing - mutate
+ * headers, execute the call, parse the response and close any previously created objects.
+ */
+public interface ClientInterceptor {
+
+  ClientInterceptor DEFAULT = new ClientInterceptor() {
+    @Override
+    public void beforeExecute(ClientInvocationContext context) {}
```
velo, line 24: I really don't like `before`/`after` interceptors, exactly cause they can't share information and then the API need a mutable context to address it.

If you look at other `Interceptors` feign has, it's an `around` interceptor.  That gives the implementing code more flexibility to share data w/o the burden of having a mutable holder.

Another weak argument I have to use an around style is that java did that for servlet filters, so, can't be that bad =)

If you really want to expose a before/after style API, I guess we can have `ClientInterceptor` with the `around` method and then a `ClientInterceptorAdapter` with before/after methods.

core/src/main/java/feign/ClientInvocationContext.java
```diff
@@ -0,0 +1,43 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package feign;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ClientInvocationContext {
```
velo, line 19: This is something I have been thinking about for other interceptors.... having a `InvocationContext` that gives you access to all possible moving parts, instead of injecting each as parameters.  Really nice.

core/src/main/java/feign/ClientInterceptor.java
velo: Just to be clear, I have no problems with the context parameter, my problem is with the mutable holder.

I know feign has a lot of mutable states, and if I could I would eliminate them all, but, legacy can be hard.

micrometer/pom.xml
jonatan-ivanov: I guess this is temporary to be able to compile and run tests.
1.10 is needed for this since the new Observation API is in that release.

micrometer/pom.xml
jonatan-ivanov: I guess this temporary so that the PR can depend on Micrometer 1.10 (not yet GA).

micrometer/pom.xml
marcingrzejszczak: Yeah this PR would be merge once Micrometer 1.10.0 is out. Then there would be no SNAPSHOT dependencies.

micrometer/pom.xml
marcingrzejszczak: Yeah this PR would be merge once Micrometer 1.10.0 is out. Then there would be no SNAPSHOT dependencies.

core/src/main/java/feign/ClientInterceptor.java
marcingrzejszczak: Absolutely that's not a problem to do it with an `around` mechanism. The only concern I have is that with `before` and `after` you can have a list of these. If you have `around` you can't really have a list of those (unless I'm missing sth)

marcingrzejszczak: Actually I did manage to rewrite it to use around. Check the latest commit.

marcingrzejszczak:
In general if you like how things are looking like with this PR I can continue with ensuring that we don't double instrument etc. I just wanted to know if the direction in which this PR goes is acceptable

kdavisk6:
> In general if you like how things are looking like with this PR I can continue with ensuring that we don't double instrument etc. I just wanted to know if the direction in which this PR goes is acceptable

I'm pleased with the progress.  You have my 👍🏻

core/src/main/java/feign/Capability.java
```diff
@@ -137,4 +141,8 @@ default <C> AsyncContextSupplier<C> enrich(AsyncContextSupplier<C> asyncContextS
   default MethodInfoResolver enrich(MethodInfoResolver methodInfoResolver) {
     return methodInfoResolver;
   }
+
+  default <B extends BaseBuilder<B>> BaseBuilder<B> enrich(BaseBuilder<B> baseBuilder) {
```
velo, line 145: this doesn't make sense... the whole point of `Capability` is to change builders.... having a method that let you change the builder only indicates that the `Capability` API is incomplete.

this needs to be removed and whatever field needs customization needs to be added to the API

micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
```diff
@@ -24,18 +25,35 @@
...
+  public MicrometerCapability(MeterRegistry meterRegistry,
+      ObservationRegistry observationRegistry) {
     this.meterRegistry = meterRegistry;
+    this.observationRegistry = observationRegistry;
+  }
+
+  @Override
+  public <B extends BaseBuilder<B>> BaseBuilder<B> enrich(BaseBuilder<B> baseBuilder) {
+    if (!observationRegistry.isNoop()) {
+      baseBuilder.clientInterceptor(new ObservedClientInterceptor(observationRegistry));
```
velo, line 54: no good. use `enrich(ClientInterceptor)` instead.

velo (changes requested):

micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
marcingrzejszczak: That doesn't make sense cause I don't want to enrich an existing `ClientInterceptor` - I want to want to add a ClientInterceptor whenever this capability is being used. Regardless, I've applied the suggested change and needed to change the tests accordingly (I needed to ensure that there is a `ClientInterceptor.DEFAULT` that is being enriched.

marcingrzejszczak: I'm reverting that change cause it makes no sense. We would be overriding the provided client interceptor. In this case I don't think that capabilities are the way to go. Users will need to manually add a `ClientInterceptor` on a `Feign.Builder` and that's it.

marcingrzejszczak:
>     Currently when using the capability and observations you will get both, old and new metrics. Most likely we should opt out of part of old metrics (potentially timers & samples) if ObservationRegistry is used

That shouldn't really be a problem cause we're just instrumenting a single place (the exact point of making a request) so worst case scenario only that would be doubled. Of course only if the MicrometerCapability is turned on AND once adds the observed client interceptor.

I guess if this PR is the way to go we need to wait for Micrometer GA ?

velo:
Do you mind if I split this PR and make ONE just for the new interceptor?

marcingrzejszczak:
Of course, be my guest

marcingrzejszczak:
Hey @velo , what do you think of the plan?

marcingrzejszczak:
Hey @velo I see that you've released 12.0 of feign without any of these changes (this is the only PR that has been left in this repo). 

I understand that you're not planning to release a version that includes the `ClientInterceptor` nor the micrometer changes? It's been a month and a half since my PR was opened and it's been almost a month without any reply here. We're going GA with micrometer on 7/11 and there's Spring Cloud OpenFeign in the line and we need to plan our work somehow so it would be great if you could answer what your plans are.

OlgaMaciaszek:
@velo @kdavisk6 could you take a look at [Marcin's proposal](https://github.com/OpenFeign/feign/pull/1760#issuecomment-1301837389)?

velo:
>  (this is the only PR that has been left in this repo).

Yes, until micrometer is released this can't be merged.
Also, here are conflicts that need to be address before merging.

I pretty much do feign releases as needed, so can do another release later on the week

marcingrzejszczak:
I've updated the PR to remove conflicts + I extended the feature to work with Async builder too.

OlgaMaciaszek:
@velo could you please tak a look after the changes?

marcingrzejszczak:
@velo if you merge this and make a release we can then upgrade our version in Spring Cloud OpenFeign. What's the schedule of Feign so that we know how to plan things in the Spring Cloud train?

velo:
> @velo if you merge this and make a release we can then upgrade our version in Spring Cloud OpenFeign. What's the schedule of Feign so that we know how to plan things in the Spring Cloud train?

@marcingrzejszczak I refactored your change, to eliminate the need for another interceptor... we already have two o them, and introducing a third one didn't felt right just to intercept the client.

https://github.com/OpenFeign/feign/pull/1828

lemme know if that cover your needs.  I will still try to reduce my change scope, as 600 lines still looks a lot

marcingrzejszczak:
I've reviewed the PR and that looks like it should cover all needs indeed!

marcingrzejszczak:
I think we're ready to merge this, I've applied all the changes

velo (approved):

marcingrzejszczak:
Thanks for doing the merge! When are you planning to make the release?

velo:
@marcingrzejszczak @OlgaMaciaszek 
here you go:
https://github.com/OpenFeign/feign/releases/tag/12.1


BTW, would you guys consider having the spring contract as part of feign and have spring just concern with the dependency inject and the more springy side of business?

OlgaMaciaszek:
Thanks a lot, @velo . Interesting idea. We'll have our team meeting in Dec. Will be able to get back to you on it then.

PR #1766, opened 2022-09-22
Fix Truncated JSON literals in Headers Values

Zighome24:
Submitted as a resolution of #1464 
Compared to previous attempts to solve the issue, this PR utilizes truncated versions of the Public Domain org.json classes to perform JSON validation locally in an efficient manner. This addresses the need to validate JSON as well as the possibility of multiple values existing in the array of headers, not all of which may be JSON literals.


velo:
Pulling a library to core is not an option for the modular architecture of feign.... This should go to a module

velo (changes requested):
Sorry, feign core must be lightweight and not drag libraries, specially as source code

Issue #1606, opened 2022-04-07
Is it possible to customize the length of the body in FeignException?


jiahui-qin:
feign in master branch

I found that `FeignException` enforces the length of body not to exceed 400, use below two param:

```
    private static final int MAX_BODY_BYTES_LENGTH = 400;
    private static final int MAX_BODY_CHARS_LENGTH = 200;
```

Is there a way I can customize the length of these two variables to get more error messages?

Thank you!

prashanth057:
I need this to. Any way to bump up the priority on this?

velo:
Hi @prashanth057 , feign is mainly run by volunteers.  Either you need to get a volunteer interested to do it for you, or you file a PR for this and I will happily review it.

ddeath:
This can be closed as the proposal was merged right?

PR #1767, opened 2022-09-23
Issue #1606 Customize the length of the body in FeignException builder

coungard:


core/src/main/java/feign/FeignException.java
```diff
@@ -168,6 +168,10 @@ static FeignException errorReading(Request request, Response response, IOExcepti
   }
 
   public static FeignException errorStatus(String methodKey, Response response) {
+    return errorStatus(methodKey, response, null, null);
+  }
+  public static FeignException errorStatus(String methodKey, Response response,
+                                           Integer maxBodyBytesLength, Integer maxBodyCharsLength) {
```
velo, line 174: This method is never invoked by feign.

So what would be the point of the change.

I'm sure I'm missing something

velo (changes requested):

core/src/main/java/feign/FeignException.java
coungard: @velo it's that issue https://github.com/OpenFeign/feign/issues/1606, need for me, as a feign client.

With these parameters, my exception body will be full information in error decoder handler.

velo: @coungard what I'm not understanding is how.

That new method is never invoked by feign... how is that going to change anything?

coungard: @velo the result can be clearly seen on the test -> FeignExceptionTest::lengthOfBodyExceptionTest (143)

coungard: @velo Sorry, but isn't that enough? The code is called by Feign by default (null, null), but for us it will be overridden parameters:
изображение

velo: ow, I see now, my bad

velo (approved):

PR #1772, opened 2022-09-30
Exchange "Produces" and "Consumes"

chengda:
The annotation "ResponseBody" means that the return value of the controller method is a JSON, so it should be right to call the method "handleProducesAnnotation" . 
The annotation "RequestBody" means that the parameter of the controller method is a JSON, so it should be right to call the method "handleConsumesAnnotation" .


velo (changes requested):
Can we get some testing for this?!

chengda:
> Can we get some testing for this?!

Yes, you can test it . In my project, i met a problem about this , so i debugged it  and modify some codes like that, then i solved the problem.

chengda:
> Can we get some testing for this?!

One of my controller is like this:
`
@PostMapping(value="/test", produces="plain/text")
public String test(){
 ...
}
`
Before the modification，the request header 'Accept' that feign generated is 'application/json'; After the modification, it became 'plain/text'.

velo (changes requested):
Please include a unit test covering the change

chengda:
> Please include a unit test covering the change
Ok, it's done, please check it.

velo (approved):

Issue #1019, opened 2019-07-25
null should not be resolved in URITemplate


ddoubleday:
In versions of Feign before 10.x, if the URI template looked like http://base/action/{arg1}/{arg2}, and null was passed as the value to arg1, the URI 

http://base/action/{arg1}/somevalue  

would be tried, and rejected with IllegalArgumentException. That probably wasn't the right thing to do, but NOW, the null is actually resolved, and the situation is actually worse, resulting in the URI 

http://base/action//someValue 

resulting in, if you are using Spring Boot 2, a "not normalized URI exception" because they are doing security scans on URIs and disallowing consecutive slashes.

Even worse, if a value is passed to arg1 and null is passed to arg 2 and it is resolved, it can map to the entirely wrong controller method as:

http://base/action/someValue/

if these are GET operations, that might get the entire list associated with someValue, instead of the intended one value "someValue2".

Shouldn't feign refuse to resolve PATH variables if the supplied argument is null, and just throw some kind of NotNullable exception?

kdavisk6:
@ddoubleday 

Feign does it's best to adhere to [RFC 6570 - Uri Templates](https://tools.ietf.org/html/rfc6570) which states:

> **2.3.  Variables**
> An expression MAY reference variables that are unknown to the
   template processor or whose value is set to a special "undefined"
   value, such as undef or null.  Such undefined variables are given
   special treatment by the expansion process (Section 3.2.1).
>A variable value that is a string of length zero is not considered
   undefined; it has the defined value of an empty string.

>**3.2.1.  Variable Expansion**
>   A variable that is undefined (Section 2.3) has no value and is
   ignored by the expansion process.  If all of the variables in an
   expression are undefined, then the expression's expansion is the
   empty string.

Given this, Feign treats `null` values as `undef` values and will ignore them in the process.  There is no expectation that this treatment changes based on where the expression is defined.

With that said, the specification does have support for your use case via [Path Segment Expansion](https://tools.ietf.org/html/rfc6570#section-3.2.6), which is consider a Level 3 expansion.  Feign currently supports Level 1 or simple expansion at this time.  

With all of that, the behavior you are seeing is expected.

wilkystorm:
I second the concerns raised by @ddoubleday and am facing the same issues that he has enumerated above. Is it in the Feign roadmap to support these use cases and in effect correct this unintended behavior via support for level 3 path segment expansion or otherwise?

kdavisk6:
[...]

PR #1776, opened 2022-10-02
Implement a Regex based Check on Expressions

### LINKED ISSUE #1464: Truncated JSON header value with @Headers

JKomoroski:
There are a few things that result in Json headers getting truncated. Firstly, headers appear to get processed more than one time during the [RequestTemplate#resolve](https://github.com/OpenFeign/feign/blob/ca2f2003824a75f9ae1a2e95a3582075d978e39f/core/src/main/java/feign/RequestTemplate.java#L167) method. I reviewed that code and could not come up with simple way to address the issue without possibly breaking or making major changes to the external Request Template contract.

This fixes #1464

My solution was to simply validate that the string being passed to the expression code was actually an expression.

I don't know about code formatting or style on this. I'd like to see it back ported to the 11.x release if possible.

See my issue [comment](https://github.com/OpenFeign/feign/issues/1464#issuecomment-1264662983) for more context.

### DIFF
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -15,7 +15,6 @@ package feign.template;
 
 import feign.Param.Expander;
 import feign.Util;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


JKomoroski:
@velo You've reviewed some PRs that were seeking to fix this issue in the past. Are you the right person to review this?
- https://github.com/OpenFeign/feign/pull/1638
- https://github.com/OpenFeign/feign/pull/1766

@kdavisk6 You committed the change that introduced the regression. You may be well equipped to offer insight.

JKomoroski:
Rebased after a dependabot bump

kdavisk6 (commented):
While I don't have specific feedback on this change, I'd rather try to understand why the templates are being processed more than once, instead of attempting to make an educated guess at usage of the template.

JKomoroski:
> While I don't have specific feedback on this change, I'd rather try to understand why the templates are being processed more than once, instead of attempting to make an educated guess at usage of the template.


It basically comes down to the request storing a map of header templates and exposing that as a public api.

The code that resolves those templates by processing the header expressions turns the resolved espression strings back into templates to return them to the map, which triggers another pass through the code that checks for literals vs expressions.

Since inputs aren't really sanitized it processes the JSON as an expression.

An alternative would be to refactor header template with a processed flag or something, but I didn't want to modify any public interfaces -- that seemed far more invasive and less likely to be accepted.

kdavisk6:
I ended up creating an alternative that does modify `HeaderTemplate`, but doesn't expose any new methods on `RequestTemplate` https://github.com/OpenFeign/feign/pull/1781

JKomoroski:
> We are pretty open to these types of changes, you should go with your gut on these and we'll work with you to get them merged.

Formally defining what a legal Feign expression is, seemed like the place to start. It aligns with the goals on the readme as well.

If your PR is lower risk and fixes the issue that's great. Let's get it merged and released asap. I'd like to understand the concerns with using a regex to define the bounds of an expression as well. 

I use feign everyday and I'd like to get context to contribute more.

velo (approved):

kdavisk6:
> I'd like to understand the concerns with using a regex to define the bounds of an expression as well.

I don't have any concerns with using a regex here, however; by your own analysis, it doesn't correct the root cause, when appending already resolved headers, they are treated like expressions again.

For this minor nuance, I'd consider this PR an enhancement over a bug fix for #1464.  

Regardless, it's a good change, and we'll take it.

JKomoroski:
It sounds like the immediate problem https://github.com/OpenFeign/feign/issues/1464 (JSON headers being truncated) is fixed in https://github.com/OpenFeign/feign/pull/178. 

Are you maintainer folks are on board with tightening down this expression validation, I'd be happy to take another pass at this.

There is quite a bit of the expression processing code that could be improved by it. And I think we could prep the code base to implement the rest of the modified RFC for Version 12.

Thoughts?

kdavisk6:
I'll approve this and then feel free to look at #1019 for more information.

Onarki:
Hi, I don't know if this is the right place to ask, just wondering when this fix is going to be released ?
Thanks in advance

xiaohundun:
Could you please let me know if this fix has been released?

PR #1778, opened 2022-10-03
Fixes missing Content-Length header when body is empty

### LINKED ISSUE #1229: Default Client does not include Content-Length header for POST with Empty Body
We are using Feign 10.7.4 to execute a POST request with an empty body. In this case Feign does not add a Content-Length header to the request. According to the HTTP specification this is fine but some proxies, in our case Azure API Management, do not accept POST request with no Content-Length header. Would it be possible to add a `Content-Length: 0` header to PATCH, POST and PUT requests with an empty body?

Further context:
- https://stackoverflow.com/a/4198969
- https://github.com/request/request/issues/89
- https://github.com/urllib3/urllib3/issues/555
- https://github.com/gatling/gatling/issues/3648

c00ler:
The `DefaultClient` was modified to add the missing `Content-Length` header to `POST` and `PUT` requests. `PATCH` request is not supported by the `HttpUrlConnection`. Since `Content-Length` is the restricted header, I set the empty body, and its length is calculated by the `HttpUrlConnection`.

Fixes #1229

### DIFF
--- a/java11/pom.xml
+++ b/java11/pom.xml
@@ -52,16 +52,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.assertj</groupId>
-      <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>io.github.openfeign</groupId>
       <artifactId>feign-core</artifactId>
       <version>${project.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -517,13 +517,9 @@
         </dependencies>
       </plugin>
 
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -517,13 +517,9 @@
         </dependencies>
       </plugin>
 
-      <!-- Ensures checksums are added to published jars -->
       <plugin>
         <artifactId>maven-install-plugin</artifactId>
         <version>${maven-install-plugin.version}</version>
-        <configuration>
-          <createChecksum>true</createChecksum>
```
c00ler, line 525: This property was removed in https://issues.apache.org/jira/browse/MINSTALL-143

java11/pom.xml
```diff
@@ -51,16 +51,6 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
```
c00ler, line 54: Duplicated dependencies

kdavisk6 (approved):

Issue #1773, opened 2022-09-30
Illegal HTTP status code when IOException thrown


Regyl:
When external service is unavailable, or just not respond, SynchronousMethodHandler throws RetyableException with HTTP status code -1. The reference from master branch below.
```java
try {
      response = client.execute(request, options);
      // ensure the request is set. TODO: remove in Feign 12
      response = response.toBuilder()
          .request(request)
          .requestTemplate(template)
          .build();
    } catch (IOException e) {
      if (logLevel != Logger.Level.NONE) {
        logger.logIOException(metadata.configKey(), logLevel, e, elapsedTime(start));
      }
      throw errorExecuting(request, e);
    }
```
and the status:
```java 
static FeignException errorExecuting(Request request, IOException cause) {
    return new RetryableException(
        -1,
        format("%s executing %s %s", cause.getMessage(), request.httpMethod(), request.url()),
        request.httpMethod(),
        cause,
        null, request);
 }
```

**The problem:** 
It's _really_ hard to find because we even can't obtain response body due to unsupported numeric data type (negative).  In our case, we seen status code 65535 in Google Chrome (I assume Google Chrome use an unsigned short int datatype for http status code), what makes debug much more difficult. Tested on Postman, Windows PowerShell, Google Chrome v.101, cURL.

**The solution:**
So I'm offering to throw by default 500 (or maybe 0, if you don't want to throw an existing http status code) and provide an ability to customize this parameter. I'd be glad to fix it by myself.

gupta-rajkumar:
#1775

gupta-rajkumar:
Hi raised a PR for it can you please review? #1779

Regyl:
Can I ask you to see my PR too? https://github.com/OpenFeign/feign/pull/1780

gupta-rajkumar:
#1779

kdavisk6:
I've responded in the PR, so I'll echo that here:

> FeignException is not a stand-in for a proper Http Status driven exception. While we do provide Http Status-like exceptions for convenience, our intent is that a FeignException be processed by an ErrorDecoder and converted into a different exception type more suitable for the context it's being used in.

PR #1779, opened 2022-10-03
Fix Illegal HTTP status code thrown

gupta-rajkumar:
Added a new constructor with hard-coded status = 500.

### DIFF
--- a/core/src/main/java/feign/RetryableException.java
+++ b/core/src/main/java/feign/RetryableException.java
@@ -48,6 +48,16 @@ public class RetryableException extends FeignException {
   }
 
   /**
+   * @param retryAfter usually corresponds to the {@link feign.Util#RETRY_AFTER} header.
+   */
+  public RetryableException(String message, HttpMethod httpMethod, Throwable cause,
+      Date retryAfter, Request request) {
+    super(500, message, request, cause);
+    this.httpMethod = httpMethod;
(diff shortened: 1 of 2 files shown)


gupta-rajkumar:
#1773

core/src/main/java/feign/RetryableException.java
```diff
@@ -47,6 +47,16 @@ public RetryableException(int status, String message, HttpMethod httpMethod, Dat
...
   }
 
+  /**
+   * @param retryAfter usually corresponds to the {@link feign.Util#RETRY_AFTER} header.
+   */
+  public RetryableException(String message, HttpMethod httpMethod, Throwable cause,
+      Date retryAfter, Request request) {
+    super(500, message, request, cause);
+    this.httpMethod = httpMethod;
+    this.retryAfter = retryAfter != null ? retryAfter.getTime() : null;
```
kdavisk6, line 57: This exception is meant to portray a situation where the `Retry-After` header is present.  Given that, it's status should be taken from the use case.  Forcing it to 500 is incorrect in this use case.

kdavisk6 (changes requested):
`FeignException` is not a stand-in for a proper Http Status driven exception.  While we do provide Http Status-like exceptions for convenience, our intent is that a `FeignException` be processed by an `ErrorDecoder` and converted into a different exception type more suitable for the context it's being used in.

PR #1780, opened 2022-10-04
Fix illegal HTTP status code

### LINKED ISSUE #1773: Illegal HTTP status code when IOException thrown

Regyl:
Changes:
- Add connectExceptionStatus to the Options class which contains configurations for each FeignClient
- Switch from wrong status code to the connectExceptionStatus above (by default 500)
- Add tests

Fixes #1773

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Version 12.0
+
+* Ability to customize http status code when service unavailable by regyl (#1773)
+
 ### Version 11.9
 
 * `OkHttpClient` now implements `AsyncClient`
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -261,9 +261,11 @@ public class FeignException extends RuntimeException {
     }
   }
 
-  static FeignException errorExecuting(Request request, IOException cause) {
+  static FeignException errorExecuting(Request request,
+                                       IOException cause,
+                                       int connectExceptionStatus) {
     return new RetryableException(
-        -1,
+        connectExceptionStatus,
         format("%s executing %s %s", cause.getMessage(), request.httpMethod(), request.url()),
         request.httpMethod(),
         cause,
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -264,6 +264,7 @@ public final class Request implements Serializable {
     private final long readTimeout;
     private final TimeUnit readTimeoutUnit;
     private final boolean followRedirects;
+    private final int connectExceptionStatus;
 
     /**
      * Creates a new Options instance.
@@ -278,7 +279,7 @@ public final class Request implements Serializable {
     public Options(int connectTimeoutMillis, int readTimeoutMillis, boolean followRedirects) {
       this(connectTimeoutMillis, TimeUnit.MILLISECONDS,
           readTimeoutMillis, TimeUnit.MILLISECONDS,
-          followRedirects);
+          followRedirects, 500);
     }
 
     /**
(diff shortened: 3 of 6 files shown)


velo (changes requested):
Sorry, I don't think it's reasonable to assign an http response code when there was no http connection... 

Why is that necessary?

Regyl:
We have an exception handler where we parse response status code from feign exception like below:
```ava
   @ExceptionHandler(FeignException.class)
    public ErrorResponse handleFeignClientException(FeignException feignException, HttpServletResponse response) {
        response.setStatus(feignException.status());
    }
```
But when feign client throws exception with status code -1, we can't read the entire response. That makes debugging much more difficult (faced this problem on production).
You're right, there is no http connection. So how about send 0 as a status code?
In this case we won't send the correct http response status (not a part of RFC 2068) and simplify debugging for users.

velo:
> So how about send 0 as a status code?

Wouldn't 0 also be a invalid case too and cause the same issue? In any case, I would prefer to avoid breaking compatibility, as people may be using -1 already.

I think the correct answer on example above is for you to translate that -1 into 500.... or 503.

I don't think feign can say what is the appropriate code for everyone out there.

Regyl:
The general message that almost all tools whose interacting with HTTP couldn't read response if status code is _negative_.
I agree, this can broke someone's mechanism. To avoid crunches like checking status, can we add an ability to customize this value (e.g. from application.properties for Spring Boot) and set by default -1 if not specified another?

kdavisk6:
> The general promise that almost all tools whose interacting through HTTP couldn't read response if status code is _negative_. I agree, this can broke someone's mechanism. To avoid crunches like checking status, can we add an ability to customize this value (e.g. from application.properties for Spring Boot) and set by default -1 if not specified another?

>
>```java
> @ExceptionHandler(FeignException.class)
>  public ErrorResponse handleFeignClientException(FeignException feignException, HttpServletResponse response) {
>     response.setStatus(feignException.status());
>}
>```

This doesn't really fit with how Feign works.  Feign is not an HTTP client, so it doesn't need to adhere to HTTP standards.  It's our intention that you use an `ErrorDecoder` to produce your own exceptions from Feign's processing loop.

Regyl:
In this case ErrorDecoder didn't invoke. If I right checking response status on negative values the only way for me. This is correct behavior and you won't gonna change it?

velo:
> In this case ErrorDecoder didn't invoke. If I right checking response status on negative values the only way for me. This is correct behavior and you won't gonna change it?

well, in this case, there was no HTTP connection at all.... that is why feign is reporting a "-1 I don't know what is going on"

Regyl:
Have no questions anymore. Thank you for patience and explanation so much!

PR #1785, opened 2022-10-09
Remove the unused code from the MethodInfo

wplong11:
The MethodInfo.configKey field is not more used since PR #1757 removing AsyncInvocation

### DIFF
--- a/core/src/main/java/feign/MethodInfo.java
+++ b/core/src/main/java/feign/MethodInfo.java
@@ -20,19 +20,15 @@ import java.util.concurrent.CompletableFuture;
 
 @Experimental
 public class MethodInfo {
-  private final String configKey;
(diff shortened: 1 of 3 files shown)


kdavisk6 (changes requested):
`configKey` is required for use with `HystrixFeign`.  This is not unused code.

wplong11:
@kdavisk6 

> configKey is required for use with HystrixFeign. This is not unused code.

No problem  :)
MethodInfo.configKey is not used.
HystrixFeign uses `Feign.configKey`, not `MethodInfo.configKey`.
I deleted configKey from MethodInfo, and HystrixFeign does not reference MethodInfo

kdavisk6 (approved):

PR #1789, opened 2022-10-10
Refactor AsyncFeign

wplong11:
With the integration of AsyncFeign and Feign in mind,
I refactored the AsyncFeign logic in a way that reduces it as much as possible.


wplong11:
@velo I modified the commit and force pushed it, but Circle CI doesn't seem to be newly triggered

velo (approved):

PR #1795, opened 2022-10-12
Fix false negative test case for AsyncFeign Retry feature

wplong11:
False negative test case is added in PR #1757


wplong11:
@velo CircleCI does not work correctly

velo (approved):

PR #1799, opened 2022-10-21
Fix async doc to use asyncBuilder instead of builder

gabfssilva:
I think someone missed it. I lost a few hours trying to understand why it wasn't working anymore. One tip for the future: maybe somehow throw a UOE when someone tries to use `builder` + `AsyncFeign`? Don't know if it's possible tho.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1077,7 +1077,7 @@ interface GitHub {
 
 public class MyApp {
   public static void main(String... args) {
-    GitHub github = AsyncFeign.builder()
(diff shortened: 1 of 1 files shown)


wplong11:
`AsyncFeign.asyncBuild<C>` is deprecated in PR #1755. But it's not released now, so it would be nice if the README.md directs you to use `AsyncFeign.asyncBuild<C>`.

wplong11:
@gabfssilva PR #1755 was released as feign 12

PR #1800, opened 2022-10-22
Add Unit Test For Json Literals Being Passed as Header Values

JKomoroski:
Adding a unit test to cover the case described in issue 1305.

That issue is fixed. This test *should* prevent future regressions.


JKomoroski:
Rebased and Fixed Formatting

velo (approved):

Issue #1810, opened 2022-10-30
Got an error when I use feign-core to POST multipart/mixed request


wind13:
I am try to send batch request with feign-core, but got some 400 request error.

I have debug the code and found the problem is coming from feign-core set a `Content-Length` header. The code is `/feign/RequestTemplate.java` of feign-core 11.9.1 version.

I think the problem is because the multipart/mixed POST request should not send the `Content-Length` header.

I send the same request with Postman, it works success. And I could reproduce the same error with Postman when I add the `Content-Length` header.

So I guess there should have a condition code to check if the `Content-Type` is `multipart/mixed; boundary=batch_xxx`, then it should not add `Content-Length` header.

Thank you, with you fix this ASAP, waiting for use it on our project to send batch request.

wind13:
I found the solution of this problem, it is because the end-of-line problem. I am using \n in my code, but it is calculate with \r\n in the feign-core RequestTemplate.java.

So it is no need to fix then. Sorry, I will close this issue then.

PR #1811, opened 2022-10-30
Fix multipart/mixed POST 400 because adding `Content_Length` (#1810)

wind13:
…eader (#1810)


wind13:
I found the solution of this problem, it is because the end-of-line problem.  I am using `\n` in my code, but it is calculate with `\r\n` in the `feign-core` `RequestTemplate.java`.

So it is no need to fix then. Sorry, I will close this issue then.

PR #1819, opened 2022-10-31
Allow usage of GET with body in java11 Module

f-cramer:
Although discouraged there are unfortunately some web services that require a body to be sent with GET requests. Java 11's integrated http client allows this under the hood, but feigns java11 module currently does not. This PR adds support for bodys with any request method.

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -190,20 +190,7 @@ public class Http2Client implements Client, AsyncClient<Object> {
       requestBuilder.headers(asString(headers));
     }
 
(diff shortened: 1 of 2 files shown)


velo (commented):
Can we also get a test updated to match this change?

f-cramer:
I added two short tests that use the same API as the other tests an show that a Request Body transmitted via GET or DELETE is also present in the response

velo:
> I added two short tests that use the same API as the other tests an show that a Request Body transmitted via GET or DELETE is also present in the response

Thanks for that, seems build is not passing, recommend running locally to check

f-cramer:
Formatting... sorry for that

velo (approved):

PR #1828, opened 2022-11-09
Micrometer observations

velo:
Alternative micrometer observation using capability


micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
```diff
@@ -0,0 +1,69 @@
...
+        FeignObservationDocumentation.HttpClientTags.STATUS
+            .withValue(getStatusValue(context.getResponse())));
+  }
+
+  String getStatusValue(@Nullable Response response) {
+    return response != null ? String.valueOf(response.status()) : "CLIENT_ERROR";
+  }
+
+  String getMethodString(@Nullable Request request) {
+    return request.httpMethod().name();
```
marcingrzejszczak, line 66: If a `request` is `@Nullable` you might get NPE here (although I doubt that it would be the case in Feign)

micrometer/src/main/java/feign/micrometer/FeignObservationConvention.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.micrometer.core.instrument.binder.httpcomponents.DefaultApacheHttpClientObservationConvention;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationConvention;
+
+/**
+ * {@link ObservationConvention} for Feign.
+ *
+ * @since 1.10.0
+ * @see DefaultApacheHttpClientObservationConvention
```
marcingrzejszczak, line 24: Why this `@see` ? Also `@since 1.10.0` seems like a copy from micrometer ;)

micrometer/src/main/java/feign/micrometer/FeignContext.java
```diff
@@ -0,0 +1,34 @@
...
+import feign.Request;
+import feign.Response;
+import io.micrometer.observation.transport.RequestReplySenderContext;
+import io.micrometer.observation.transport.SenderContext;
+
+/**
+ * A {@link SenderContext} for Feign.
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.10.0
```
marcingrzejszczak, line 25: `@since 13.0` I guess?

micrometer/src/main/java/feign/micrometer/FeignObservationDocumentation.java
```diff
@@ -0,0 +1,81 @@
...
+
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationConvention;
+import io.micrometer.observation.docs.ObservationDocumentation;
+
+/**
+ * {@link ObservationDocumentation} for Feign.
+ * 
+ * @since 1.10.0
```
marcingrzejszczak, line 24: `@since 13.0` ?

micrometer/src/main/java/feign/micrometer/FeignObservationDocumentation.java
```diff
@@ -0,0 +1,81 @@
...
+      return HttpClientTags.values();
+    }
+  };
+
+  enum HttpClientTags implements KeyName {
+
+    STATUS {
+      @Override
+      public String asString() {
+        return "status";
```
marcingrzejszczak, line 45: I've made a mistake in the original PR, let's go with the semantic conventions from OTel https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

- `method` -> `http.method`
- `status` -> `http.status_code`
- `target.port` -> `net.peer.port`
- `target.host` -> `net.peer.name`
- `uri` -> `http.url`

I don't see `scheme` in the conventions so we can go with `http.scheme`

micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
```diff
@@ -0,0 +1,69 @@
...
+
+import feign.Request;
+import feign.Response;
+import io.micrometer.common.KeyValues;
+import io.micrometer.common.lang.Nullable;
+
+/**
+ * Default implementation of {@link FeignObservationConvention}.
+ *
+ * @since 1.10.0
```
marcingrzejszczak, line 24: `@since 13.0` ?

marcingrzejszczak (commented):
I've added small remarks but in general your suggestion looks much better than mine!

wplong11 (approved):

micrometer/src/main/java/feign/micrometer/FeignObservationDocumentation.java
velo: What if you merge my chances into your PR and then you can do the required changes?

marcingrzejszczak: No problem

micrometer/src/main/java/feign/micrometer/FeignObservationConvention.java
velo: I'm not sure either, you wrote that =)

micrometer/src/main/java/feign/micrometer/FeignContext.java
velo: FWIW, Feign next release going to be `12.1`

kdavisk6 (approved):

marcingrzejszczak:
Merged this PR to this one https://github.com/OpenFeign/feign/pull/1760

PR #1829, opened 2022-11-09
Add support for multiple ResponseInterceptors

velo:
In general, I avoid breaking feign API contracts like the plague, but I couldn't keep the old contract and have multiple response interceptors.

Take a look if this change makes any sense, and lemme know if we can include this, probably on feign 13

FWIW, this is heavily inspired on how spring graphql does interceptors chaining

### Summary by CodeRabbit

- New Feature: Introduced the ability to chain multiple response interceptors in Feign. This allows for more flexible and powerful handling of responses.
- Refactor: Replaced single `responseInterceptor` with a list of `responseInterceptors` across various classes, allowing for multiple interceptors.
- New Feature: Added `enrich(ResponseInterceptor.Chain chain)` method in `Capability` interface to enhance the functionality of response interceptor chains.
- Test: Updated tests in `BaseBuilderTest` to accommodate the new response interceptor chaining feature.
- Refactor: Renamed `InvocationContext` class to `ResponseInterceptor.Context` for better clarity and understanding of its role in the interception process.


wplong11 (approved):
Good

core/src/main/java/feign/BaseBuilder.java
```diff
@@ -269,5 +281,19 @@ List<Field> getFieldsToEnrich() {
         .collect(Collectors.toList());
   }
 
+  protected ResponseInterceptor.Chain executionChain() {
```
kdavisk6, line 284: can we consider a more descriptive name?  How about `interceptors` or `responseInterceptorChain?`

core/src/main/java/feign/ResponseInterceptor.java
```diff
@@ -13,26 +13,101 @@
...
+  Object intercept(Context context, Chain chain) throws IOException;
+
+  /**
+   * Return a new {@link ResponseInterceptor} that invokes the current interceptor first and then
+   * the one that is passed in.
+   * 
+   * @param nextInterceptor the interceptor to delegate to after the current
+   * @return a new interceptor that chains the two
+   */
+  default ResponseInterceptor andThen(ResponseInterceptor nextInterceptor) {
```
kdavisk6, line 46: This feels an awful log like `Function.andThen` or `Consumer.andThen`.  Do you think there's a way we could use those instead?

kdavisk6 (commented):
Love the idea, feel like we may be re-inventing the wheel though.  Any thoughts to using `Function.andThen` or `Consumer.andThen` instead?

core/src/main/java/feign/BaseBuilder.java
```diff
@@ -197,11 +197,23 @@ public B requestInterceptors(Iterable<RequestInterceptor> requestInterceptors) {
...
+    for (ResponseInterceptor responseInterceptor : responseInterceptors) {
+      this.responseInterceptors.add(responseInterceptor);
+    }
+    return thisB;
+  }
+
   /**
    * Adds a single response interceptor to the builder.
    */
   public B responseInterceptor(ResponseInterceptor responseInterceptor) {
```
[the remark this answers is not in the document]
velo: But it's not deprecated, it just changed from you can have only one, to you can have as many as you want, like requestInterceptor

PR #1841, opened 2022-11-16
Allow using SOAPEncoder constructor taking builder

### LINKED ISSUE #1839: Modify SOAP message and change SOAP protocol
I need to call a SOAP service which uses SOAP 1.2 but I also need to change the `SOAPMessage` to set a SOAP header.

At the Moment this can't be done because to set the SOAP protocol I have to use the Builder and to set SOAP headers I have to extend the class and overwrite the `modifySOAPMessage` method.

I see multiple options to fix this:
a) Change the visibility of the constructor which takes a Builder from private to protected
b) Add an protected all args constructor
c) Create an setter for the SOAP protocol

In my opinion option a) is the best. With it I could call super like this:
```java
public MySOAPEncoder(JAXBContextFactory jaxbContextFactory) {
        super(new SOAPEncoder.Builder()
                .withSOAPProtocol("SOAP 1.2 Protocol")
                .withJAXBContextFactory(jaxbContextFactory));
    }
```

Nicklas2751:
Change the SOAPEncoder constructor which takes a builder from private to protected

Fixes #1839

### DIFF
--- a/soap/src/main/java/feign/soap/SOAPEncoder.java
+++ b/soap/src/main/java/feign/soap/SOAPEncoder.java
@@ -93,7 +93,7 @@ public class SOAPEncoder implements Encoder {
   private final JAXBContextFactory jaxbContextFactory;
   private final String soapProtocol;
 
-  private SOAPEncoder(Builder builder) {
+  protected SOAPEncoder(Builder builder) {
     this.jaxbContextFactory = builder.jaxbContextFactory;
     this.writeXmlDeclaration = builder.writeXmlDeclaration;
     this.charsetEncoding = builder.charsetEncoding;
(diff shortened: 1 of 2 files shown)


soap/src/main/java/feign/soap/SOAPEncoder.java
```diff
@@ -93,7 +93,7 @@ public class SOAPEncoder implements Encoder {
   private final JAXBContextFactory jaxbContextFactory;
   private final String soapProtocol;
 
-  private SOAPEncoder(Builder builder) {
+  protected SOAPEncoder(Builder builder) {
```
velo, line 96: May be make it public instead?

velo (approved):

Issue #1869, opened 2022-12-03
`Flux` type reponse shoud be corresponding to `List`


CoderYellow:
the example [here](https://github.com/OpenFeign/feign/blob/d507ca3b1d19e7dd630e92e059af0ecdc75388ac/reactive/README.md?plain=1#L17) is not actually working with `JacksonDecoder`, got these errors:
```shell
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `feign.reactive.GitHubReactor$Contributor` from Array value (token `JsonToken.START_ARRAY`)
```
To support the json array response, I think we should wrap the response type to `List `for `Flux`, so that the decoder can decode it appropriately, some thing like [this](https://github.com/OpenFeign/feign/compare/master...CoderYellow:feign:feature/feign-reactor?expand=1)

CoderYellow:
is it ok for pr

velo:
Seems Jackson don't allow you to deserialize flux. Feign just delegates the job Jackson

CoderYellow:
The key is, a json array response should be mapped to `Flux`, single object response should be mapped to `Mono`. So does rxjava, should use `Flowable` and `Single` respectively. We should do [this](https://github.com/OpenFeign/feign/blob/d507ca3b1d19e7dd630e92e059af0ecdc75388ac/reactive/README.md?plain=1#L108) for user transparently. That's what other framework like retrofit do. @kdavisk6, do you agree?

PR #1871, opened 2022-12-05
Correct the corresponding response type of Flux

CoderYellow:
Fix issue #1869

### DIFF
--- a/reactive/pom.xml
+++ b/reactive/pom.xml
@@ -31,6 +31,7 @@
     <reactor.version>3.4.24</reactor.version>
     <reactive.streams.version>1.0.4</reactive.streams.version>
     <reactivex.version>2.2.21</reactivex.version>
+    <commons-lang.version>3.12.0</commons-lang.version>
   </properties>
 
   <dependencies>
@@ -93,6 +94,12 @@
       <artifactId>mockwebserver</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>${commons-lang.version}</version>
+    </dependency>
+
   </dependencies>
 
 </project>
--- a/reactive/src/main/java/feign/reactive/ReactiveDelegatingContract.java
+++ b/reactive/src/main/java/feign/reactive/ReactiveDelegatingContract.java
@@ -20,7 +20,9 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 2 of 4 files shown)


reactive/pom.xml
```diff
@@ -93,6 +94,12 @@
       <artifactId>mockwebserver</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
velo, line 97: Can we avoid the extra dependency?

reactive/src/main/java/feign/reactive/ReactiveDelegatingContract.java
```diff
@@ -55,7 +57,12 @@ public List<MethodMetadata> parseAndValidateMetadata(Class<?> targetType) {
           throw new IllegalArgumentException(
               "Streams are not supported when using Reactive Wrappers");
         }
-        metadata.returnType(actualTypes[0]);
+        if (isReactiveOf(type, Flux.class)) {
+          ParameterizedType wrappedType = TypeUtils.parameterize(List.class, actual);
```
velo, line 61: I'm not sure swapping the type from under the user is safe, I think the decoder should know how to deal with it.

Imagine we are talking about `LocalDateTime`.  Feign won't change it to `java.util.Date` so decoders do their trick and later change the type back to `LocalDateTime`.

I don't think this change is safe for feign

velo (changes requested):

reactive/src/main/java/feign/reactive/ReactiveDelegatingContract.java
CoderYellow: > I'm not sure swapping the type from under the user is safe, I think the decoder should know how to deal with it.
> 
> Imagine we are talking about `LocalDateTime`. Feign won't change it to `java.util.Date` so decoders do their trick and later change the type back to `LocalDateTime`.
> 
> I don't think this change is safe for feign

But we've already done this here, we change the type `Publisher` type to the normal one. It's a delegation class, it's meant to do this. So why not distinguish the `Publisher` type further. The `Mono` and `Flux` definitely have different semantics, as I mentioned [here](https://github.com/OpenFeign/feign/issues/1869#issuecomment-1336363851).

The best solution is to implement the real reactive decoder, but currently we just need a wrapper.

reactive/pom.xml
CoderYellow: Could you tell me what your concern is, the apache common library is a widely used library? Can I just copy the related code here since it is under apache license too?

velo: Big dependency tree, conflicts with other libraries (which would lead to runtime errors)... When possible, we avoid extra dependencies on feign

PR #1881, opened 2022-12-15
support method option and add UT

TaoJing96:
Background:
- I want to add request timeout for each feign api, but none of them has param Options.

Change:
- Support method option and add UT
- Format code style

Usage:
1. You can see `socketTimeoutWithMethodOptionsTest` in OptionsTest.java
2. In spring Cloud

```
interface MyClient {
   public MyResponse getMyData();
}
```

```
@FeignClient(name = "my-service", url = "xxxx", configuration = MyConfig.class)
public interface MyFeignService extends MyClient {
}
```

```
@Configuration
public class MyConfig {

    @Bean
    @ConditionalOnMissingBean(name = "options")
    Request.Options options() {
        return new Request.Options(5000, 10000);
    }
}
```

```
@Component
public class MyServiceClient {
    @Resource
    private MyFeignService myFeignService;

    @Resource
    private Request.Options options;

    public MyResponse getMyData(Request.Options options) {
        options.setMethodOptions("getMyData", options); //you can also use aop.
        //write your code below
    }
}
```

## Summary by CodeRabbit

New Features:
- Added the ability to set and retrieve method-specific options in the `Options` class, providing more flexibility in handling different methods.
- Introduced a new utility function `getThreadIdentifier()` for retrieving the current thread's identifier.

Tests:
- Implemented two new tests to verify the correct behavior of method-specific options in both normal and timeout scenarios.

Refactor:
- Enhanced the `findOptions` method in the `RequestTemplate` class to handle method-specific options, improving the overall request handling process.


velo:
You spoke a lot of spring up there, I'm not sure if I got it all.

Lemme see if I got it correctly, what you want is to be able to change timeouts for individual methods dynamically.

So, you want to call `getMyData` sometimes with 500ms timeout and other times `getMyData` with 5s timeout.  Is that it? Why?

TaoJing96:
> You spoke a lot of spring up there, I'm not sure if I got it all.
> 
> Lemme see if I got it correctly, what you want is to be able to change timeouts for individual methods dynamically.
> 
> So, you want to call `getMyData` sometimes with 500ms timeout and other times `getMyData` with 5s timeout. Is that it? Why?

Yes, i want to change timeouts for individual methods dynamically and i don't need add param Options for each method.

velo:
> In case 1, i want to call `getMyData` with 500ms, in case 2, i i want to call `getMyData` with 5000ms, it makes sense.

If you have any concurrence, you might end up with case 1 executing the call with 5s timeout and case 2 with 500ms, the exact opposite of what you are after.

Could you describe me a bit more of what you are trying to accomplish?

TaoJing96:
> > In case 1, i want to call `getMyData` with 500ms, in case 2, i i want to call `getMyData` with 5000ms, it makes sense.
> 
> ~If you have any concurrence, you might end up with case 1 executing the call with 5s timeout and case 2 with 500ms, the exact opposite of what you are after.~ nvm, just noticed the thread name... still, looks so fragile and a memory leakage
> 
> Could you describe me a bit more of what you are trying to accomplish?

OK, let me show you the code.
I wrote the code last year.
```
interface MyClient {
   public MyResponse getMyData();
}
```
However, i want to add request timeout while calling `getMyData`, how can i accomplish it on the condition that i don't add param Options in `accomplish `?

TaoJing96:
> > In case 1, i want to call `getMyData` with 500ms, in case 2, i i want to call `getMyData` with 5000ms, it makes sense.
> 
> ~If you have any concurrence, you might end up with case 1 executing the call with 5s timeout and case 2 with 500ms, the exact opposite of what you are after.~ nvm, just noticed the thread name... still, looks so fragile and a memory leakage
>
Why it looks so fragile and a memory leakage? Could you give more infos?
> Could you describe me a bit more of what you are trying to accomplish?

core/src/main/java/feign/Request.java
```diff
@@ -303,6 +306,33 @@ public static class Options {
...
     private final boolean followRedirects;
+    private final Map<String, Map<String, Options>> threadToMethodOptions;
+
+    /**
+     * Get an Options by methodName
+     * 
+     * @param methodName it's your FeignInterface method name.
+     * @return method Options
+     */
+    public Options getMethodOptions(String methodName) {
```
velo, line 317: My spider sense is telling me this will cause headaches at some point.

Since I don't have a good suggestions for right now, just annotate and the new public methods with `@Experimental` and I will merge.  Just want to reserve the right to undo/redo in the future.

velo (changes requested):

core/src/main/java/feign/Request.java
TaoJing96: Ok, added.

velo: Missing `@Experimental` on this one =)

velo (approved):

PR #1883, opened 2022-12-15
set root cause as the exception_name in micrometer tags

### LINKED ISSUE #1882: Can we set the root cause of exception in metrics?
Setting the root cause(e.g. `SocketTimeoutException` ) as the tag value might be more helpful than the exception in feign layer(e.g. `RetryableException`)

https://github.com/OpenFeign/feign/blob/5b8cd4a1e7d1e3a0ab09002a84351369acd8ddd6/micrometer/src/main/java/feign/micrometer/FeignMetricTagResolver.java#L52

chao-chang-paypay:
resolves #1882

### DIFF
--- a/dropwizard-metrics4/src/main/java/feign/metrics4/BaseMeteredClient.java
+++ b/dropwizard-metrics4/src/main/java/feign/metrics4/BaseMeteredClient.java
@@ -18,6 +18,7 @@ import com.codahale.metrics.Timer;
 import feign.FeignException;
 import feign.RequestTemplate;
 import feign.Response;
+import feign.utils.ExceptionUtils;
 
 class BaseMeteredClient {
 
@@ -64,7 +65,7 @@ class BaseMeteredClient {
             MetricRegistry.name(
                 httpResponseCode(template),
                 "exception_name",
-                e.getClass().getSimpleName(),
+                ExceptionUtils.getRootCause(e).getClass().getSimpleName(),
                 "status_group",
                 e.status() / 100 + "xx",
                 "http_status",
@@ -81,7 +82,7 @@ class BaseMeteredClient {
             MetricRegistry.name(
                 httpResponseCode(template),
                 "exception_name",
-                e.getClass().getSimpleName(),
+                ExceptionUtils.getRootCause(e).getClass().getSimpleName(),
                 "uri",
                 template.methodMetadata().template().path()),
             metricSuppliers.meters())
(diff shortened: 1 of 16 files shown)


dropwizard-metrics5/src/main/java/feign/metrics5/MeteredDecoder.java
```diff
@@ -63,14 +64,14 @@ public Object decode(Response response, Type type)
     } catch (IOException | RuntimeException e) {
       metricRegistry.meter(
           metricName.metricName(template.methodMetadata(), template.feignTarget(), "error_count")
-              .tagged("exception_name", e.getClass().getSimpleName())
+              .tagged("exception_name", ExceptionUtils.getRootCause(e).getClass().getSimpleName())
```
velo, line 67: I'm thinking a bit more about this.... we are breaking the existing contract with this change.

Can we leave `exception_name` as was and add `root_cause_name` for this?

Cheers

chao-chang-paypay: yes, that makes more sense, made the same change in https://github.com/OpenFeign/feign/pull/1883/commits/23cb5ed45d71937dc61d5214861997f99cbe561b ,  can you take a look at it again?

dropwizard-metrics4/src/main/java/feign/metrics4/BaseMeteredClient.java
```diff
@@ -65,6 +66,8 @@ protected void recordFailure(RequestTemplate template, FeignException e) {
                 httpResponseCode(template),
                 "exception_name",
                 e.getClass().getSimpleName(),
+                "root_cause_name",
+                ExceptionUtils.getRootCause(e).getClass().getSimpleName(),
```
velo, line 70: `getCause` can return null, in that case would cause NPE

chao-chang-paypay: As long as e is not null, getRootCause won’t return null as we can see here
https://github.com/OpenFeign/feign/pull/1883/files#diff-3bfdf878b3ea38d5268a9d72389c1f54ebabf1d76ad9a308e1eed93d95ed0048R38

velo: Ow, awesome, great!

velo (approved):

PR #1900, opened 2022-12-30
fix(sec): upgrade org.springframework:spring-web to 

ns-cn:
### What happened？
There are 1 security vulnerabilities found in org.springframework:spring-web 4.3.30.RELEASE
- [CVE-2016-1000027](https://www.oscs1024.com/hd/CVE-2016-1000027)

### What did I do？
Upgrade org.springframework:spring-web from 4.3.30.RELEASE to  for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/spring4/pom.xml
+++ b/spring4/pom.xml
@@ -30,7 +30,7 @@
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
 
-    <spring.version>4.3.30.RELEASE</spring.version>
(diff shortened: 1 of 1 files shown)


kdavisk6:
We use other means to keep our dependencies updated.  Also, the library you modified is intended for Spring 4 and this PR will break that module.

PR #1902, opened 2022-12-31
Fix elapsedTime with space for easy sort -nrk 

qxo:


kdavisk6:
Changes without associated issues are discouraged.

Looking over this change, it doesn't seem to address any existing concerns other than to modify the output of the logger.  Given that you can provide your own Logger that you can customize to fit your needs, this pull request is not necessary.

PR #1926, opened 2023-02-04
Update Maven wrapper to pull latest Maven (3.8.7)

aalmiray:
Also, use mvnw on CI instead of default mvn provided. Apply -ntp flag to reduce noise in logs


velo (approved):
Great work, thanks for it

PR #1930, opened 2023-02-08
updates reactor-core dependency for the feign-reactive-wrappers

### LINKED ISSUE #1711: feign-reactive-wrappers:11.9.1 not compatible with reactor-core:3.5.0
```java
package feign.reactive;
public class ReactorFeign extends ReactiveFeign {
    public static class Builder extends ReactiveFeign.Builder {
        private Scheduler scheduler = Schedulers.elastic();    // there is no .elastic()  method
    }
}
```

nlsbchmnn:
removed deprecated usage of <code>Schedulers.elastic()</code> and using <code>Schedulers.boundedElastic()</code> instead

Fixes #1711

### DIFF
--- a/reactive/src/main/java/feign/reactive/ReactorFeign.java
+++ b/reactive/src/main/java/feign/reactive/ReactorFeign.java
@@ -25,12 +25,20 @@ import feign.Target;
 public class ReactorFeign extends ReactiveFeign {
 
   public static Builder builder() {
-    return new Builder();
+    return new Builder(Schedulers.boundedElastic());
+  }
+
+  public static Builder builder(Scheduler scheduler) {
+    return new Builder(scheduler);
   }
 
   public static class Builder extends ReactiveFeign.Builder {
 
-    private Scheduler scheduler = Schedulers.boundedElastic();
+    private final Scheduler scheduler;
+
+    Builder(Scheduler scheduler) {
+      this.scheduler = scheduler;
+    }
 
     @Override
     public Feign build() {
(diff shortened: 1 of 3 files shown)


reactive/src/main/java/feign/reactive/ReactorFeign.java
```diff
@@ -30,7 +30,7 @@ public static Builder builder() {
 
   public static class Builder extends ReactiveFeign.Builder {
 
-    private Scheduler scheduler = Schedulers.elastic();
```
velo, line 33: Problem is this breaks binary compatibility.... is so annoying when projects break compatibility on minor releases

velo: can we make `Scheduler scheduler` a parameter, and if never set, we default to `Schedulers.boundedElastic()`?

And then document how to use 3.4

nlsbchmnn: Just for my understanding: Why is it breaking binary compatibility?

It appears to me that the <code>scheduler</code> is already a parameter, which you can set with <code>public Builder scheduleOn(Scheduler scheduler)</code> and <code>Schedulers.elastic()</code> is the current default.

velo: `elastic` to `boundedElastic` breaks anyone using the old version of the library.

Tell you what, as a compromise, move the initialization to constructor and have one that allows user to pick how to create the scheduler and the other defaults to new code

```
  public static Builder builder() {
    return new Builder();
  }

  public static Builder builder(Scheduler scheduler) {
    return new Builder(scheduler);
  }

  public static class Builder extends ReactiveFeign.Builder {

    private Scheduler scheduler;

.. constructors...
```

velo (approved):

PR #1935, opened 2023-02-13
Add Optional BeanParam support for JAXRS2Contract

JKomoroski:
Implements a change to the JAXRS2Contract to optionally support `@BeanParam`  parameter aggregators. It only adds support for aggregated parameters already supported by the existing JAXRS2Contract (ie not `@Context` or `@Suspended`).

fixes #1264

PR #1942, opened 2023-02-17
Ignore Jakarta Suspended and Context, Support BeanParam

### LINKED ISSUE #1940: JakartaContract does not support JAX-RS 2.x Annotations
The `JakartaContract` is a copy of the `JAXRSContract` and does not have the added features of the `JAXRS2Contract`. As far as I know there is no reason for supporting the 1.x specification under Jakarta namespace and not accounting for the 2.x specification.

JKomoroski:
A small bit of house keeping to bring the Jakarta package in line with the Jaxrs2 package.

Includes the work from jaxrs2 bean param on pr https://github.com/OpenFeign/feign/pull/1935

Fixes #1940

### DIFF
--- a/jakarta/README.md
+++ b/jakarta/README.md
@@ -32,3 +32,5 @@ Links the value of the corresponding parameter to a query parameter.  When invok
 Links the value of the corresponding parameter to a header.
 #### `@FormParam`
 Links the value of the corresponding parameter to a key passed to `Encoder.Text<Map<String, Object>>.encode()`.
+#### `@BeanParm`
+Aggregates the above supported parameter annotations under a single value object.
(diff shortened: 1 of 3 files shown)


velo:
Any ideas on how to keep this 2 modules always in sync?

JKomoroski:
> Any ideas on how to keep this 2 modules always in sync?

@velo I don't think there's a great way to handle it as written without getting too fancy with a shade plugin.

Maybe a really generic base class could be extended by jaxrs1, jaxrs2, and jakarta implementations, but this would probably require a new module and the amount of code saved seems minimal to me. 

It would almost certainly solve the test dependency quirks we were seeing in jdk8 though

PR #1948, opened 2023-02-20
Add module-info support and run CI with Java 11

aalmiray:
Related to https://github.com/OpenFeign/feign/pull/1943

I can confirm that `./mvnw verify -Dmoditect.skip=true` works with Java 8. This additional flag is set when running tests.
The CNFE seen at #1943 no longer occurs when running the build with Java 8.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,27 +12,12 @@
 # the License.
 #
 
-orbs:
-  android: circleci/android@1.0.3
-
 # common executors
 executors:
   java:
-    parameters:
-      version:
-        description: 'jdk version to use'
-        default: '11.0'
-        type: string
     docker:
-      - image: cimg/openjdk:<<parameters.version>>
-  android:
-    parameters:
-      version:
-        description: 'jdk version to use'
-        default: '8.0'
-        type: string
-    docker:
-      - image: cimg/openjdk:<<parameters.version>>
+      - image: velo/toolchains-4-ci-builds
+
 # common commands
 commands:
   resolve-dependencies:
@@ -60,16 +45,7 @@ commands:
           name: 'Deploy Core Modules Sonatype'
           command: |
             ./mvnw -ntp -nsu -s .circleci/settings.xml -P release -pl -:feign-benchmark -DskipTests=true deploy
-  nexus-deploy-jdk11:
-    steps:
-      - run:
-          name: 'Build JDK 11 Release modules locally'
-          command: |
-            ./mvnw -ntp -B -nsu -s .circleci/settings.xml -P java11 -pl :feign-java11,:feign-jakarta -am -DskipTests=true install
-      - run:
-          name: 'Deploy JDK 11 Modules to Sonatype'
-          command: |
-            ./mvnw -ntp -B -nsu -s .circleci/settings.xml -P release,java11 -pl :feign-java11,:feign-jakarta -DskipTests=true deploy
+
 # our job defaults
 defaults: &defaults
   working_directory: ~/feign
@@ -98,14 +74,8 @@ version: 2.1
 
 jobs:
   test:
-    parameters:
-      jdk:
-        description: 'jdk version to use'
-        default: '11.0'
-        type: string
     executor:
       name: java
-      version: <<parameters.jdk>>
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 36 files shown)


velo:
Release/snapshot process happens on jdk8... only a couple of modules are released using jdk11

aalmiray:
If release/snapshot happens on jdk8 then moditect cannot be used and thus no module infos.
By setting `maven.compiler.source`, `maven.compiler.target`,  and `maven.compiler.release` (when running for sure on Java11+) the generated bytecode remains Java8 compatible even if the build uses a JDK that's not jdk8.

```
$ ./mvnw --version
Using maven at '/Users/aalmiray/dev/github/feign/mvnw' to run buildFile '/Users/aalmiray/dev/github/feign/pom.xml':
Apache Maven 3.8.7 (b89d5959fcde851dcb1c8946a785a163f14e1e29)
Maven home: /Users/aalmiray/.m2/wrapper/dists/apache-maven-3.8.7-bin/678cc9d4/apache-maven-3.8.7
Java version: 11.0.17, vendor: Azul Systems, Inc., runtime: /Users/aalmiray/.sdkman/candidates/java/11.0.17-zulu/zulu-11.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "11.6.5", arch: "x86_64", family: "mac"

$ jarviz bytecode show --file core/target/feign-core-12.2-SNAPSHOT.jar 
Unversioned classes. Bytecode version: 52 total: 156
Unversioned classes. Bytecode version: 53 total: 1

y$ jarviz module descriptor --file core/target/feign-core-12.2-SNAPSHOT.jar 
name: feign.core
version: 12.2-SNAPSHOT
open: false
automatic: false
exports:
  feign
  feign.auth
  feign.codec
  feign.optionals
  feign.querymap
  feign.stream
  feign.template
  feign.utils
requires:
  java.base mandated
  java.desktop
  java.logging
```

.circleci/config.yml
```diff
@@ -90,8 +90,11 @@ jobs:
           key: feign-dependencies-{{ checksum "pom.xml" }}
       - run:
           name: 'Test'
+          # Do not generale module-info when running tests
+          # because one of the current JDK targets is 8
+          # Moditect requires 11 as a minimum
           command: |
-            ./mvnw -ntp -B test
+            ./mvnw -ntp -B test -Dmoditect.skip=true
```
velo, line 97: Think we can revert this now, right?!

annotation-error-decoder/pom.xml
```diff
@@ -50,4 +50,13 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.moditect</groupId>
```
velo, line 57: Move to the top pom, and skip plugin on the one project that can't be excuted

velo (changes requested):
New toolchain build in place.... build passes just have a few improvements ideas.

annotation-error-decoder/pom.xml
aalmiray: OK. This adding the moditect plugin to the parent, enabled by default, skip on no modular projects, right?

pom.xml
```diff
@@ -66,6 +66,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
+    <moditect.skip>true</moditect.skip>
```
velo, line 69: Can we make default false and only declare skip = true on modules that need it?

So most submodules will be unchanged

velo (changes requested):

pom.xml
aalmiray: Be aware that the root must skip otherwise it will trigger an error, that's why the default is set to `true`.

.circleci/config.yml
velo: Please revert this, build now runs using java 17 and just compiler brings java 8 to the mix.

aalmiray: Will do.

pom.xml
```diff
@@ -460,6 +462,36 @@
...
+                    <addServiceUses>true</addServiceUses>
+                  </moduleInfo>
+                </module>
+                <jdepsExtraArgs>
+                  <arg>--multi-release=9</arg>
+                </jdepsExtraArgs>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
```
velo, line 494: can we move this whole section to line 533 (outside `pluginManagement`), I see no reason to keep it in here

velo (approved):

pom.xml
aalmiray: Yes, it can be safely moved.

velo: Yeah. I moved, push and merged... didn't realized the push got a conflict and didn't moved 🤦‍♂️

PR #1952, opened 2023-02-24
Attempt to simplify build process using maven toolchains

velo:


velo:
@aalmiray if this work, adding module information could become easier

kdavisk6 (approved):

PR #1999, opened 2023-03-29
add a way to pass param to RequestInterceptor

Branhub:
I want to pass param to the RequestInterceptor but found RequestTemplate only has the infos to build the http request.
I add a new annotation so that the method param it marked can be accessed through RequestTemplate in the RequestInterceptor.It can only mark one method param and can only be accessed as java.lang.Object for now.

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -321,6 +321,11 @@ public interface Contract {
             "HeaderMap annotation was present on multiple parameters.");
         data.headerMapIndex(paramIndex);
       });
+      super.registerParameterAnnotation(ExtraParam.class, (queryMap, data, paramIndex) -> {
+        checkState(data.extraParamIndex() == null,
+            "HeaderMap annotation was present on multiple parameters.");
+        data.extraParamIndex(paramIndex);
+      });
     }
 
     private static Map<String, Collection<String>> toMap(String[] input) {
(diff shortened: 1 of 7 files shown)


velo:
Hi @Branhub 

Thanks for you contribution.
Wonder what is your goal here. Would you like to read or set a parameter at `RequestInterceptor` level?

I would prefer to make this possible without adding a new annotation

Branhub:
> 

Thank you for your reply.
The goal is to read a parameter at RequestInterceptor. But since the RequestTemplate return an Object , there is no way to prevent the RequestInterceptor from modifing it now.
It can be achived by setting a holder class ThreadLocal field if no new thread is involved. But that means you need to set the holder class field before the call to the FeignClient mehtod every time.
Or It can also be achived by setting some field in the body or header so that the RequestInterceptor can get access to it but that will change the http request.

Branhub:
> Hi @Branhub
> 
> Thanks for you contribution. Wonder what is your goal here. Would you like to read or set a parameter at `RequestInterceptor` level?
> 
> I would prefer to make this possible without adding a new annotation

Thank you for your reply.
The goal is to read a parameter at RequestInterceptor. But since the RequestTemplate return an Object , there is no way to prevent the RequestInterceptor from modifing it now.
It can be achived by setting a holder class ThreadLocal field if no new thread is involved. But that means you need to set the holder class field before the call to the FeignClient mehtod every time.
Or It can also be achived by setting some field in the body or header so that the RequestInterceptor can get access to it but that will change the http request.

kdavisk6:
Request Interceptors are typically used to manage headers and other ancillary parts of the request pipeline, like adding tags, etc...  For manipulating the `RequestTemplate` itself, this sort of work is done in the `Encoder`, which can modify the `RequestTemplate` since it's part of the resolution stage.  Have you considered that approach?

PR #2005, opened 2023-04-04
Add a way to choose between class or package instantiation mode for JaxbContext

jeremybalan:
I suggest to add a way to choose between class or package instantiation for JaxbContext (class mode is still used by default).

To allow this, the builder of JaxbContextFactory has a new optional method "withJAXBContextInstantiationMode" to provide an enum for the choice between class and package mode. No call to this method = usage of class mode (backward compatible).

Benefits of a package way can be : 
- a better usage of the JAXBContext cache : one entry for all class of the package instead of one entry (one JaxbContext) per class
- a usage of the ObjectFactory class with all elements available to be loaded (using class mode needs an analyze to every dependents elements)
- this precedent point can also be a workaround to some problems with class loading mode (for example with Jaxb Moxy implementation, see https://github.com/eclipse-ee4j/eclipselink/issues/1856)

### DIFF
--- a/jaxb/src/main/java/feign/jaxb/JAXBContextCacheKey.java
+++ b/jaxb/src/main/java/feign/jaxb/JAXBContextCacheKey.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2012-2023 The Feign Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


jeremybalan:
Class mode (by default) instantiate a JAXBContext using JAXBContext.newInstance(clazz) as it is actually

And package mode, if chosen, instantiate using JAXBContext.newInstance(clazz.getPackage().getName(), clazz.getClassLoader())

velo (approved):

Issue #1918, opened 2023-01-25
feign-soap change from javax to jakarta  


ambikurup:
For spring boot 3 support, will feign-soap using javax be changed to jakarta. Currently using 12.1 version I'm getting 

javax.xml.bind.JAXBException Implementation of JAXB-API has not been found on module path or classpath

I added jaxb dependency but still getting same error.

sandalovsergey:
Have a same problem.. @velo can you please take a look?

I've found this one https://github.com/OpenFeign/feign/pull/1700 but it still use javax packages instead jakarta packages. F.e. see this: https://github.com/OpenFeign/feign/blob/12.1/jaxb/src/main/java/feign/jaxb/JAXBContextFactory.java#L21

It is from jaxb-api (javax.xml.bind:jaxb-api:2.3.1) but afaik shoud be from bind-api (jakarta.xml.bind:jakarta.xml-bind).

Did i make a mistake?

velo:
Probably a new module is required for Jakarta change

sandalovsergey:
@velo thanks for the answer!

Did I understood correctly that I can create a PR with new module(f.e feign-jaxb-jakarta), where instead of `javax.xml.bind`-packages, there will be `jakarta.xml.bind`-packages ?

velo:
If you can do that, that would be amazing.

Someone already did it for jaxrs, so seems reasonable

ambikurup:
can you please let me know when will be the new module published and ready to use?

EarthCitizen:
@kdavisk6  @velo  

As mentioned above by @virtual-machinist, this original problem of this issue is not solved by the release of module `jaxb-jakarta` as `soap` still uses the old APIs. There needs to be a `soap-jakarta` as laid out by @virtual-machinist .

kistlers:
@virtual-machinist @EarthCitizen 
I already implemented the feign-soap-jakarta. And it is already merged and released with feign 12.4!. See https://github.com/OpenFeign/feign/pull/2094

I just included the new module this week. In a project, where i replaced feign-soap by the new jakarta module and managed to replace old jaxb by jakarta dependencies.

Please go ahead and use it. I'd gladly hear feedback if there are any bugs or missing features.

PR #2014, opened 2023-04-14
added jaxb-jakarta module on the basis of jaxb module

FloLei:
Created a new jaxb-jakarta module based on discussion #1918. Code is only based on the existing jaxb module with the jakarta api packages.


steam0:
Any idea when this new module will be released?

PR #2028, opened 2023-04-27
Added support to add literals directly to the header

schaoukicoveo:
Added support to add literals directly to the header. 

In short, I exposed a public method that will provide a way to add literals directly to the header. Beforehand, a literal would never be evaluated/seen as one up until we try to create the `Template` [object.](https://github.com/OpenFeign/feign/blob/9b5d27a0e88474f5b419e9297285fb59e63a6c53/core/src/main/java/feign/template/Template.java#L218), it would be beneficial to have such a method call as it will prevent extra processing with trying to match a complex regex expression with something that wishes to be processed as literal and not an expression. This will provide a clearer and simpler way for end users to send literals, especially for json strings. 

Properly fixes [#1464](https://github.com/OpenFeign/feign/issues/1464)

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -15,12 +15,6 @@ package feign;
 
 import static feign.Util.CONTENT_LENGTH;
 import static feign.Util.checkNotNull;
-import feign.Request.HttpMethod;
-import feign.template.BodyTemplate;
-import feign.template.HeaderTemplate;
-import feign.template.QueryTemplate;
-import feign.template.UriTemplate;
-import feign.template.UriUtils;
 import java.io.Serializable;
 import java.net.URI;
 import java.nio.charset.Charset;
@@ -39,6 +33,12 @@ import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import feign.Request.HttpMethod;
+import feign.template.BodyTemplate;
+import feign.template.HeaderTemplate;
+import feign.template.QueryTemplate;
+import feign.template.UriTemplate;
+import feign.template.UriUtils;
 
 /**
  * Request Builder for an HTTP Target.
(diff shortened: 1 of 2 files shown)


core/src/test/java/feign/RequestTemplateTest.java
```diff
@@ -15,13 +15,10 @@
 
 import static feign.assertj.FeignAssertions.assertThat;
 import static java.util.Arrays.asList;
-import static org.assertj.core.api.Assertions.assertThat;
```
schaoukicoveo, line 18: Removal of unused import

core/src/test/java/feign/RequestTemplateTest.java
```diff
@@ -243,6 +242,18 @@ public void resolveTemplateWithHeaderWithNestedJson() {
...
+    String json = "{ \"foo\": \"bar\", \"hello\": \"world\"}";
+    RequestTemplate template = new RequestTemplate().method(HttpMethod.GET)
+        .headerLiteral("A-Header", "{aHeader}");
+
+    template = template.resolve(mapOf("aHeader", json));
+
+    assertThat(template)
+        .hasHeaders(entry("A-Header", Collections.singletonList("{aHeader}")));
+  }
+
```
schaoukicoveo, line 256: Makes sure that if a expression that is added as a literal will not be evaluated as an expression when trying to resolve.

core/src/main/java/feign/RequestTemplate.java
```diff
@@ -787,6 +803,21 @@ private RequestTemplate appendHeader(String name, Iterable<String> values, boole
...
+   * @return an Iterable of strings
+   */
+  private Iterable<String> validateHeaderValues(String name, Iterable<String> values) {
+    if (name == null || name.isEmpty()) {
+      throw new IllegalArgumentException("name is required.");
+    }
+
+    return values == null ? Collections.emptyList() : values;
+  }
+
```
schaoukicoveo, line 820: Provide a private method that will contain the necessary validation for headers

Issue #2040, opened 2023-05-10
Exit span doesn't include meaningful peer information when using micrometer as skywalking reporter


wapkch:
# Description
I tried to use Micrometer to report OpenFeign traces to Skywalking OAP server. maven dependencies are following:
```xml
<dependency>
      <groupId>org.apache.skywalking</groupId>
      <artifactId>apm-toolkit-micrometer-1.10</artifactId>
      <version>8.15.0</version>
   </dependency>

<dependency>
      <groupId>io.github.openfeign</groupId>
      <artifactId>feign-micrometer</artifactId>
      <version>12.1</version>
   </dependency>

<dependency>
      <groupId>org.springframework.cloud</groupId>
      <artifactId>spring-cloud-starter-openfeign</artifactId>
      <version>4.0.2</version>
   </dependency>
```

Feign Client definition:
```java
@FeignClient(name = "jsonplaceholder", url = "http://jsonplaceholder.typicode.com")
public interface JsonPlaceholderService {

    @GetMapping("/posts")
    List<Post> loadPosts();

}
```

ObservationRegistryCustomizer:
```java
@Bean
    public ObservationRegistryCustomizer<ObservationRegistry> observationRegistryCustomizer() {
        return registry -> {
            // Here we add a meter handler
            registry.observationConfig()
                .observationHandler(new ObservationHandler.FirstMatchingCompositeObservationHandler(
                    new SkywalkingMeterHandler(new SkywalkingMeterRegistry())
                ));
            // Here we add tracing handlers
            registry.observationConfig()
                .observationHandler(new ObservationHandler.FirstMatchingCompositeObservationHandler(
                    new SkywalkingSenderTracingHandler(), new SkywalkingReceiverTracingHandler(),
                    new SkywalkingDefaultTracingHandler()
                ));
        };
    }
```

When i start the application and invoke `JsonPlaceholderService` loadPosts method, the skywalking-api.log has errors:
```
ERROR 2023-05-10 17:14:34.584 http-nio-9000-exec-1 InstMethodsInter : class[class org.apache.skywalking.apm.toolkit.micrometer.observation.SkywalkingSenderTracingHandler] before method[onStart] intercept failure 
java.lang.IllegalStateException: Exit span doesn't include meaningful peer information.
	at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:169)
	at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:149)
[...]

PR #2065, opened 2023-05-26
Add RemoteService to FeignContext

wapkch:
As mentioned in #2040, i think RemoteService should be set to FeignContext, so components like Skywalking could retrieve the host and port as the peer instead of using templated URL path.

@marcingrzejszczak Please review this draft PR.

### DIFF
--- a/micrometer/src/main/java/feign/micrometer/FeignContext.java
+++ b/micrometer/src/main/java/feign/micrometer/FeignContext.java
@@ -13,10 +13,14 @@
  */
 package feign.micrometer;
 
-import feign.Request;
-import feign.Response;
+import io.micrometer.common.util.StringUtils;
 import io.micrometer.observation.transport.RequestReplySenderContext;
 import io.micrometer.observation.transport.SenderContext;
+
+import java.net.URL;
+
+import feign.Request;
+import feign.Response;
 
 /**
  * A {@link SenderContext} for Feign.
@@ -29,6 +33,15 @@ public class FeignContext extends RequestReplySenderContext<Request, Response> {
   public FeignContext(Request request) {
     super(Request::header);
     setCarrier(request);
+    setRemoteServiceName(request.requestTemplate().feignTarget().name());
+    try {
+      URL url = new URL(request.requestTemplate().feignTarget().url());
+      if (StringUtils.isNotBlank(url.getProtocol()) && StringUtils.isNotBlank(url.getHost())) {
+        setRemoteServiceAddress(url.getProtocol() + "://" + url.getHost() + ":" + url.getPort());
+      }
+    } catch (Exception ignored) {
+
+    }
   }
 
 }
--- a/micrometer/src/main/java/feign/micrometer/FeignContext.java
+++ b/micrometer/src/main/java/feign/micrometer/FeignContext.java
@@ -13,14 +13,12 @@
  */
 package feign.micrometer;
 
+import feign.Request;
+import feign.Response;
+import java.net.URL;
 import io.micrometer.common.util.StringUtils;
 import io.micrometer.observation.transport.RequestReplySenderContext;
 import io.micrometer.observation.transport.SenderContext;
-
-import java.net.URL;
-
-import feign.Request;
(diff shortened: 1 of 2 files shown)


velo (approved):

marcingrzejszczak:
Hey! How does it look like in Zipkin after your changes? Do you have a screenshot?

wapkch:
> Hey! How does it look like in Zipkin after your changes? Do you have a screenshot?

Sure, before the changes:
711685610291_ pic

After the changes:
701685610115_ pic

The FeignClient:
```java
@FeignClient(name = "jsonplaceholder", url = "jsonplaceholder.typicode.com")
public interface JsonPlaceholderFeignClient {

    @GetMapping("/posts")
    List<Post> loadPosts();

}

```

wapkch:
> What will happen if the other application (e.g. `jsonplaceholder`) is also instrumented and has a different name (e.g. `foo`)? The standard practice is that the service name gets automatically resolved as the server's local service name. Can you test this against an app (e.g. a Spring Boot 3 app) that has an MVC filter that is instrumented and show the screenshot?

It will be like this:
image
In this test, `jsonplaceholder` application's `spring.application.name` is `foo`, and the FeignClient's name is still `jsonplaceholder`

marcingrzejszczak:
So this is with the changes from this PR? Cause the second span still has `application` there as a service name?

wapkch:
> So this is with the changes from this PR? Cause the second span still has `application` there as a service name?

Yes, i confirmed, the second span just had a different service name compared with the case that the remote service did not report traces.

marcingrzejszczak:
I'm asking cause this works well if you're not using apache skywalking. WDYT about doing the setting of remote service name and address inside skywalking via the observation convention mechanism? Or the observation filter?

wapkch:
> I'm asking cause this works well if you're not using apache skywalking. WDYT about doing the setting of remote service name and address inside skywalking via the observation convention mechanism? Or the observation filter?

I considered this option, but i think the problem of doing the setting inside skywalking is that skywalking have to distinguish the different Sender implementations(Retrieve the remote service name and address from carrier then set to SenderContext), there are so many SenderContext and carrier implementations, it will be hard to maintain in skywalking. So i think doing the settings in seperated projects will be better.

marcingrzejszczak:
>  there are so many SenderContext and carrier implementations(e.g., Feign, Resttemplate, messaging...)

So you're saying that there will be the same problem with all of these implementations? I'm just curious if this is not a problem related to Skywalking as such. I don't recall having such problems with Zipkin / Tempo etc.

micrometer/src/main/java/feign/micrometer/FeignContext.java
```diff
@@ -29,6 +31,15 @@ public class FeignContext extends RequestReplySenderContext<Request, Response> {
   public FeignContext(Request request) {
     super(Request::header);
     setCarrier(request);
+    setRemoteServiceName(request.requestTemplate().feignTarget().name());
+    try {
+      URL url = new URL(request.requestTemplate().feignTarget().url());
+      if (StringUtils.isNotBlank(url.getProtocol()) && StringUtils.isNotBlank(url.getHost())) {
+        setRemoteServiceAddress(url.getProtocol() + "://" + url.getHost() + ":" + url.getPort());
```
velo, line 38: Wouldn't `setRemoteServiceAddress(url.toString())` do the trick?

micrometer/src/main/java/feign/micrometer/FeignContext.java
```diff
@@ -29,6 +31,15 @@ public class FeignContext extends RequestReplySenderContext<Request, Response> {
...
+      }
+    } catch (Exception ignored) {
```
velo, line 40: what exception is been ignored here?

probably should re-throw it, as will be impossible for people to understand why something is not working if we are eating the exception

velo (changes requested):
When ready, please move from draft

velo:
No answer to feedback and still on draft, will close, if needed please reopen

PR #2066, opened 2023-05-27
Re-format the overview mindmap

luciolucio:
Balances it out visually by splitting the first level in two sides, i.e., it will look like this:

mindmap-with-left-side


velo:
Outstanding, thanks for it

Issue #1479, opened 2021-08-10
Feign Soap Exceptions


alabotski:
JAXB by default silently ignores errors.  
Can you add this code to throw an exception if something goes wrong.
To SoapEncoder & SoapDecoder
```

 unmarshaller.setEventHandler(
    new ValidationEventHandler() {
        @Override
        public boolean handleEvent(ValidationEvent event ) {
            throw new RuntimeException(event.getMessage(),
                                       event.getLinkedException());
        }
});
```

virtual-machinist:
By `javax.xml.bind.ValidationEventHandler#handleEvent` contract you can simply return `false` and JAXB will throw an exception:
```
unmarshaller.setEventHandler(event -> false);
```
However I wouldn't add this sort of handling by default, as all of the written code already relies on the default JAXB behavior and instead provide means to customize `feign.jaxb.JAXBContextFactory.Builder`.

kdavisk6:
I agree with @virtual-machinist, this is something you'll have to provide to the `SOAP` encoder/decoder components.  Those components should accept an external `JAXBContext`.  If not, please open an enhancement issue/pr to do so.

virtual-machinist:
@kdavisk6 I'm afraid it's not that simple and we have to modify `feign.jaxb.JAXBContextFactory` even if we allow external `JAXBContext` for SOAP encoders and decoders.

AFAIK you can only set the `ValidationEventHandler` when `Marshaller` and `Unmarshaller` instances are created, i.e. at some point in `feign.jaxb.JAXBContextFactory#createUnmarshaller` or `feign.jaxb.JAXBContextFactory#createMarshaller`.

In other words something like
```
public Builder withUnmarshallerValidationEventHandler(ValidationEventHandler handler) {
 ...
}

public Builder withMarshallerValidationEventHandler(ValidationEventHandler handler) {
 ...
}
```
is needed + modifications to the `create` factory methods.

If this is a reasonable proposal, I can make a branch + PR. :smiley:

virtual-machinist:
@kdavisk6 I've created a PR (#2084) that hopefully addresses this issue.

PR #2084, opened 2023-06-14
Allow schema and validation event handler customization in JAXBContextFactory

virtual-machinist:
The default JAXB behavior is to silently ignore errors caused by input being incorrect, so eg. an `Integer` field will remain `null` for an input `foo`. This isn't always preferred, but to fix that a `ValidationEventHandler` implementation must be assigned to the created `Unmarshaller`. It may be as simple as `event -> false`.

For a more complex validation scenario a `Schema` may be needed.

However at the moment there is no way to set up either of those in `Marshaller` and `Unmarshaller` instances created in JAXB and SOAP encoders/decoders.

The PR addresses this problem by adding the customization options for `feign.jaxb.JAXBContextFactory` in `jaxb` and `jaxb-jakarta` modules. A customized `JAXBContextFactory` may then be used to address the original issue in #1479 .

### DIFF
--- a/jaxb-jakarta/src/main/java/feign/jaxb/JAXBContextFactory.java
+++ b/jaxb-jakarta/src/main/java/feign/jaxb/JAXBContextFactory.java
@@ -14,6 +14,7 @@
 package feign.jaxb;
 
 import jakarta.xml.bind.*;
+import javax.xml.validation.Schema;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


jaxb-jakarta/src/main/java/feign/jaxb/JAXBContextFactory.java
```diff
@@ -51,6 +78,10 @@ public Unmarshaller createUnmarshaller(Class<?> clazz) throws JAXBException {
   public Marshaller createMarshaller(Class<?> clazz) throws JAXBException {
     Marshaller marshaller = getContext(clazz).createMarshaller();
     setMarshallerProperties(marshaller);
+    if (marshallerEventHandler != null) {
```
velo, line 81: So, this is piling more and more functionality/code into feign to allow fine tuning...  wonder if it would be possible to turn this into an interface that allows user to do whatever they to `JAXBContext` and pass it to feign in a more transparent way..

virtual-machinist:
JAXBContextFactory is IMO a misnomer, as it's mainly used to create
Marshaller and Unmarshaller instances - things that actually do the
serialization and deserialization. The contexts are also created, but are
cached per class, and usually an application can have a single context for
all supported classes if there's no package/namespace collisions.

But JAXBContext once created isn't very customizable by itself and I don't
see a way to achieve the needed fine-tuning unless we do some subclassing
hackery and delegate to the real thing from Glassfish or your favorite JAXB
provider. This is IMO much more fragile than the solution proposed here.

That said we can actually create some sort of MarshallerCustomizer /
UnmarshallerCustomizer interface that does the needed post-processing. We
may even wrap the already implemented property setters as instances of such
customizers and execute them in sequential order like done for request
interceptors.

How's that sound?

velo:
>  How's that sound?


sounds great, would avoid all ifs inside feign, if you keen to do it, I would like to see it

velo (approved):

PR #1463, opened 2021-07-21
Allow QueryMapEncoder accept null value

quaff:
```java
	@GetMapping
	public Paged<User> list(int page, int size, @QueryMap User user);
```
Currently feign doesn't accept null `user`
```java
Paged<User> paged = userClient.list(1, 10, new User()); // query map will include primitive type default values which is unexpected
Paged<User> paged = userClient.list(1, 10, null); // will leave query map empty
```

PR #2086, opened 2023-06-16
Allow QueryMapEncoder accept null value

quaff:
For
```java
@GetMapping
public Paged<User> list(int page, int size, @QueryMap User user);
```
now we can use
```java
Paged<User> paged = userClient.list(1, 10, null);
```
instead of
```java
Paged<User> paged = userClient.list(1, 10, new User()); // query map will include primitive type default values which is unexpected
```


quaff:
@velo Please review this one since you close https://github.com/OpenFeign/feign/pull/1463.

velo (approved):

PR #2093, opened 2023-06-19
Adds convenient overloaded constructor for Request.Options

iwek7:
This proposes adding new convenient constructor for Request.Options that enables us to pass timeout as duration. In my apps timeout is very often passed around as duration, which is imo better than moving it around as magic number without explicit unit attached. This constructor overload simply feels like it should be there from user perspective.


velo (approved):
LGTM, just need to pass build

iwek7:
build is fixed now

PR #2094, opened 2023-06-19
feign-soap-jakarta module like feign-soap, but with feign-jaxb-jakarta instead of feign-jaxb

kistlers:
This allows using feign clients in conjunction with soap using the newer Jakarta packages instead of old Javax. This feature builds on top of https://github.com/OpenFeign/feign/pull/2014 to add the functionality.

The module is mostly a copy of the feign-soap module with old javax dependencies replaced by jakarta.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,7 @@
     <module>slf4j</module>
     <module>spring4</module>
     <module>soap</module>
+    <module>soap-jakarta</module>
     <module>reactive</module>
     <module>dropwizard-metrics4</module>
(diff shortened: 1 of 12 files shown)


velo (approved):
LGTM,

can you also added it to the features?

https://github.com/OpenFeign/feign/blob/master/src/docs/overview-mindmap.iuml#L21

kistlers:
@velo Done, I also added JAXB Jakarta to the list, as I realized it was missing.

PR #2103, opened 2023-06-22
Apache HC5 Client uses Request.Options to manage follow redirects

### LINKED ISSUE #2060: Enable/Disable followRedirects via Request.Options in ApacheHttp5Client
Hi I was using `ApacheHttp5Client` and there was no way to set redirectsEnabled `false` using `Request.Options` as by default it is set `true`, only `connectTimeout` and `responseTimeout` are set from Request.Options in method `configureTimeouts()` Line: 95. So it is not possible to disable redirections using feign configurations as below:

```javascript
feign:
    client:
        config:
            feignName:
                connectTimeout: 1000
                follow-redirects: false
``` 

It can be fixed by adding the following line:
```java
.setRedirectsEnabled(options.isFollowRedirects())
```

radio-rogal:
Fixes #2060


radio-rogal:
There is another PR #2104 from the author of issue.

PR #2113, opened 2023-07-03
Allow override of maxBodyBytesLength in ErrorDecoder

ddeath:
Allow override of maxBodyBytesLength, maxBodyCharsLength in ErrorDecoder

Currently if we use Default error decoder from this lib it will cut error messages so we cant really see them in the logs. This will allow us to configure max error message length.

Added ability to specify our own max values in Default error decoder.

Let me know what you think about this

### DIFF
--- a/core/src/main/java/feign/codec/ErrorDecoder.java
+++ b/core/src/main/java/feign/codec/ErrorDecoder.java
@@ -82,6 +82,10 @@ public interface ErrorDecoder {
    *         retryable, it should be wrapped, or a subtype of {@link RetryableException}
    */
   public Exception decode(String methodKey, Response response);
+  default Exception decode(String methodKey, Response response, Integer maxBodyBytesLength,
+      Integer maxBodyCharsLength) {
+    return decode(methodKey, response);
+  }
 
   public static class Default implements ErrorDecoder {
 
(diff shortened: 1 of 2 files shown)


radio-rogal:
IMHO:

1. ErrorDecoder is called by ResponseHandler that is used by *MethodHandler. Users usually do not call it manually. We do not call it in any Client's child.
2. There is the (default) `decode` method that have two parameters that it ignores. For me it is bad way.
3. Why do we need to pass maxBodyBytesLength and maxBodyCharsLength each time when we call the `decode` method? I guess these values are the same.

If the question _"How to set maxBodyBytesLength and maxBodyCharsLength"_ the solution is "use the constructor": we can add another constructor that gives these parameters and passes them to [`errorStatus(String methodKey, Response response, Integer maxBodyBytesLength, Integer maxBodyCharsLength)`.](https://github.com/OpenFeign/feign/blob/26dadc26c315dbb20fda825c4beda7655fefce0d/core/src/main/java/feign/FeignException.java#L174C51-L174C51)

ddeath:
> ~IMHO:~
> 
> ~1. ErrorDecoder is called by ResponseHandler that is used by *MethodHandler. Users usually do not call it manually. We do not call it in any Client's child. 2. There is the (default) `decode` method that have two parameters that it ignores. For me it is bad way. 3. Why do we need to pass maxBodyBytesLength and maxBodyCharsLength each time when we call the `decode` method? I guess these values are the same.~
> 
> ~If the question _"How to set maxBodyBytesLength and maxBodyCharsLength"_ the solution is "use the constructor": we can add another constructor that gives these parameters and passes them to [`errorStatus(String methodKey, Response response, Integer maxBodyBytesLength, Integer maxBodyCharsLength)`.](https://github.com/OpenFeign/feign/blob/26dadc26c315dbb20fda825c4beda7655fefce0d/core/src/main/java/feign/FeignException.java#L174C51-L174C51)~
> 
> As far as I see you have implemented it :)

Yeah I realized later that via constructor it is much simpler and does not require breaking change

velo (approved):

PR #2117, opened 2023-07-03
The builder clones itself before enrichment

### LINKED ISSUE #1961: MicrometerCapability is applied multiple times when using a Feign.Builder to build multiple Feign instances
Each time `Feign.Builder.build()` is used to build a new `Feign` instance, it calls `enrich` to apply the `Capability`s. The `MicrometerCapability` then wraps the `Client`, `Encoder`, `Decoder` etc. in `MeteredXxx` decorators. So with each time I build a new `Feign` from the same `Builder`, the decorator chain grows by one, so I get e.g a `MeteredDecoder` wrapped in a `MeteredDecoder` wrapped in a `MeteredDecoder` and so on. In our case, this caused a `StackOverflowError` after the application ran for a while.

Suggestion: Check if the `Decoder` is already an instance of `MeteredDecoder` and skip wrapping if it is. (Same for the other objects as well, of course.) But beware: This will not help if you have two capabilities that decorate these objects, because then one capability might only see the decorator from the other capability, not its own...

Other option: Change the code in `Feign.Builder` to apply the capabilities only once.

vitalijr2:
I have tried to resolve #1961: now when the builder makes a new *Feign it wraps own fields, on second call it wraps already wrapped fields. I replace the link to itself builder `thisB` and own fields by a clone and its fields.

### Summary by CodeRabbit

**Refactor:**
- Introduced a new method `internalBuild()` across various classes in the Feign library, improving code modularity and maintainability.
- Enhanced the `BaseBuilder` class with cloning capabilities to enrich builder objects, enhancing code consistency and logic.
- Updated test cases to reflect changes in the `BaseBuilder` class.

> Hopping through the code, making it neat 🐇,
> Refactoring here and there, isn't that sweet? 🍭
> With every change, we're on a roll,
> Making our codebase, beautifully whole. 🎉

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -186,30 +186,32 @@ public final class AsyncFeign<C> {
     }
 
     public AsyncFeign<C> build() {
-      super.enrich();
+      AsyncBuilder<C> enrichedBuilder = super.enrich();
 
       AsyncResponseHandler responseHandler =
           (AsyncResponseHandler) Capability.enrich(
               new AsyncResponseHandler(
-                  logLevel,
-                  logger,
-                  decoder,
-                  errorDecoder,
-                  dismiss404,
-                  closeAfterDecode, responseInterceptor),
+                  enrichedBuilder.logLevel,
+                  enrichedBuilder.logger,
+                  enrichedBuilder.decoder,
+                  enrichedBuilder.errorDecoder,
+                  enrichedBuilder.dismiss404,
+                  enrichedBuilder.closeAfterDecode, enrichedBuilder.responseInterceptor),
               AsyncResponseHandler.class,
-              capabilities);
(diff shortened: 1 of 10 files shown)


core/src/main/java/feign/AsyncFeign.java
```diff
@@ -186,30 +186,32 @@ public AsyncBuilder<C> invocationHandlerFactory(InvocationHandlerFactory invocat
     }
 
     public AsyncFeign<C> build() {
-      super.enrich();
+      AsyncBuilder<C> enrichedBuilder = super.enrich();
```
velo, line 189: My major concern is that in the future, someone will access a field directly, forget the enrichBuilder prefix

What is,  we move the builder body to a new method, and the we invoke the "internalBuild()"  in the enriched instance?

velo (approved):

vitalijr2:
Oh, AI rules world :)

PR #2136, opened 2023-07-25
Fix expression pattern compilation on android

iProdigy:
The ICU regex engine used by Android requires `}` in `EXPRESSION_PATTERN` to be escaped.

Fixes `java.lang.NoClassDefFoundError: feign.template.Expressions` on Android (reported by @Awakened-Redstone)


velo (commented):
Can we get a unit test around this? Just to prevent breaking again in the future

Issue #1721, opened 2022-08-09
The "Accept-Encoding" header is added twice when the " compression" option is enabled


pastrehe:
feign-bom.version 11.9.1
spring-boot 2.7.2

When `feign.compression.response.enabled` is set to `true` in configuration, feign adds "Accept-Encoding" header(s) when sending request. Currently, two header fields are added (`Accept-Encoding: gzip` and `Accept-Encoding: deflate`).
According to rfc7230 section 3.2.2
<cite>_A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception_</cite>
Expected behaivour is to have one header field with the name "Accept-Encoding" and value "gzip, deflate".
ATM, having 2 "Accept-Encoding" headers causing remote side to "flatten" them and pick only one value randomly ("gzip" or "deflate")

radio-rogal:
As far as I undertand you use **httpclient**. Could you confirm, please?

I had made a test project with Spring 2.7.2 and Spring Cloud's OpenFeing and met the same issue:
```
DEBUG i.g.r.t.Monobank - [Monobank#getRates] ---> GET https://api.monobank.ua/bank/currency HTTP/1.1
DEBUG i.g.r.t.Monobank - [Monobank#getRates] Accept-Encoding: gzip
DEBUG i.g.r.t.Monobank - [Monobank#getRates] Accept-Encoding: deflate
DEBUG i.g.r.t.Monobank - [Monobank#getRates] Content-Type: application/json
DEBUG i.g.r.t.Monobank - [Monobank#getRates] ---> END HTTP (0-byte body)
```

I suppose that it is Apache HTTP Client's issue and does not relates with Spring or Feing.

radio-rogal:
This loop adds another header for each its value https://github.com/OpenFeign/feign/blob/307a322c96e28339d90fef31e962f107588e6e2b/core/src/main/java/feign/Request.java#L247

muzinian:
Hi,this fix may cause add `Content-Length` twice if `Content-Encoding` not be set.
https://github.com/OpenFeign/feign/blob/7336c3e548612f90f7c3d05e84f3c8793dcef2a4/core/src/main/java/feign/Client.java#L184-L196

izdt:
> Hi,this fix may cause add `Content-Length` twice if `Content-Encoding` not be set.
> 
[...]

PR #2139, opened 2023-07-26
Fix issuse #1721 Accept-Encoding header is added twice

izdt:
Fix issuse #1721 "Accept-Encoding" header is added twice When feign.compression.response.enabled is set to true in configuration, feign adds "Accept-Encoding" header(s) when sending request. Currently, two header fields are added (Accept-Encoding: gzip and Accept-Encoding: deflate).
After fix, "Accept-Encoding" value is "gzip, deflate".

### DIFF
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -14,6 +14,7 @@
 package feign;
 
 import static feign.Util.CONTENT_ENCODING;
+import static feign.Util.ACCEPT_ENCODING;
 import static feign.Util.CONTENT_LENGTH;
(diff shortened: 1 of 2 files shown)


velo:
Is it valid to have duplicated headers?!

The server would need to combine it anyway right, maybe we should combine them all

izdt:
> Is it valid to have duplicated headers?!
> 
> The server would need to combine it anyway right, maybe we should combine them all


Proxy may ignore one of the duplicated headers, then it is not valid.  According to rfc7230 section 3.2.2, mentioned at #2139,  combine them to one is better.

PR #2183, opened 2023-09-26
Optimize moshi decoding

iProdigy:
Follow-up to #2182

* Decode JSON more efficiently by leveraging InputStream rather than Reader so the full response body does not need to be written to an intermediate String
* Eagerly return null if the response body is empty (but not-null)

### Summary by CodeRabbit

- Refactor: Simplified and optimized the `MoshiDecoder` in the Moshi library. The changes include removing unused imports, replacing the `Reader` with `BufferedSource` from Okio for improved performance, and directly parsing JSON using `jsonAdapter.fromJson()`. This refactor enhances the efficiency of the decoding process and simplifies the underlying code structure.

### DIFF
--- a/moshi/src/main/java/feign/moshi/MoshiDecoder.java
+++ b/moshi/src/main/java/feign/moshi/MoshiDecoder.java
@@ -13,19 +13,17 @@
  */
 package feign.moshi;
 
-import com.google.common.io.CharStreams;
 import com.squareup.moshi.JsonAdapter;
 import com.squareup.moshi.JsonDataException;
-import com.squareup.moshi.JsonEncodingException;
 import com.squareup.moshi.Moshi;
 import feign.Response;
 import feign.Util;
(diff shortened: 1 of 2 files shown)


moshi/src/main/java/feign/moshi/MoshiDecoder.java
```diff
@@ -13,19 +13,16 @@
  */
 package feign.moshi;
 
-import com.google.common.io.CharStreams;
```
velo, line 16: Can you also remove the guava dependency from pom?

iProdigy: Good point, done in https://github.com/OpenFeign/feign/pull/2183/commits/f3cd4d8e0b6d3575094d7af040879318bc417838

velo (approved):

Issue #686, opened 2018-04-24
Allow to ignore methods on provided interface


dmarkhas:
When the Feign annotated interface (or class passed to the builder) contains methods that are not intended to be invoked by a web request (i.e. not annotated with RequestMapping), Feign (actually the BaseContract) throws an exception saying the method is not annotated with HTTP method type.

While this is legitimate, I'd like to request an enhancement to allow the contract to ignore methods on demand (kinda like it ignores static methods), for example via some annotation ("@FeignIgnore"?).

The motivation for this is that the same interface class can be used in different contexts, not just for http based communication. Specifically, we use the same interface class both for Feign and for sending messages to remote services via a queue - if there is a method that should only be invoked through a queue, we'd expect it to not be annotated with RequestMapping, which causes Feign to fail to process the class entirely.

rage-shadowman:
IMO, the cleaner solution would be to make multiple interfaces (your concrete class on the server side can implement them both). The one for Feign, the other for your queue.

But if you are unable to do that for some (business-related?) reason, then it is likely you do not have control of the interface and thus can't add any sort of `@FeignIgnore` annotation either.

So if Feign wants to support this, then (again, IMO -- YMMV) I think it should ignore all methods it doesn't recognize as supported.

dmarkhas:
We are doing exactly that just now - splitting the interface :)
It has some drawbacks for us though, since using a single interface would have allowed us to clean up our own APIs.
Your suggestion to ignore anything that is not annotated with RequestLine, RequestMapping, etc. sounds OK, I proposed the annotation simply because it is probably easier to implement and gives the developer finer control over what gets picked up by Feign, but both approaches are OK really.

kdavisk6:
Complete untested and unverified suggestion, but has any tried to extend the interface ala Spring Data?

https://docs.spring.io/spring-data/jpa/docs/2.0.6.RELEASE/reference/html/#repositories.custom-implementations

kdavisk6:
From #615, we may want to consider either adding an explicit `ignore` annotation or update the default `Contract` to only intercept methods annotated.  Thoughts?

XhstormR:
I want to ignore some method on api interface too.

hbprotoss:
Any progress here?

XhstormR:
4 years have passed

ashakirin:
Would be very practical to ignore default methods especilly for generated interfaces, for instance, from Open API

velo:
[...]

PR #2218, opened 2023-10-27
Allow to ignore methods on provided interface

gromspys:
Issue: https://github.com/OpenFeign/feign/issues/686

### Summary by CodeRabbit

- New Feature: Introduced the `@FeignIgnore` annotation. This allows developers to mark certain methods to be ignored during processing, enhancing control over the application's behavior.
- New Feature: Added error handling for attempts to invoke ignored methods. If a method marked with `@FeignIgnore` is called, the system will now throw an `UnsupportedOperationException`, improving the robustness of the application.
- Test: Added new test cases to ensure the correct functioning of the `@FeignIgnore` annotation and the associated error handling mechanism.


velo (changes requested):
I like the change, and makes lots of sense.  But I think it should fire a clear exception if a ignored method is invoked.

velo (approved):

Issue #2237, opened 2023-11-19
Capability can't find correct enrich method


JamieYoungYang:
**test-code:**
public class CapabilityInvokeTest {

static class TestCapability implements Capability {
        public Father enrich(Father father) {
            father.balance = 10;
            return father;
        }
        public Son enrich(Son son) {
            son.balance = 5;
            return son;
        }
    }

    static class Father {
        int balance;
    }

    static class Son extends Father {
    }

    public static void main(String[] args) {
        Son son = new Son();
        Father father = new Father();
        // enrich Son
        Capability.invoke(son, new TestCapability(), Son.class);
        System.out.println(son.balance); // 10
        // enrich Father
        Capability.invoke(father, new TestCapability(), Father.class);
        System.out.println(father.balance); // 10
    }
}

**reason:**Capability.invoke method
Arrays.stream(capability.getClass().getMethods())
        .filter(method -> method.getName().equals("enrich"))
        .filter(method -> method.getReturnType().isAssignableFrom(capabilityToEnrich))
        .findFirst()

**problem:**
`.filter(method -> method.getReturnType().isAssignableFrom(capabilityToEnrich)).findFirst()`

Father.class.isAssignableFrom(Son.class)
Son.class.isAssignableFrom(Son.class)
**both return true**,so findFirst allways return the first method For Father.enrich

There seem to be two ways to solve this problem
**first:** adjust the order in which methods are defined,defind the Son.enrich before the Father.enrich
**second**: `.filter(method -> method.getReturnType() == capabilityToEnrich)`

velo:
Hrmmm, ok, how are you finding this issue with feign classes?

JamieYoungYang:
When I was learning the source code of feign, I accidentally found it, I think it seems to be a bug, so I went to verify my idea

JamieYoungYang:
> Hrmmm, ok, how are you finding this issue with feign classes?

When I was learning the source code of feign, I accidentally found it, I think it seems to be a bug, so I went to verify my idea

JamieYoungYang:
> Hrmmm, ok, how are you finding this issue with feign classes?

May I ask if I can submit merge in the second way to fix this problem, which will give me a great incentive for the future programming road, thank you

velo:
Sure, a pull request is welcome.

JamieYoungYang:
> Sure, a pull request is welcome.

ok, my pleasure

JamieYoungYang:
[...]

PR #2239, opened 2023-11-20
Fix bug in Capability invoke

JamieYoungYang:
to fix the bug, capability can't enrich the correct method
issue:https://github.com/OpenFeign/feign/issues/2237

### DIFF
--- a/core/src/main/java/feign/Capability.java
+++ b/core/src/main/java/feign/Capability.java
@@ -57,7 +57,7 @@ public interface Capability {
   static Object invoke(Object target, Capability capability, Class<?> capabilityToEnrich) {
     return Arrays.stream(capability.getClass().getMethods())
         .filter(method -> method.getName().equals("enrich"))
-        .filter(method -> method.getReturnType().isAssignableFrom(capabilityToEnrich))
+        .filter(method -> method.getReturnType() == capabilityToEnrich)
         .findFirst()
         .map(method -> {
           try {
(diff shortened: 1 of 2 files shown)


velo (changes requested):
Can we get a unit test please?

JamieYoungYang:
> Can we get a unit test please?

Oh yes, of course

core/src/main/java/feign/Capability.java
```diff
@@ -57,7 +57,7 @@ static Object enrich(Object componentToEnrich,
   static Object invoke(Object target, Capability capability, Class<?> capabilityToEnrich) {
     return Arrays.stream(capability.getClass().getMethods())
         .filter(method -> method.getName().equals("enrich"))
-        .filter(method -> method.getReturnType().isAssignableFrom(capabilityToEnrich))
+        .filter(method -> method.getReturnType() == capabilityToEnrich)
```
velo, line 60: I don't think this will work for Feign, all enrichers are based on interfaces....

I even wonder how this didn't broke any unit test.

JamieYoungYang: > I don't think this will work for Feign, all enrichers are based on interfaces....
> 
> I even wonder how this didn't broke any unit test.

I think you're right, if you look at the feign application as a whole, this rarely happens, since I'm thinking about the logic of this place in terms of this method, thank you, I'll close this pull request,

Issue #2245, opened 2023-11-27
HC5 client can't send compressed POST requests


fbus:
ApacheHttp5Client  does not check if there is an 'Accept-Encoding' header when building the HttpRequest

https://github.com/OpenFeign/feign/blob/885f3e3400e564c8a4aa2642629452e9604c552e/hc5/src/main/java/feign/hc5/ApacheHttp5Client.java#L124C12-L124C12

when there is a gzip header, it should create a GzipCompressingEntity to wrap the StringEntity / ByteArrayEntity.

Because of this, the body is not compressed, while my request has an 'Accept-Encoding: gzip' header. And the server replies with a '500 Not in GZIP format' error.

fbus:
as far as I can tell, the default Client is the only one handling properly this case : https://github.com/fbus/feign/blob/885f3e3400e564c8a4aa2642629452e9604c552e/core/src/main/java/feign/Client.java#L225-L226
OkHttp and HC4 are doint have the same bug as hc5

PR #2246, opened 2023-11-27
Add support for request body gzip Content-Encoding

fbus:
When there is a gzip 'Content-Encoding' header, feign should create a GzipCompressingEntity to wrap the StringEntity / ByteArrayEntity.

This is a fix for this issue: https://github.com/OpenFeign/feign/issues/2245


fbus:
I tried to follow the coding style.
I'm really not sure why there's so much code duplication between ApacheHttp5Client / AsyncApacheHttp5Client. They are very similar but AsyncApacheHttp5Client seems to be based on an older copy of ApacheHttp5Client.
I didn't try to solve this though, and just duplicated my fix

fbus:
I closed this one, because editing git history fails to update the pr

PR #2269, opened 2023-12-17
Use bytecode transformation to reduce code duplication on jakarta modules

velo:


jaxrs3/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.moditect</groupId>
+        <artifactId>moditect-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+          <module>
+            <moduleInfo>
+              <exports>!jsr311.api;
```
velo, line 83: @aalmiray any idea what went wrong here?

aalmiray: I think this might have been a split package.

steam0:
Hey @velo, I don't know if I have missed something, but how am I supposed to use JAXRS3Contract with jakarta?

In versions before 13.1 I would use jakarta annotations in my specifications, but after 13.2 when I'm changing from `JakartaContract` to `JAXRS3Contract` my tests fails because the following code fails

Contract.java
```java
     (...)
     
      for (final Annotation methodAnnotation : method.getAnnotations()) {
        processAnnotationOnMethod(data, methodAnnotation, method); // No annotations found
      }
      if (data.isIgnored()) {
        return data;
      }
      checkState(data.template().method() != null,
          "Method %s not annotated with HTTP method type (ex. GET, POST)%s",
          data.configKey(), data.warnings());
          
      (...)
```

I see that the annotation processor fails as my specification uses `@jakarta.ws.rs.Path` but the `GuardedAnnotationProcessor.test` function tries to match it with `javax.ws.rs.Path`.

Screenshot 2024-05-21 at 15 12 21

Is there any documentation that belongs together with this commit that explains how to use jakarta annotations?

velo:
That is odd, how my tests are working?!

Anything peculiar about what you are doing?

PR #2298, opened 2024-01-20
Allow generics deserialization by JacksonJrDecoder

### LINKED ISSUE #2296: JacksonJrDecoder is not able to deserialize generics
# Description
Although `jackson-jr` library is capable of generics deserialization, `feign-jackson-jr` module throws `DecodeException` when you try to decode the response.

# How to reproduce
I created a simple demo to demonstrate this:
```java
public class JacksonJrDemo {
    public static void main(String[] args) throws IOException {
        String json = "{\"data\":2024}";

        DataHolder<?> jrResult = JSON.std.beanFrom(DataHolder.class, json);
        System.out.println(jrResult); // works

        Object jrDecoderResult = new JacksonJrDecoder().decode(
            Response
                .builder()
                .request(Request.create(
                    Request.HttpMethod.GET,
                    "/v1/dummy",
                    Collections.emptyMap(),
                    Request.Body.empty(),
                    null
                ))
                .body(json, StandardCharsets.UTF_8)
                .build(),
            new TypeReference<DataHolder<Integer>>() {}.getType()
        );
        System.out.println(jrDecoderResult); // throws DecodeException
    }

    @Data
    private static class DataHolder<T> {
        private T data;
    }
}
```
Output:
```
JacksonJrDemo.DataHolder(data=2024)
Exception in thread "main" feign.codec.DecodeException: Cannot decode type: io.github.yvasyliev.JacksonJrDemo$DataHolder<java.lang.Integer>
	at feign.jackson.jr.JacksonJrDecoder.findTransformer(JacksonJrDecoder.java:109)
	at feign.jackson.jr.JacksonJrDecoder.decode(JacksonJrDecoder.java:67)
	at io.github.yvasyliev.JacksonJrDemo.main(JacksonJrDemo.java:21)
```
Tested with:
```xml
<dependency>
    <groupId>io.github.openfeign</groupId>
    <artifactId>feign-jackson-jr</artifactId>
    <version>13.1</version>
</dependency>
```
# Initial investigation
I see two problems here:
1. `feign.jackson.jr.JacksonJrDecoder#findTransformer` always throws `DecodeException` unless `type` is instance of `Class`, `List` or `Map`.
2. This behevior is not possible to override since `feign.jackson.jr.JacksonJrDecoder#findTransformer` has `private static` modifiers. (why?)
# Suggestion
1. In this issue I suggest `jackson-jr` to hand the control over generics deserialization. Because now `JacksonJrDecoder` is very limited.
2. I'm going to create another issue to make `feign.jackson.jr.JacksonJrDecoder#findTransformer` overridable.

Both problems requires very few code changes, and I'll try to handle both of them.

Actually, if problem 2 will be solved first, there is no urgent need to solve problem 1. Because in this case the user will be able to handle deserialization by extending `JacksonJrDecoder`. Yet both changes would be great to see.

yvasyliev:
Fixes #2296

### DIFF
--- a/jackson-jr/src/main/java/feign/jackson/jr/JacksonJrDecoder.java
+++ b/jackson-jr/src/main/java/feign/jackson/jr/JacksonJrDecoder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
--- a/jackson-jr/src/main/java/feign/jackson/jr/JacksonJrDecoder.java
+++ b/jackson-jr/src/main/java/feign/jackson/jr/JacksonJrDecoder.java
@@ -93,9 +93,6 @@ public class JacksonJrDecoder extends JacksonJrMapper implements Decoder {
   }
 
   private static Transformer findTransformer(Response response, Type type) {
-    if (type instanceof Class) {
-      return (mapper, reader) -> mapper.beanFrom((Class<?>) type, reader);
-    }
     if (type instanceof ParameterizedType) {
       Type rawType = ((ParameterizedType) type).getRawType();
       Type[] parameterType = ((ParameterizedType) type).getActualTypeArguments();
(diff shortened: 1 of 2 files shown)


velo:
There is some conflict to resolve, can you please take a look?

velo (approved):

PR #2304, opened 2024-01-24
Fixed NullPointerException Bug

pingpingy1:
`RequestTemplate.header(name, values)` method has two overloaded implementations.
In the case where `values` has type `Iterable<String>`, `values` is guarded against `null`.
This does not happen when `values` has type `String...`, which is fixed by this commit.

Concrete testcase:
```java
import feign.RequestTemplate;

public class Test{
	public static void main(String args[]) throws Exception {
		RequestTemplate rt = new RequestTemplate();
		String[] values = null;
		String name = "";
		rt.header(name, values); // NPE
	}
}
```
Result:
```console
Exception in thread "main" java.lang.NullPointerException
        at java.base/java.util.Objects.requireNonNull(Objects.java:233)
        at java.base/java.util.Arrays$ArrayList.<init>(Arrays.java:4238)
        at java.base/java.util.Arrays.asList(Arrays.java:4223)
        at feign.RequestTemplate.header(RequestTemplate.java:706)
        at Test.main(Test.java:8)
```

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -703,6 +703,10 @@ public final class RequestTemplate implements Serializable {
    * @see RequestTemplate#header(String, Iterable)
    */
   public RequestTemplate header(String name, String... values) {
+    if (values == null) {
+      return appendHeader(name, Collections.emptyList());
+    }
+
     return header(name, Arrays.asList(values));
   }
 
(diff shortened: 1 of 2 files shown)


velo (changes requested):
Hi, appreciated the effort you put into fixing this problem.

Seems your change includes some git syntax instructions, please fix that.

Also, It looks like nothing what changed on JacksonCodecTest, would you please revert the formatting changes?

Last, but also a blocked, please create a new test case inside RequestTemplateTest.java that reproduced the issue you found.

Thanks for the assistance

pingpingy1:
@velo Thank you for the kind instructions.
I apologize for not fully obliging by the guidelines; this is my first public PR, and I was quite nervous.
Will promptly get on with it.

velo:
> this is my first public PR, and I was quite nervous

All good, you probably did far better than I did on my first one

velo (approved):

velo:
I will be releasing feign next Friday, this will be included

pingpingy1:
Thank you for the encouraging messages sir!!

velo:
Hey, I'm not going to get more help by being nasty

Look forward your next PR

PR #2305, opened 2024-01-24
Add custom tags to MicrometerCapability

### LINKED ISSUE #1456: [Feature Request] Support custom tags for Micrometer
If I add `.addCapability(new MicrometerCapability(meterRegistry))` to my Feign client, there seems to be no way to add custom tags to the Micrometer metrics. It would be nice to do this, for example by extending `MicrometerCapability` to `MicrometerCapability(meterRegistry, mapOfTags())`.

BartekBanachowicz:
This pull request adds new constructors to `MicrometerCapability`, so the clients are now able to add custom tags to `MeterRegistry` from the builder level.

Two new constructors were added. The first one can take a `List<Tags>`:
```
public MicrometerCapability(MeterRegistry meterRegistry, List<Tag> tagsList)
```

The second one can take a `Map<String, String>` of Keys and Values:
```
public MicrometerCapability(MeterRegistry meterRegistry, Map<String, String> tagsMap)
```
---
Closes #1456

### DIFF
--- a/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
+++ b/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
--- a/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
+++ b/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
@@ -22,8 +22,13 @@ import feign.codec.Encoder;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.simple.SimpleConfig;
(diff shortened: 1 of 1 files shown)


micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
```diff
@@ -38,6 +43,17 @@ public MicrometerCapability(MeterRegistry meterRegistry) {
...
   }
 
+  public MicrometerCapability(MeterRegistry meterRegistry, List<Tag> tagsList) {
+    meterRegistry.config().commonTags(tagsList);
+    this.meterRegistry = meterRegistry;
+  }
+
+  public MicrometerCapability(MeterRegistry meterRegistry, Map<String, String> tagsMap) {
+    List<Tag> tagsList = mapTags(tagsMap);
+    meterRegistry.config().commonTags(tagsList);
```
velo, line 53: Do you think this would be testable under our current tests?

velo (approved):
Look good, any chance to replicate this on dropwizard metrics too?

BartekBanachowicz:
> Look good, any chance to replicate this on dropwizard metrics too?

Of course, but I'd prefer to do it in a separate issue

micrometer/src/main/java/feign/micrometer/MicrometerCapability.java
velo: @BartekBanachowicz what about this?

BartekBanachowicz: For now, the answer from my side is no. Not because it isn't doable, but because I don't know how to achieve it yet. I'm still digging through existing test classes

PR #2311, opened 2024-01-25
Update Client.java

izdt:
Fix Isuue #1721 Content-Length may be added twice


velo (changes requested):
Please take a look on why CI is broken.

Also, please include unit tests so it never happens again

velo:
> OK, I I will add unit tests later.

All good, please reopen the PR once done

Issue #2317, opened 2024-02-02
Breaking change with Alpha version of OkHttp3 Library


danieldisu:
Hello

This [PR](https://github.com/OpenFeign/feign/pull/2261) introduced an alpha version of OkHttp that contains breaking changes. This change in version was introduced in a PR that changes other things and is not mentioned anywhere in the changelog.

- Why it was introduced?
- It is really required to use an Alpha version?
- It is breaking downstream builds that use stable versions of okhttp3

velo:
Hi Daniel,

I don't think it's absolutely necessary. It's quite late for me here, but feel free to send a PR with latest stable version and let's see what CI think about it.

Nice catch, thanks

Cheers

danieldisu:
Thanks @gromspys for the work and @velo for the review 🙇

PR #2318, opened 2024-02-04
Breaking change with Alpha version of OkHttp3 Library

gromspys:
Issue: https://github.com/OpenFeign/feign/issues/2317


velo (approved):
Outstanding work and nice catch.

Will be merging and releasing asap

PR #2329, opened 2024-02-15
OpenFeign#2249 | Support concurrent access to requestInterceptor.

### LINKED ISSUE #2249: Feign.Builder have a requestInterceptor bug
for example:
 public static <T extends BaseApi> T getHttpApi(Class<T> api, String url) {
        ApplicationContext applicationContext = SpringContext.getApplicationContext();
        return applicationContext.getBean(Feign.Builder.class).requestInterceptor(new CustomFeignInterceptor()).target(api, url);
    }
add the requestInterceptor at here cause :
### java.util.ConcurrentModificationException
    at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1631)....
suggest:
```
 /**
     * Adds a single request interceptor to the builder.
     */
    public Builder requestInterceptor(RequestInterceptor requestInterceptor) {
      this.requestInterceptors.add(requestInterceptor);
      return this;
    }
 private final List<RequestInterceptor> requestInterceptors =
        new ArrayList<RequestInterceptor>();
```
**the soureCode requestInterceptors should be a CopyOnWriteArrayList**

youssef3wi:
The Interceptors now have support for concurrent access, which fixes #2249 in this pull request.

### DIFF
--- a/core/src/main/java/feign/BaseBuilder.java
+++ b/core/src/main/java/feign/BaseBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


velo:
Why would you have multiple threads at build time?

youssef3wi:
To simulate the issue.

velo:
I'm not convinced that builder should be thread safe. Special cause it can't be safe safe. You can always build with some concurrency changing an encoder or anything else.

Issue #2332, opened 2024-02-21
feign-micrometer: No client tag on MicrometerObservation meters


iCiaran:
When using the `MicrometerObservationCapability` (as autoconfigured by spring-cloud-openfeign) no information about the client is attached to the `http.client.requests` or `http.client.requests.active` meters.

This can be confusing for applications that have more than one feign client, especially when multiple clients have the same path template resulting in the metrics being merged.

For monitoring it would also useful to be able to filter metrics by client (different downstreams have different latency expected etc).

PR #2333, opened 2024-02-21
Adds a client tag to the default micrometer observation

iCiaran:
Include a client tag on the default feign micrometer observation, so it's added to the `http.client.requests` meters, in the same way it's added to the `feign.*` meters.

This is useful for applications with more than one feign client, to avoid merging unrelated metrics, and improve observability of multiple downstreams which may have different latency etc

Opened to address #2332

### DIFF
--- a/micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
+++ b/micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
--- a/micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
+++ b/micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
@@ -14,6 +14,7 @@
 package feign.micrometer;
 
 import feign.Request;
+import feign.RequestTemplate;
 import feign.Response;
 import io.micrometer.common.KeyValues;
 import io.micrometer.common.lang.Nullable;
@@ -21,8 +22,8 @@ import io.micrometer.common.lang.Nullable;
 /**
  * Default implementation of {@link FeignObservationConvention}.
  *
- * @since 12.1
  * @see FeignObservationConvention
+ * @since 12.1
  */
 public class DefaultFeignObservationConvention implements FeignObservationConvention {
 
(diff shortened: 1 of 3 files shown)


micrometer/src/main/java/feign/micrometer/FeignObservationDocumentation.java
```diff
@@ -74,6 +75,12 @@ public String asString() {
       public String asString() {
         return "net.peer.port";
       }
+    },
+    CLIENT {
+      @Override
+      public String asString() {
+        return "client";
```
iCiaran, line 82: Potentially would be better named  `clientName` to match the spring RestTemplate, WebClient etc

micrometer/src/test/java/feign/micrometer/FeignHeaderInstrumentationTest.java
```diff
@@ -82,9 +88,49 @@ void getTemplatedPathForUriForAsync(WireMockRuntimeInfo wmRuntimeInfo)
...
+  }
+
+  private void assertTags() {
+    Id requestsId = meterRegistry.get(METER_NAME).meter().getId();
+
+    assertMetricIdIncludesMethod(requestsId);
+    assertMetricIdIncludesURI(requestsId);
+    assertMetricIdIncludesStatus(requestsId);
+    assertsMetricIdIncludesClient(requestsId);
+  }
```
iCiaran, line 105: This might not be the right place for these tests, however since they're only applicable to the micrometer observation meters adding to [AbstractMetricsTestBase.java](https://github.com/OpenFeign/feign/blob/master/micrometer/src/test/java/feign/micrometer/AbstractMetricsTestBase.java) would probably require some changes to the dropwizard 4/5 and MicrometerCapability tests

micrometer/src/main/java/feign/micrometer/FeignObservationDocumentation.java
velo: Makes sense

iCiaran: Updated tag to `clientName`

velo (approved):

micrometer/src/main/java/feign/micrometer/DefaultFeignObservationConvention.java
```diff
@@ -48,14 +49,16 @@ public String getContextualName(FeignContext context) {
 
   @Override
   public KeyValues getLowCardinalityKeyValues(FeignContext context) {
-    String templatedUrl = context.getCarrier().requestTemplate().methodMetadata().template().url();
+    RequestTemplate requestTemplate = context.getCarrier().requestTemplate();
```
mihalyr-prospect, line 52: Hey guys, just came across this PR, as I needed to include the client name and was writing a similar implementation, but noticed something in this change. Isn't `context.getCarrier()` nullable and shouldn't we check for `null` like it is done in `getMethodString(@Nullable Request request)` a bit further down?

PR #2361, opened 2024-03-20
Add JAXRS3Decoder to allow using jakarta.ws.rs.core.Response as respo…

### LINKED ISSUE #188: [question] How to use Feign with a JAX-RS interface defining jaxrs.Response as return types
Hi,

I'm using Feign with a JAX-RS interface using JAX-RS Response return types (not feign.Response). 
For instance:

```java
@Path("/users")
@Produces(MediaType.APPLICATION_JSON)
public interface UserContract {
    @POST
    @Path("/")
    @Consumes(MediaType.APPLICATION_JSON)
    public Response create(User user);

    ...
```

When using feign and calling the "create" method, it seems that feigns tries to deserialize the jaxrs.Response object from the HTTP reponse.

Well it makes sense since it's the return type but it's obviously not what I expect. What would be the best way to handle jaxrs.Response type with feign ?

Thanks

chrste90:
…nse type

fixes #188

@velo I tried to create a pr for the old issue https://github.com/OpenFeign/feign/issues/188

What do you think about it? I can add tests if this is likely to be merged.

### DIFF
--- a/jaxrs3/pom.xml
+++ b/jaxrs3/pom.xml
@@ -49,6 +49,11 @@
       <artifactId>jakarta.ws.rs-api</artifactId>
       <version>3.1.0</version>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.17.0</version>
+    </dependency>
 
     <!-- for example -->
     <dependency>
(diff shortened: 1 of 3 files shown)


velo (commented):
I think this is something merge-able... my only concern is dragging jackson (or any other JSON parsing library into the mix).

Is there any official jakarta.ws.Response.magicParser() available?

PR #2405, opened 2024-05-05
Http2Client: added clients cache

illia2m4ch:
In case feign.Request.Options settings differ from the default native client java.net.http.HttpClient, a new client is created for each request, which leads to creation of a huge number of clients in memory.
Added a map for caching soft references of clients.

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -15,6 +15,7 @@ package feign.http2client;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.lang.ref.SoftReference;
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


velo:
This makes me wonder, shouldn't feign be caching and reusing client at core level instead of the implementation?

illia2m4ch:
> This makes me wonder, shouldn't feign be caching and reusing client at core level instead of the implementation?

As I understand it, the problem is that feign has a rich set of parameters for each request (e.g. followRedirects), which may not match the parameters of the native client. Because of this, you need to create a new native client with the necessary parameters.  That is, if a native client supports all the same parameters as feign, it does not have this problem.

Therefore, each native client will have its own key (its own set of parameters that it does not maintain at the request level) to retrieve the client. So I'm not sure if there is a better abstract solution at the core level of feign.

PR #2428, opened 2024-05-21
Using Protocol.name when resolving HTTP protocol for responses

### LINKED ISSUE #2427: Resolving response protocol version using OkHttp3 fails due to OkHttp3.Protocoll class .toString() method not returning name of class
When resolving http version for responses when using OkHttp3 client, the function `enumForName()` fails due to the toString()-method on Okhttp3.Protocoll returning `protocol` instead of `name`.

Screenshot 2024-05-21 at 13 19 02

It would seem like the invocation of `enumForName` should be fed `protocol.name` instead if protocol in order to set the correct response protocol.

steam0:
The `toString()` method on `OkHttp3.Protocol` does not return a value that can be used to resolve to enum `ProtocolVersion`.

This change extracts the `name` from the Protocol and uses that instead of the whole Protocol object.

Fixes #2427

### DIFF
--- a/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
+++ b/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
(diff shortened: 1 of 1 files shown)


velo (approved):

steam0:
@velo, thanks for the merge. When do you expect that the next version will be released?

PR #2444, opened 2024-06-06
fix: Response.charset does not support RFC 7231 compliant Content-Type headers using quotation marks as application/json; charset="utf-8"

stefanvitz:
Hi,

when migrating from rest template to feign i noticed that the feign Response object does not determine the charset in an RFC compliant way. As this charset() method is used e.g. by the JacksonDecoder class decoding an response with an RFC compliant Content-Type header like `application/json; charset="utf-8"` will fail.

See also:
https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1

> For example, the following
>    examples are all equivalent, but the first is preferred for
>    consistency:
> 
>      text/html;charset=utf-8
>      text/html;charset=UTF-8
>      Text/HTML;Charset="utf-8"
>      text/html; charset="utf-8"

### DIFF
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -158,8 +158,8 @@ public final class Response implements Closeable {
 
   /**
    * Nullable and not set when using http/2
-   *
-   * See https://github.com/http2/http2-spec/issues/202
+   * See <a href="https://github.com/http2/http2-spec/issues/202">...</a>
+   * See <a href="https://github.com/http2/http2-spec/issues/202">...</a>
    */
   public String reason() {
     return reason;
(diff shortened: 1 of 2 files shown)


core/src/main/java/feign/Response.java
```diff
@@ -314,7 +320,7 @@ public Reader asReader() {
     }
 
     @Override
-    public Reader asReader(Charset charset) throws IOException {
```
velo, line 317: Can we undo these `throws IOException`?

My concern is breaking existing feign users

stefanvitz: this is just on the implementation which is private final.
see above:

`private static final class InputStreamBody implements Response.Body`
so no feign user would be able to override this. and the exception is not thrown

velo (approved):

Issue #1987, opened 2023-03-21
A bug of header adding and getting in OpenFeign RequestInterceptor


HaojunRen:
Refer to the following issue at Spring Cloud community

https://github.com/spring-cloud/spring-cloud-openfeign/issues/760

radio-rogal:
Hi!

Please try to put keys to doublequoters, it works:
```java
  @Test
  public void headerWithJSONValue() {
    RequestTemplate template = new RequestTemplate()
        .header("key", "{\"x\":1.0,\"y\":1.1}");

    assertThat(template.headers()).hasSize(1);
    assertThat(template.headers().keySet()).containsExactly("key");
    assertThat(template.headers().get("key")).containsExactly("{\"x\":1.0,\"y\":1.1}");
  }
```

radio-rogal:
This is JSON's rule: string are wrapped double quotes, both in keys and values.

And as far as I guess this rule is used to avoid processing of header value, see [Request Headers Expansion](https://github.com/OpenFeign/feign#request-body-expansion) that use curly brackets like JSON does.

kdavisk6:
> It does not make sense, that means i should add double-quoters to every header value with json format one by one?

As @radio-rogal has stated, Feign supports expression based headers.  This means you can use URI templates in your header values and have them resolved at execution time.  The incompatibility you are seeing is related to the use of JSON as a header value, which conflicts with the URI Template specification where dynamic expressions are wrapped in braces `{}`.  

If you chose to use Feign and require JSON based header values, you will need to quote them as discussed.

yangyang122:
![1699345955608](https://github.com/OpenFeign/feign/assets/12775735/59808171-7f3a-485f-bddb-330a2298f123)

Can you add a switch here to enable parsing JSON data？

yangyang122:
I try : 
 ` String data="{\"x\":1.0,\"y\":2.0}";
        RequestTemplate rq = requestTemplate.header("v", data);
        Map<String, Collection<String>> headers = rq.headers();
        headers.forEach((key, values) -> {
            System.out.println(key+"--"+values);
        });`
![Uploading 图片.png…]()  
but it's not the expected value.

hyfsy:
> Hi!
> 
> Please try to put keys to doublequoters, it works:
> 
> ```java
>   @Test
>   public void headerWithJSONValue() {
>     RequestTemplate template = new RequestTemplate()
[...]

PR #2471, opened 2024-07-10
Allows adding literal header values to RequestTemplate

### LINKED ISSUE #2252: Can't set request header in json format
image

dvag-yannick-reifschneider:
This change adds a new method RequestTemplate#headerLiteral which allows adding headers which are not interpreted as Template expressions. This allows adding empty JSON objects to headers in RequestInterceptor implementations.

Fixes #2252, #1987

### DIFF
--- a/core/src/main/java/feign/RequestTemplate.java
+++ b/core/src/main/java/feign/RequestTemplate.java
@@ -729,6 +729,36 @@ public final class RequestTemplate implements Serializable {
   }
 
   /**
+   * @see RequestTemplate#headerLiteral(String, Iterable)
+   */
+  public RequestTemplate headerLiteral(String name, String... values) {
+    if (values == null) {
+      return appendHeader(name, Collections.emptyList(), true);
+    }
+
+    return headerLiteral(name, Arrays.asList(values));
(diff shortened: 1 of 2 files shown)


core/src/main/java/feign/RequestTemplate.java
```diff
@@ -728,6 +728,36 @@ public RequestTemplate header(String name, Iterable<String> values) {
     return appendHeader(name, values);
   }
 
+  /**
+   * @see RequestTemplate#headerLiteral(String, Iterable)
+   */
+  public RequestTemplate headerLiteral(String name, String... values) {
+    if (values == null) {
+      return appendHeader(name, Collections.emptyList(), true);
```
velo, line 736: Any reason not to call `headerLiteral` here?

velo (approved):

core/src/main/java/feign/RequestTemplate.java
dvag-yannick-reifschneider: No specific reason other then to align this code with the implementation of RequestTemplate#header(...)

PR #2530, opened 2024-08-29
Closes Response.body() in Logger after rebuffering to avoid resources leak

### LINKED ISSUE #2529: feign.Logger doesn't close response.body()
When feign.Logger copies Response in method `logAndRebufferResponse`, it makes new instance of Response.Body, but it doesn't close original Response.Body. 
It only closes underlying InputStream in the next line.
https://github.com/OpenFeign/feign/blob/9b9d74ff18f8a41a9fa9061cf56f62f62549bc94/core/src/main/java/feign/Logger.java#L128

Looks like, it could be potential leak of resources/connections. What do you think?

For example Response.Body implementation for apache hc5 is never executed.
```
      @Override
      public void close() throws IOException {
        try {
          EntityUtils.consume(entity);
        } finally {
          httpResponse.close();
        }
```

Nazukin-Dmitry:
Fixes #2529

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


velo (changes requested):
The change look good, may I just get a unit test covering it for future generations?!


cheers

Nazukin-Dmitry:
@velo please have a look. 
I didn't add test into LoggerTest because it is used as parent, so decided to create new test class

velo (approved):

PR #2531, opened 2024-08-30
Polish static nested test classes

Nazukin-Dmitry:
IDEA suggested to make some changes:

- Removed `@Nested` annotaion from inner static tests.
`@Nested` annotation is designed for nestsed **non-static** classes. From javadoc:
> @Nested is used to signal that the annotated class is a nested, non-static test class (i.e., an inner class) that can share setup and state with an instance of its enclosing class.

- Renamed some inner classes to have suffix Test.

- Changed exclude configuration in maven-surefire-plugin to execute static nested test classes.
By default plugin excludes all inner classes.
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#excludes
https://github.com/junit-team/junit5/issues/1377#issuecomment-381964988

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -480,7 +480,7 @@
             <trimStackTrace>false</trimStackTrace>
             <argLine>${jvm.options}</argLine>
             <excludes>
-              <exclude/>
+              <exclude></exclude>
             </excludes>
           </configuration>
         </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -479,6 +479,9 @@
             <redirectTestOutputToFile>true</redirectTestOutputToFile>
             <trimStackTrace>false</trimStackTrace>
             <argLine>${jvm.options}</argLine>
+            <excludes>
+              <exclude/>
+            </excludes>
           </configuration>
         </plugin>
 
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -479,6 +479,9 @@
             <redirectTestOutputToFile>true</redirectTestOutputToFile>
             <trimStackTrace>false</trimStackTrace>
             <argLine>${jvm.options}</argLine>
+            <excludes>
```
velo, line 482: Why is this needed?

Nazukin-Dmitry: I found that inner static test classes are not executed. Tests inside LoggerTest are ignored.

https://github.com/junit-team/junit5/issues/1377#issuecomment-381964988
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#excludes

Nazukin-Dmitry: From other pull request
https://app.circleci.com/pipelines/github/OpenFeign/feign/2890/workflows/95131078-58e3-4743-8fbf-b8c17d602ef5/jobs/8126?invite=true#step-105-18848_57

From this one
https://app.circleci.com/pipelines/github/OpenFeign/feign/2892/workflows/216b89f1-c576-4130-a322-cf2c23646607/jobs/8128?invite=true#step-105-21007_57

velo (approved):

PR #2545, opened 2024-09-11
Fixed extracting charset from response.

### LINKED ISSUE #2540: Building FeignException can throw llegalCharsetNameException
WHERE:
In this class https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/FeignException.java is piece of code:

```
if (!Charset.isSupported(group)) {
    return null;
}
```

WHATS WRONG:
Some external apis can send header Content-Type: text/xml;charset="utf-8" in response. Problem is with quotes.
Method Charset.isSupported throws exception when quotes are part of input string as it checks valid characters.
So building FeignException throws exception in this case.

EXPECTED BEHAVIOUR:
IF statement does not throw exception. It returns true or null in this case. We can't change external api and preprocess response headers is impossible as they are unmodifiable.

SOLUTION:
A. If you want to return true just sanitize String input group and remove for example quotes. Or change regexp to extract it without quotes.
`group.replaceAll("\"", "");`
B. Catch llegalCharsetNameException and return null from IF statement.

kubav182:
While building FeignException, ignore quotes in regexp, ignore case and catch IllegalCharsetNameException to be sure it does not raise in other cases.

These content types are all valid and results in utf-8: text/html;charset=utf-8
text/html;charset=UTF-8
Text/HTML;Charset="utf-8"
text/html; charset="utf-8"

Fixes #2540

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 The Feign Authors
+ * Copyright 2012-2024 The Feign Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -21,6 +21,7 @@ import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


kdavisk6:
Please correct the build by running `./mvnw install` locally at least once to ensure formatting, test cases, and license requirements.

kubav182:
I ran install with toolchain skipped. License header was changed. I also added parametrized test to FeignExceptionTest to cover different cases.

kdavisk6 (approved):

kdavisk6:
These changes are approved.  We'll leave this PR open for a few more days to allow for additional feedback.  If there is none, we'll merge this. 

Thanks again for your help!

Issue #2068, opened 2023-05-30
POST requests with empty body have "Content-Type: application/x-www-form-urlencoded" header since Feign 12.0


mgr32:
### Description

Consider the following client sending a POST request with no body:
```
public interface TestFeignClient {
  @RequestLine("POST /post")
  String testPost();
}
```

Starting from Feign 12.0 requests from such clients contain an additional header `Content-Type: application/x-www-form-urlencoded`, which seems to be:
* unexpected - why such type? (see also these discussions on Stack Overflow - [#1](https://stackoverflow.com/a/21029402), [#2](https://stackoverflow.com/a/15032957))
* a breaking change (not listed in the [12.0 release notes](https://github.com/OpenFeign/feign/releases/tag/12.0)), that can cause errors such as "Media type not supported" on the server side

### Expected behavior

POST requests with no body should by default add `Content-Length: 0` header (as described in [#1229](https://github.com/OpenFeign/feign/issues/1229)), but no `Content-Type` header.

### Steps to reproduce

See a minimal example at https://github.com/mgr32/feign-post-empty-body-issue.

### Analysis

* In Feign 11.10 when a request has no body, Feign’s `Client` does not send anything to outputStream ([code](https://github.com/OpenFeign/feign/blob/b0c5db0ddfd24e0515a9143d82353a8d03def32d/core/src/main/java/feign/Client.java#L199)), and JDK class `HttpUrlConnection` does not add `Content-type` in this case (the if [here](https://github.com/openjdk/jdk17/blob/4afbcaf55383ec2f5da53282a1547bac3d099e9d/src/java.base/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L696) does not match because `poster == null`)
* In Feign >= 12.0 when a request has no body and the request method is POST, Feign’s `Client` sends an empty array to outputStream ([code](https://github.com/OpenFeign/feign/blob/68c92775a28e4d295ed780089751305b695a95a1/core/src/main/java/feign/Client.java#L201)), and JDK class `HttpUrlConnection` adds `Content-type` in this case (the if [here](https://github.com/openjdk/jdk17/blob/4afbcaf55383ec2f5da53282a1547bac3d099e9d/src/java.base/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L696) matches because `poster != null`)
[...]

PR #2555, opened 2024-09-18
Remove Content-Type Header when request with empty body POST method

jay-choe:
Hi.

I have looked [this issue](https://github.com/OpenFeign/feign/issues/2068) and got an idea to deal with.

In current code, convert null body to not null and it write to outputstream.

In that process, when writing empty string in outputstream,  unwanted **Content-Type** Header is added which is ''application/x-www-form-urlencoded".

So, I rearrange null body code to avoid unwanted header.

With this, **Content-Length** header is not set, so I added. But, this header is restricted header so **'sun.net.http.allowRestrictedHeaders'** system property should be true to add content-length header.

And this property should be set before client class loaded. I guess there is no suitable way to inject restricted header to connection.

Feel Free to comment this idea. thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -815,6 +815,12 @@ in a context-specific manner -- for example, thread-local storage can be used to
 header values depending on the invoking thread, which can be useful for things such as setting
 thread-specific trace identifiers for requests.
 
+#### Set zero Content-Length Header
+
+To specify `Content-Length: 0` header when making a request with empty body, system property `sun.net.http.allowRestrictedHeaders` should be set to `true`
+
+If not, the `Content-Length` header will not be added.
+
 ### Advanced usage
 
 #### Base Apis
--- a/core/src/main/java/feign/Client.java
+++ b/core/src/main/java/feign/Client.java
@@ -204,19 +204,14 @@ public interface Client {
         connection.addRequestProperty("Accept", "*/*");
       }
 
-      boolean hasEmptyBody = false;
       byte[] body = request.body();
-      if (body == null && request.httpMethod().isWithBody()) {
-        body = new byte[0];
-        hasEmptyBody = true;
-      }
 
       if (body != null) {
         /*
          * Ignore disableRequestBuffering flag if the empty body was set, to ensure that internal
          * retry logic applies to such requests.
          */
-        if (disableRequestBuffering && !hasEmptyBody) {
+        if (disableRequestBuffering) {
           if (contentLength != null) {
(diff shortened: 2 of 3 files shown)


kdavisk6:
This look good.  Could you please add some information to the README to indicate how to take advantage of this change, helping folks understand the System Property requirement?

jay-choe:
@kdavisk6 sure, I added it.

velo:
@jay-choe @kdavisk6 FWIW, I have rebased this PR after the formatting changes I did on master

velo (approved):

zoomout:
Please revert this change. Setting "sun.net.http.allowRestrictedHeaders" is breaking security. 
restrictedHeaders include e.g. Host, enabling CSRF attack.

    private static final String[] restrictedHeaders = {
        /* Restricted by XMLHttpRequest2 */
        //"Accept-Charset",
        //"Accept-Encoding",
        "Access-Control-Request-Headers",
        "Access-Control-Request-Method",
        "Connection", /* close is allowed */
        "Content-Length",
        //"Cookie",
        //"Cookie2",
        "Content-Transfer-Encoding",
        //"Date",
        //"Expect",
        "Host",
        "Keep-Alive",
        "Origin",
        // "Referer",
        // "TE",
        "Trailer",
        "Transfer-Encoding",
        "Upgrade",
        //"User-Agent",
        "Via"
    };

velo:
Hello, good morning, how are you today?

Raise a PR with test case.

Thank you.

evgeniycheban:
Hi @velo, this PR turned out into a breaking change for us, currently we have some POST requests sent with empty body, for example:
```java
@PostMapping("/action/{actionId}")
String performAction(@PathVariable Long actionId);
```
it fails when the request goes through our gateway proxy which forces us to either send the request with an empty body
`"{}"` or implement an interceptor that would add `"Content-Length": 0` header for POST requests with empty body. I have looked at the code and this change also does not meet the initial requirement from gh-2068:
```
POST requests with no body should by default add Content-Length: 0 header (as described in https://github.com/OpenFeign/feign/issues/1229), but no Content-Type header.
```
We cannot use `sun.net.http.allowRestrictedHeaders` either due to environment restrictions and I also believe that setting this property could lead to potential security risks and Feign should rely on it in order to be able to set `"Content-Length": 0`, instead I think a new configuration option for `Feign.Builder` should be introduced that would enable an old behaviour by default to restore backward compatibility, I can also provide a PR for this and to address the initial gh-2068 issue.
What do you think?

PR #2584, opened 2024-10-03
refactor(jackson): remove unused imports

dancer1325:
## Description
* Remove unused imports

## How to review?
* Check in https://github.com/OpenFeign/feign/blob/master/jackson/src/main/java/feign/jackson/JacksonDecoder.java, that it's NOT used
  * `java.nio.charset.Charset`
  * `java.util.Collection`


velo (approved):
Nice catch!

@kdavisk6 I think it's about time we move to google formatter instead of my own baked eclipse code formatter

kdavisk6:
Agreed

Issue #1383, opened 2021-04-16
Allow optional request body


OlgaMaciaszek:
Allow making request body optional. This relates to https://github.com/OpenFeign/feign/pull/583. See [comment](https://github.com/OpenFeign/feign/pull/583#issuecomment-636326555).

We have an open [issue](https://github.com/spring-cloud/spring-cloud-openfeign/issues/126) for which we would use this feature in SC OpenFeign and also a PR with an [alternative solution](https://github.com/spring-cloud/spring-cloud-openfeign/pull/263), however, we think it would be a cleaner solution if it were supported in OpenFeign directly.

kdavisk6:
Seems simple enough.  I'll look into the documentation to see if there's anything that explicitly requires that a non-null body

nickmafra:
Any progress?

hhhyang:
any progress? this requirement has been talked from year 2016, haha

murtazakhussain:
Any progress? 2024 is about to start. Looking at the previous conversation its been 7 years

velo:
Still waiting a PR.

betinro:
Any news on this? We need this fix as well.

velo:
> Any news on this? We need this fix as well.

Still waiting for your PR

gromspys:
@velo @betinro PR is ready and waiting for review. https://github.com/OpenFeign/feign/pull/2616

PR #2616, opened 2024-10-22
Allow optional request body

gromspys:
Issue: https://github.com/OpenFeign/feign/issues/1383


velo:
Although I have approved this, I'm still unsure... why not having a method that just omits the body and be a bit more deterministic? I bet ya the average client will have other methods where the body is not optional and marking all bodies as optional will bring lot's of pain.

gromspys:
I think there is some restrictions (client can be generated by swagger or something else). Also parameter can be passed from upstream service and can be null.

Issue #2618, opened 2024-10-24
FeignException::responseBody() returns request body instead of response body


BartoszMiller:
I wanted to retrieve response body from a FeignException, but `FeignException::responseBody()` returns request body of my request. Is it intentional or is it a bug? One of the FeignException constructors, as the parameter name implies, expects response body. However, in the `errorReading` method `request.body()` is passed instead.

```
static FeignException errorReading(Request request, Response response, IOException cause) {
    return new FeignException(
        response.status(),
        format("%s reading %s %s", cause.getMessage(), request.httpMethod(), request.url()),
        request,
        cause,
        request.body(),
        request.headers());
  }
```

```
protected FeignException(
      int status,
      String message,
      Request request,
      Throwable cause,
      byte[] responseBody,
      Map<String, Collection<String>> responseHeaders) {
    super(message, cause);
    this.status = status;
    this.responseBody = responseBody;
    this.responseHeaders = caseInsensitiveCopyOf(responseHeaders);
    this.request = checkRequestNotNull(request);
  }
```

gromspys:
Yes. It looks like a bug.

Aryant-Tripathi:
Hey @gromspys, I want to work around feign and learn about it. Do we need to change the constructor params since it expects the response body as byte[] and response? getBody() is a String in nature. 

What do you think we can do here?

PR #2619, opened 2024-10-24
FeignException::responseBody() returns request body instead of response body

gromspys:
Issue: https://github.com/OpenFeign/feign/issues/2618

### DIFF
--- a/core/src/test/java/feign/FeignUnderAsyncTest.java
+++ b/core/src/test/java/feign/FeignUnderAsyncTest.java
@@ -490,7 +490,7 @@ public class FeignUnderAsyncTest {
     } catch (FeignException e) {
       assertThat(e.getMessage())
           .isEqualTo("timeout reading POST http://localhost:" + server.getPort() + "/");
-      assertThat(e.contentUTF8()).isEqualTo("Request body");
+      assertThat(e.contentUTF8()).isEqualTo("success!");
     }
   }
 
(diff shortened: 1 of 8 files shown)


core/src/test/java/feign/FeignUnderAsyncTest.java
```diff
@@ -490,7 +490,7 @@ void throwsFeignExceptionIncludingBody() {
     } catch (FeignException e) {
       assertThat(e.getMessage())
           .isEqualTo("timeout reading POST http://localhost:" + server.getPort() + "/");
-      assertThat(e.contentUTF8()).isEqualTo("Request body");
```
velo, line 493: So, backwards compatibility is my main concern here.

I think we need to deprecate `contentUTF8`, keep response body as result of `contentUTF8` and introduce `resquestBody` and `responseBody` or something with that meaning.

Breaking compatibility like this only on major releases.

gromspys: It looks like there is also the same bug, because `contentUTF8()` method contains following code:
```
public String contentUTF8() {
    if (responseBody != null) {
      return new String(responseBody, UTF_8);
    } else {
      return "";
    }
  }
```
Also `content()` method is deprecated and method `responseBody()` already exists.

gromspys: For example in this [case](https://github.com/OpenFeign/feign/blob/2396c4b51d9e57e097b300b007907394dcc18186/core/src/test/java/feign/codec/DefaultErrorDecoderTest.java#L61) `contentUTF8()` returns response body `hello world`

Issue #2439, opened 2024-06-03
AsyncApacheHttp5Client : Proposal to make creating the callback FutureCallback<SimpleHttpResponse>> through a public factory method


predatoralpha:
Hello,
I was faced with the problem to save thread-dependent data from the request-thread in an asynchronous **httpclient-dispatch** thread, for correctly output the necessary information in the **Slf4jLogger** logs, for example parentRequestId
The point of intersection of these threads is creation **FutureCallback<SimpleHttpResponse>>** - here thread-dependent information from request-thread can be stored in the callback for an asynchronous thread
I make workaround - I implemented this by proxying the class **CloseableHttpAsyncClient** and reflectively replacing the callback with its wrapper, with the necessary information from the request-thread to async
However, it would be convenient to have a public factory in the class with the ability to replace this factory with your own, for more convenient creation of a wrapper over a callback.
I suggest this code: 
https://github.com/predatoralpha/AsyncApacheHttp5Client_proposal/blob/main/AsyncApacheHttp5Client.java
Here the factory is implemented via **BiFunction<Request, CompletableFuture<Response>, FutureCallback<SimpleHttpResponse>>**

kdavisk6:
Thank you for your proposal and it seems like a reasonable addition.  I recommend creating a pull-request to get direct feedback on your implementation.

PR #2626, opened 2024-10-30
AsyncApacheHttp5Client_proposal

predatoralpha:
https://github.com/OpenFeign/feign/issues/2439

### DIFF
--- a/hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
+++ b/hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
@@ -23,6 +23,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BiFunction;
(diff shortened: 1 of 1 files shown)


hc5/src/main/java/feign/hc5/AsyncApacheHttp5Client.java
```diff
@@ -52,6 +53,8 @@ public final class AsyncApacheHttp5Client implements AsyncClient<HttpClientConte
 
   private final CloseableHttpAsyncClient client;
 
+  protected volatile BiFunction<Request, CompletableFuture<Response>, FutureCallback<SimpleHttpResponse>> futureCallbackFactory;
```
velo, line 56: I'm not against having a factory, but it should be something final that you set and forget...

if you need different handling on different clients, you use different builders for it....

So, can we make it final and include as part of the constructor and use current value when none is present?!

velo (changes requested):

velo (commented):
Please also include some unit testing for new functionality

PR #2631, opened 2024-11-01
Integrate Feign Vertx into the main project

hosuaby:
This PR is mostly merge of a separate stable project `feign-vertx` into the main repository. Please, feel free to take your time to review it.


core/src/main/java/feign/RequestTemplateFactoryResolver.java
```diff
@@ -47,7 +47,7 @@ public RequestTemplate.Factory resolve(Target<?> target, MethodMetadata md) {
     }
   }
 
-  private static class BuildTemplateByResolvingArgs implements RequestTemplate.Factory {
```
hosuaby, line 50: Make those classes package private, so I don't need to copy them in `feign-vertx` project.

vertx/run-tests.zsh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+declare -A vertx_versions
+vertx_versions=( [v40x]="4.0.x", [v41x]="4.1.x", [v42x]="4.2.x", [v43x]="4.3.x", [v44x]="4.4.x", [v45x]="4.5.x" )
+v40x=( "4.0.2" )
+v41x=( "4.1.8" )
+v42x=( "4.2.7" )
+v43x=( "4.3.2" )
+v44x=( "4.4.9" )
+v45x=( "4.5.10" )
```
velo, line 33: so you test it with a matrix of versions?!

we need to mimic that into circle ci build

vertx/src/main/java/feign/VertxFeign.java
```diff
@@ -0,0 +1,476 @@
...
+
+    for (final DefaultMethodHandler defaultMethodHandler : defaultMethodHandlers) {
+      defaultMethodHandler.bindTo(proxy);
+    }
+
+    return proxy;
+  }
+
+  /** VertxFeign builder. */
+  public static final class Builder extends Feign.Builder {
```
velo, line 94: You might want to take a look on this one:
https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/AsyncFeign.java#L69

velo (approved):
LGTM

velo:
Don't you want to preserve history? it's a great way to highlight all your works and other contributors too.

vertx/run-tests.zsh
hosuaby: I think, I even need to create separate submodule in feign-vertx to ensure a runtime compatibility with different versions of Vertx.

Do you have something similar in the project already ? Any suggestions how to integrate into CI ?

vertx/src/main/java/feign/VertxFeign.java
hosuaby: Checked it already. `AsyncFeign` is unfit because it's based on `CompletableFuture`s from JDK. As I told before, Verx uses its own `Future` implementation, which should be supported both internally and used in contracts. My project is more similar to `feign-rective` in that sense.

hosuaby:
> Don't you want to preserve history? it's a great way to highlight all your works and other contributors too.

Thats would be cool. How can I do that? List them as contributors in `pom.xml`?

velo:
> > Don't you want to preserve history? it's a great way to highlight all your works and other contributors too.
> 
> Thats would be cool. How can I do that? 

I merged vertx repo into this one, that way all history is preserved

PR #2644, opened 2024-11-10
Fixed feign-micrometer exception handling.

### LINKED ISSUE #2566: [BUG] MicrometerObservationCapability not reporting timeouts
Copied from https://github.com/spring-cloud/spring-cloud-openfeign/issues/958
Seems like nobody created an issue here as requested in the original issue 

> **Describe the bug** Feign cloud version: `4.1.0`
> 
> When using the MicrometerObservationCapability, any` java.net.SocketTimeoutException` is not caught and observations is never closed (`MicrometerObservationCapability#enrich only catches FeignExceptions`). Probably affects other errors from the underlying web layer too. The result is that these requests are never added to the metrics. There is either an ordering error, or `MicrometerObservationCapability` should catch more types of errors
> 
> **Sample** No specific config added. Plain FeignClients
> 
> ```
> @FeignClient(name = "NAME", url = "${url.prop}")
> interface Client {
> // calls here
> }
> 
> @Configuration
> class FeignConfig {
>   @Bean
>   def observation(observationRegistry: ObservationRegistry): MicrometerObservationCapability = new MicrometerObservationCapability(observationRegistry, new DefaultFeignObservationConvention())
> }
> ```
> 
> application.yml
> 
> ```
> spring:
>   cloud:
>     openfeign:
>       client:
>         config:
>           default:
>             connect-timeout: 3000
>             read-timeout: 3000
> ```
> 
> Stacktrace
> 
> ```
> Caused by: java.net.SocketTimeoutException: Read timed out
> 	at java.base/sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:288)
> 	at java.base/sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:314)
> 	at java.base/sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:355)
> 	at java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:808)
> 	at java.base/java.net.Socket$SocketInputStream.read(Socket.java:966)
> 	at java.base/java.io.BufferedInputStream.fill(BufferedInputStream.java:244)
> 	at java.base/java.io.BufferedInputStream.read1(BufferedInputStream.java:284)
> 	at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:343)
> 	at java.base/sun.net.www.http.HttpClient.parseHTTPHeader(HttpClient.java:826)
> 	at java.base/sun.net.www.http.HttpClient.parseHTTP(HttpClient.java:761)
> 	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1688)
> 	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1589)
> 	at java.base/java.net.HttpURLConnection.getResponseCode(HttpURLConnection.java:529)
> 	at feign.Client$Default.convertResponse(Client.java:111)
> 	at feign.Client$Default.execute(Client.java:107)
> 	at feign.micrometer.MicrometerObservationCapability.lambda$enrich$1(MicrometerObservationCapability.java:53)
> 	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:100)
> ```

r-dimitrov:
The current MR improves exception handling for **feign-micrometer** clients. Previously, exceptions such as **404**, **400**, **500**, etc., were not marked with error=true in the Observation and thus were not reported in the trace as errors.

The exception declaration has been changed from (**FeignException ex**) to (**Exception ex**), as none of the supported clients (e.g., Apache HttpClient, OkHttp) throws FeignException, but rather client-specific exceptions.

The fix is currently in production in my organization, and it has been working effectively.

Fixes #2566


r-dimitrov:
I've just added some fixes, as there were exceptions when some clients had defined a custom ErrorDecoder. The input stream is consumed, and the decoder throws an exception. This way, we are copying the input stream so it can be used multiple times.

micrometer/src/main/java/feign/micrometer/MicrometerObservationCapability.java
```diff
@@ -57,9 +59,21 @@ public Client enrich(Client client) {
 
       try {
         Response response = client.execute(request, options);
-        finalizeObservation(feignContext, observation, null, response);
+
+        if (response.body() != null) {
+          response =
+              response.toBuilder().body(Util.toByteArray(response.body().asInputStream())).build();
+        }
```
kdavisk6, line 66: If your intent is to properly flag the request with the appropriate error based on status, reading the response into memory _before_ checking the status code may be problematic in certain scenarios.  Particularly in one where a large number of successful responses are monitored with large data sets.  

By doing this, we impress upon our users that their applications must be able to stream all data from all tracked operations into memory in order to observe this.  

Can you help me understand the exceptions observed when trying to observe a client with a custom `ErrorDecoder`?

If this must be part of the solution, we should document this assumption and control this behavior with a configurable feature flag, initially off, and allow our users to make the decision whether to take this on.

kdavisk6 (changes requested):
Can we discuss the need to read the request body into memory more?

micrometer/src/main/java/feign/micrometer/MicrometerObservationCapability.java
r-dimitrov: Hey, @kdavisk6 , thank you for your comment. I've made the changes to cache exceptions (errors) only, so successful responses are no longer cached.

Regarding the custom `ErrorDecoder`, the issue is that when the exception response message body is read, the stream is consumed in `MicrometerObservationCapability.java`, making it unavailable in the `ErrorDecoder`. This is problematic because, in most cases, we need to read the response again and map it to another object. When the response is cached, we can read the response as many times as needed. For example, this is useful if we have loggers, additional logic in filters, and so on.

To summarize:
We need to read the response body for exceptions (client/server error codes in the response) to mark the span with error details and exception information. If we don't cache the response, we cannot read it in another layer, such as the `ErrorDecoder`, because it is already consumed for observability purposes.

kdavisk6: Thanks for the clarification.  

Reviewing `MicrometerObservabilityCapability`, I'm do not see it explicitly consuming the `Response.body`, only wrapping it.  Do you have an example where the `Response` is consumed during metrics processing?  I'd like to confirm that this is intentional or not.  If not, then we should address that, which should remove the need for this workaround.

If it is intentional, I'm still leery of allowing this without some level of user control.  I would like to see some way for users to control this behavior.  There may be situations where even the error response may be large enough to cause memory pressure and we should not be pressing this onto our users without their consent.  

One way is for us to check the `Content-Length` on the response and if that is over a certain threshold, we should defer streaming the response into memory.  Another is to have a configurable option as part of the `MicrometerCapability` that  explicitly enables response caching.

r-dimitrov: Hello, @kdavisk6,
The `response` is not explicitly consumed but is necessary for tracking the exact exception returned from the given HTTP call. To mark the span as an exception, a `Throwable` object is required, which is subsequently used here:
[https://github.com/micrometer-metrics/tracing/blob/e64b9195d0aec18c74353e274d9fef05022061f4/micrometer-tracing-bridges/micrometer-tracing-bridge-otel/src/main/java/io/micrometer/tracing/otel/bridge/OtelSpan.java#L171](https://github.com/micrometer-metrics/tracing/blob/e64b9195d0aec18c74353e274d9fef05022061f4/micrometer-tracing-bridges/micrometer-tracing-bridge-otel/src/main/java/io/micrometer/tracing/otel/bridge/OtelSpan.java#L171)
As you can see getMessage() is invoked for printing the details about the exception.
Without consuming and mapping the response to an Exception (Throwable), it is not possible to capture the core root cause of the exception. The image below shows the desired behaviour.

Screenshot 2025-01-28 at 15 57 34

velo:
Ops, closed by mistake, my bad

r-dimitrov:
@velo @kdavisk6 Hello, guys! Could we proceed with merging the PR? It's in good shape, and it's truly a must-have feature when it comes to distributed tracing. Thank you so much for all the great work you do — it's truly appreciated!

velo (approved):

Issue #2250, opened 2023-11-28
RetryableException should hold the methodKey as well


nlescure-pictet:
Hi there,

Referring to https://github.com/OpenFeign/feign/issues/719, I think it would make even more sense to add the methodKey to the RetryableException so that in the Retryer I can exactly pinpoint which method is being retried as it is done in the ErrorDecoder. That way, both implementations use the methodKey approach instead of having one relying on url/method approach and the other one on the methodKey one.

Let me know what you guys think.

Cheers

Regyl:
Hello @kdavisk6 , can I take this?

kdavisk6:
Please feel free.

PR #2714, opened 2025-01-05
Add methodKey to Request & RetryableException

Regyl:
Changes:

- Add passing MethodMetadata#configKey to Request/RetryableException
- Add tests

Fixes https://github.com/OpenFeign/feign/issues/2250


velo (changes requested):
I'm not sure about this...

Massive change that can be easily replaced by `request.requestTemplate().methodKey()`

Any particular reason not to use that?

PR #2738, opened 2025-01-21
Add gzip and deflate compression support in Http2Client

hicom150:
This PR has the following changes in order to add support for **gzip** and **deflate** compression.

- Enable Http2Client tests
- Fix Http2Client `timeoutTest`:
It seems that Java 11 HttpClient timeout stops when response headers are received. https://bugs.openjdk.org/browse/JDK-8208693
So to fix the test I adjust the mock accordingly.
- Override some `AbstractClientTests` as Http2Client does not expose reason phrase:
Java 11 [HttpClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html) does not expose reason phrase.
- Add gzip and deflate compression support in Http2Client

### DIFF
--- a/core/src/test/java/feign/client/AbstractClientTest.java
+++ b/core/src/test/java/feign/client/AbstractClientTest.java
@@ -127,7 +127,7 @@ public abstract class AbstractClientTest {
   }
 
   @Test
-  void parsesErrorResponse() {
+  public void parsesErrorResponse() {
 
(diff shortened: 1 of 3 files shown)


hicom150:
@velo Could you help pushing this PR forward?
Thank you!

velo (approved):

Issue #2734, opened 2025-01-18
Add support for streaming on requests (and add better handling for streaming in responses)


trumpetinc:
Related issue - my apologies if this is a double post - I wasn't sure how to make sure the discussion was being viewed by the appropriate people: https://github.com/OpenFeign/feign/issues/1243

### Problem
Feign currently uses a byte[] when processing the request body.  This does not work for transfers of large request bodies (for example, uploading large files).

### Requested solution
Make request bodies be designed around streams.

### Status/Findings
I just finished up creating a bunch of shims to make streaming requests work in Feign, and would like to see if ya'll would be open to having a discussion about how to make this part of the main library.  This works, but it is ugly.

Doing this properly will require a breaking change, so I am hesitant to create a pull request without some good discussion about the best way to do it, where we can and can't break API, etc...

### Possible Approach
Based on the workaround implementation I created, I believe that the general approach will be as follows:

First, the existing Response.Body class is perfectly suitable for handing both Request *and* Responses.  It actually looks like Response.Body was designed with this in mind.  I think that refactoring this into a standalone Body interface, with concrete implementations for ByteArrayBody, InputStreamBody, NoContentBody, and possibly a FileContentBody.  I also suggest removing calls specific to encoding, etc... from this class.  I would like to get rid of null checks on the body parameter and just use a no-op NoContentBody if possible.

Next, there will be some surgical changes to Client.Default (for requests, we need to write a stream instead of a byte[])

There are some questions about calling close() (or not calling close() ) at the end of the response.

Another big impact area is going to be in logger implementations (which need to either not do body logging on streamed bodies, or need to use a different approach that monitors the stream but does not actually read the stream as part of the logging call).

And I think there should be some discussion about supporting InputStream as a body parameter type, and possibly OutputStream as a client method response parameter type.  Maybe these could be handled with specialized encoder/decoder implementations, but it's worth discussion about whether these should be part of the core library.  Might also discuss a File type as a body parameter type (this would allow setting content-length header *and* make the body retryable, which would worthwhile.

Finally, all of my work has been on the synchronous implementations - so evaluation with respect to Async will be needed.

### Conclusion
All in all, while the design discussion will be important, I am confident that it is possible to make this change without massive amounts of surgery.  And I think this would be a huge enhancement to Feign.  This is a huge gap with Jax-RS that would be good to close.

[...]

PR #2750, opened 2025-01-31
Add support for streaming request and response bodies

trumpetinc:
Reference issue: https://github.com/OpenFeign/feign/issues/2734

This is an initial update that changes request and response handling to use streams instead of in-memory byte arrays.  Streams are retriable (up to some amount of reading).

IMPORTANT:  This is a breaking change for Feign encoders/decoders/clients, so the impact could be high on 3rd parties.  These changes should be completely transparent to regular users of Feign.

The general approach is to factor request and response bodies into a single interface (HttpBody) that supports resettable streaming (within some buffered limit if the data source is not inherently seekable).  We also now tracking encoding information in the HttpBody itself, which has dramatically reduced the non-deprecated API footprint.

To demonstrate this streaming functionality, I have added encoders and decoders for handling typical stream-oriented body types (see /core/src/main/java/feign/stream/InputStreamAndFileEncoder.java and /core/src/main/java/feign/stream/InputStreamAndReaderDecoder.java  I suspect that eventually these encoders/decoders will be broken out of the core library.

**Some notes on the sate of this branch's code:**

There are many methods that I have marked as @Deprecated - there is a lot of cruft related to request and response bodies now, and it would be good to clean that out at some point - but this could result in more breaking changes.

All of the Feign core unit tests are working.  I have *not* attempted to propagate changes into the other Feign modules as I want to make sure that the approach and design are accepted first.  There are many areas of the code base that have pending to-do items (//TODO: KD - xxxxxxx) and I expect there be some discussion about those - but for now, I'd like to focus on the overall approach.

At this point, things are "working" - but I would benefit from a discussion about whether the approach is sound, and if it is, I will start a discussion about some of the bigger TODO items that I have tagged.

### DIFF
--- a/annotation-error-decoder/src/main/java/feign/error/ExceptionGenerator.java
+++ b/annotation-error-decoder/src/main/java/feign/error/ExceptionGenerator.java
@@ -17,6 +17,7 @@ package feign.error;
 
 import static feign.Util.checkState;
 
+import feign.HttpBodyFactory;
 import feign.HttpBodyFactory.HttpBody;
 import feign.Request;
 import feign.Response;
@@ -48,7 +49,7 @@ class ExceptionGenerator {
             .headers(testHeaders)
             .request(
                 Request.create(
-                    Request.HttpMethod.GET, "http://test", testHeaders, Request.Body.empty(), null))
+                    Request.HttpMethod.GET, "http://test", testHeaders, HttpBodyFactory.empty(), null))
             .build();
   }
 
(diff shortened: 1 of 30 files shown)


trumpetinc:
Now that I've played with this code for a bit (especially for different encoder implementations), I think that I need to take a different approach.  Request should be writing to an outputstream, not providing an inputstream that we then write to the outputstream.

I will do another iteration on this design.

trumpetinc:
Cancelling this PR - will replace with another approach to solving the problem.

Issue #2680, opened 2024-11-30
Integration of OAuth2 flows in Feign


hosuaby:
Hello @velo 

I am starting working on the integration of **OAuth2 client flows** into Feign. I am opening this issue first.

Reason one: to evaluate the public interest for this feature.

Reason two: as this is not an easy task, I will need to ask for your opinions and suggestions during the development.

Are you agreeing to proceed like that? Or do you prefer to move discussion on Gitter?

velo:
> Reason one: to evaluate the public interest for this feature.

Quite sure the interest to get the feature exists.... not sure if enough interest to contribute

> Reason two: as this is not an easy task, I will need to ask for your opinions and suggestions during the development.

My main ask would be to concentrate the work on a new module dedicated to that.  Changes to core might be needed and welcome (as long you don't break code compatibility and don't add dependencies to core, anything is fair game)

> Are you agreeing to proceed like that? 

I have a very, very, very limited amount of time that I'm willing to spend on this.  If you expect more than a whole hour from me I can't commit to it.  Full time dad first, full time worker second.  Unless you want to hire me to develop tge feature.

hosuaby:
@velo No problem, man. I don't ask you to code it for me. I will take care of development. I just will need to ask you some questions when they arise and validate a few architectural choices to integrate a new module with core as seamlessly as possible.

PR #2755, opened 2025-02-05
MODULE: oauth2

hosuaby:
Integration of OAuth2/OpenID flows into feign.
Following the discution: https://github.com/OpenFeign/feign/issues/2680


hosuaby:
@velo 

Ah, build fails cause I am using Testcontainers. What do you suggest? Do I disable all Tests using Testcontainers or there is a way to get a build environment with Docker on CircleCI?

velo:
Well, I would say to either test using something else. Or figure how to get circle Ci working with test containers or need to move to GitHub actions

hosuaby:
> Well, I would say to either test using something else. Or figure how to get circle Ci working with test containers or need to move to GitHub actions

I am doing experiments with pipeline. Please ignore notifications. When I will come up with a solution, I will inform you and make a separate PR.

core/src/test/java/feign/BaseBuilderTest.java
```diff
@@ -36,9 +38,20 @@ void checkEnrichTouchesAllAsyncBuilderFields()
...
+        Feign.builder()
+            .requestInterceptor(template -> {})
+            .responseInterceptor((ic, c) -> c.next(ic)),
+        12);
+  }
+
   private void test(BaseBuilder<?, ?> builder, int expectedFieldsCount)
       throws IllegalArgumentException, IllegalAccessException {
-    Capability mockingCapability = Mockito.mock(Capability.class, RETURNS_MOCKS);
+    Capability mockingCapability = new MockingCapability();
```
velo: I'm not sure why this change is needed

src/docs/overview-mindmap.iuml
```diff
@@ -41,6 +41,7 @@ left side
 *** Dropwizard Metrics 5
 *** Micrometer
 ** extras
+*** OAuth2
```
velo, line 44: NICE!

velo (approved):
Amazing work, tons of tests, nice improvements, no breaking change, sounds like a win win

kdavisk6:
Circle should be disabled now.  Please update the branch and we'll try again.

hosuaby:
Hello
@velo @kdavisk6 

Don't wanna be pushy, but I would like to know when review will be done? I have already multiple PRs and more will come very soon.

velo (changes requested):
Thanks for the update. I can’t approve this state yet because required CI validation is not in a mergeable/green state for this PR branch. Please rebase/update the branch and ensure required checks are green, then I can re-review.

velo (commented):
Re-checked during an open-PR sweep. Position is unchanged: the module itself looks good and is well covered by tests, but the branch conflicts with master and no CI checks have run against it, so I can't validate it.

Please rebase onto master and push.

One sequencing note: this PR touches `BaseBuilder` and `Capability`, and so does #2804 (interceptors lost on `BaseBuilder#clone`). I plan to merge #2804 first, so please rebase after that lands to keep the conflict resolution to one pass.

velo (changes requested):
This is a substantial, well-organized module and the overall shape (Capability-based wiring, IDP client abstraction) is a reasonable design. A few things need to be addressed before merge, though:

1. Bug: `UnauthorizedRetryer.continueOrPropagate` doesn't return after either of its two guard branches. When `optionalDelegate.continueOrPropagate(...)` returns normally (its contract for "retry, don't propagate"), execution falls through into the reauthentication logic below regardless of status code or the `reauthenticated` flag. Concretely: a non-401 retryable failure will still force a token refresh and rewrite the Authorization header, and the `reauthenticated` guard against repeated forced reauth doesn't actually stop a second reauth once a delegate is configured. Needs an explicit `return` after both branches. No test currently exercises a non-401 status or a repeated-401-with-delegate scenario through this class — worth adding both.

2. `OAuth2Authentication` mixes Capability (a config object meant to be reusable/shared across builders) with mutable session state (`httpClient`, `httpOptions`, `jsonDecoder`, token cache) set as side effects of `enrich(...)`. If the same instance is passed to `.addCapability(...)` on two different `Feign.builder()` calls, the second build silently overwrites state the first client's live interceptor still depends on. Given the public factories (`KeycloakAuthentication.withClientSecretBasic`, etc.) read like reusable config, this is a real footgun. Please either make the object explicitly single-use (throw if `beforeBuild` runs twice) or build an immutable per-target session inside `beforeBuild` instead of mutating shared fields.

3. Relatedly, `beforeBuild()` depends on `enrich(Client)`/`enrich(Options)`/`enrich(Decoder)` having already populated instance fields by the time it runs — an ordering contract enforced only by `BaseBuilder.enrich()`'s current loop structure, not by the type system. Worth considering whether `beforeBuild` should just receive the resolved client/options/decoder as parameters.

4. `OAuth2IDPClient.executeRequest` and `OpenIdProviderClient.discover` duplicate "execute, check status, decode-or-throw" but have diverged: one uses the canonical `FeignException.errorStatus(...)` helper, the other hand-builds a `FeignClientException` with an empty body array and re-reads the body separately. Please consolidate into one shared error-handling path.

5. `OpenIdAuthentication.beforeBuild` calls `super.beforeBuild(...)` (which constructs an `OAuth2IDPClient`) and then immediately overwrites `idpClient` with an `OpenIdProviderClient` — every real subclass (Keycloak/Auth0/Cognito all extend `OpenIdAuthentication`) throws away the superclass's work. A `protected createClient(...)` factory method overridden once would avoid the discarded construction.

6. `ConfidentialLogger` only filters the `Authorization` header on `logRequest`; response logging isn't filtered, so an echoed `Authorization` header in a response (some IDPs, some proxies) would log unredacted through a class whose name implies full coverage. Either filter symmetrically or narrow the documented scope to request-only.

Happy to take another look once these are addressed — #1 is the one I'd consider a hard blocker, the rest are worth fixing before this ships as a public module other users will build on.

PR #2756, opened 2025-02-06
FEATURE: base feign-vertx on Vertx WebClient instead of bare vertx

hosuaby:
I made a number of changes to feign-vertx module:

-  Switch from bare vertx to high level WebClient. It will give the user more configuration options, including the possibility of enabling OAuth2/OIDC authentication (cause feign-vertx and feign-oauth2 are not and cannot be compatible).
- More reliable tests.
- Ensured compatibility with Vertx 5.


vertx/feign-vertx/src/main/java/feign/VertxFeign.java
```diff
@@ -122,13 +120,13 @@ public Builder invocationHandlerFactory(
...
 
     /**
-     * Sets a vertx instance to use to make the client.
+     * Sets a vertx WebClient.
      *
-     * @param vertx vertx instance
+     * @param webClient vertx WebClient
      * @return this builder
      */
-    public Builder vertx(final Vertx vertx) {
```
velo, line 130: My main concern about this change is breaking compatibility with previous version.


Would it be possible to keep this (probably as deprecate) and run `WebClient.create(vertx)`

Then we can instruct to use the new method and mark for removal on next major release

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
```diff
@@ -263,34 +261,6 @@ public Builder errorDecoder(final ErrorDecoder errorDecoder) {
...
-     */
-    @Override
-    public Builder options(final Request.Options options) {
-      checkNotNull(options, "Argument options must be not null");
-      this.options =
-          new HttpClientOptions()
-              .setConnectTimeout(options.connectTimeoutMillis())
-              .setIdleTimeout(options.readTimeoutMillis());
-      return this;
-    }
```
velo, line 292: Same as above, is it possible to keep as deprecated!?

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
```diff
@@ -307,22 +277,22 @@ public Builder timeout(long timeout) {
...
      *     .builder()
      *     .vertx(vertx)
-     *     .requestPreProcessor(req -&#62; req.putHeader("version", "v1"));
+     *     .requestPreProcessor(req -&#62; req.ssl(true));
      * </pre>
      *
      * @param requestPreProcessor operation to execute on each request
      * @return updated request
      */
-    public Builder requestPreProcessor(UnaryOperator<HttpClientRequest> requestPreProcessor) {
```
velo, line 325: the ask here would be the same, not sure if viable at all

vertx/feign-vertx/src/main/java/feign/vertx/VertxHttpClient.java
```diff
@@ -43,28 +44,25 @@
...
+   * @param webClient vertx WebClient
    * @param timeout request timeout
    * @param requestPreProcessor request pre-processor
    */
   public VertxHttpClient(
-      final Vertx vertx,
-      final HttpClientOptions options,
+      final WebClient webClient,
       final long timeout,
-      final UnaryOperator<HttpClientRequest> requestPreProcessor) {
```
velo, line 62: And lastly here too, keep old constructor, deprecate it, add new one

vertx/pom.xml
```diff
@@ -24,16 +24,22 @@
...
+    <module>tests-vertx-common</module>
+    <module>tests-vertx5</module>
+  </modules>
 
   <properties>
-    <vertx.version>4.5.12</vertx.version>
-    <jackson.version>2.18.2</jackson.version>
-    <slf4j-log4j12.version>2.0.0-alpha6</slf4j-log4j12.version>
-    <wiremock.version>2.35.2</wiremock.version>
+    <main.java.version>11</main.java.version>
```
velo, line 38: is this needed?!

11 is the min version already

vertx/pom.xml
```diff
@@ -24,16 +24,22 @@
...
+  <artifactId>feign-vertx-parent</artifactId>
+  <packaging>pom</packaging>
 
-  <name>Feign Vertx</name>
-  <description>Implementation of Feign on Vertx web client.</description>
+  <modules>
+    <module>feign-vertx</module>
+    <module>tests-vertx4</module>
+    <module>tests-vertx-common</module>
+    <module>tests-vertx5</module>
```
velo, line 34: nice effort on test coverage 🥇

velo: I think we can just call them `feign-vertx*-test`

vertx/tests-vertx5/src/test/java/feign/vertx/RetryingTest.java
```diff
@@ -0,0 +1,141 @@
...
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.junit5.VertxTestContext;
+import java.util.Arrays;
+import java.util.Collection;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("When server ask client to retry")
+class RetryingTest extends AbstractFeignVertxTest {
```
velo, line 44: These look to be exactly the same

https://github.com/OpenFeign/feign/pull/2756/files#diff-618b10051163eb35607cd662bc192620411ccce4da0be7a99ee6dfb0b0ee4ad9L50

Why not move to test commons and just have a simplified version here?

velo (changes requested):
I think for tests you could keep them all on feign-vertx, generate the test-jar there, and on feign-vertx5-test you just extend the same tests under feign.vertx5 package....

Then you can make the module a top level project and if eventually vext breaks compatibility we can have multiples like we do with jax-rs

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
hosuaby: No need for that. Because we never released the current version. The previous is still available, by the way, in the separate project that was merged into the main project recently.

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
hosuaby: Idem. Current version from core project was never released.

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
hosuaby: Idem. Current version from core project was never released.

vertx/feign-vertx/src/main/java/feign/vertx/VertxHttpClient.java
hosuaby: Idem. Current version from core project was never released.

vertx/pom.xml
hosuaby: Will check for it. Thanks

vertx/pom.xml
hosuaby: Thanks for the suggestion. I will study them carefully and see how I can improve tests. The most important is to avoid to produce the jar and run tests at same build. By doing so we will not be able to detect runtime binary uncompatibilities with different versions of Vertx (that arise quite offten).

hosuaby: > nice effort on test coverage 🥇

Thanks! feign-vertx is a 10-year-old project. It had plenty of time to be tested in productions and covered with tests for the most of frequent problems.

vertx/feign-vertx/src/main/java/feign/VertxFeign.java
velo: Ow, sweet, perfect time to change then =)

vertx/pom.xml
hosuaby: > I think we can just call them `feign-vertx*-test`

done

hosuaby:
> I think for tests you could keep them all on feign-vertx, generate the test-jar there, and on feign-vertx5-test you just extend the same tests under feign.vertx5 package....
> 
> Then you can make the module a top level project and if eventually vext breaks compatibility we can have multiples like we do with jax-rs

So here for the tests. What I managed and what not to achieve.

Achieved:

1) Moved all tests into the main module with application code feign-vertx.
2) Got rid of the module tests-vertx-common.
3) The main module feign-vertx now produces a test-jar to be imported in test modules feign-vertx4-test and feign-vertx5-test.
4) The module feign-vertx5-test contains no code. It will only be used to test binary compatibility with different versions of Vert.x 5.x.
5) The module feign-vertx4-test contains tests rewritten using the API of Vert.x 4.x. It is used to test binary compatibility with different versions of Vert.x 4.x.

What I could not achieve:

1) You suggested getting rid of the parent module, moving all the production code and tests to the root folder vertx, and producing two jars from it: one with production code and another with tests. By default, this is not possible with Maven and requires a ton of plugins to achieve. Due to architectural complexity, I did not pursue it.
2) I moved common test code used by both modules feign-vertx4-test and feign-vertx5-test into the test jar produced by feign-vertx, but I could not factorize it further. Vert.x 4.x and 5.x have different APIs, so I need to write different code to create a WebClient with the required configuration. I just decided to copy test classes, even if they are similar. Otherwise, it would require a complex test architecture, making them completely unreadable.

@velo  Please tell me what you think and what you would like to improve.

Issue #2782, opened 2025-02-23
BaseBuilder not cloned properly


hosuaby:
@edudar @vitalijr2 

Guys, I have a serious question for ya. I am working on a new `feign-oauth2` module now and having an issue with current implementation of `BaseBuilder`. If I clone the builder using method `Object#clone` and then add an request or response interceptors, those interceptors are simply forgotten. Here an example:

```java
B enrichedBuilder = (B) builder.clone();
return enrichedBuilder
        .requestInterceptor(new AuthenticationInterceptor()); // <-- this interceptor is lost
```

Why does it happens:
1) `BaseBuilder` has the private field `B thisB` pointing to itself.
2) When builder is cloned, the value of that field is copied as is, and continue to point on the object it was cloned from.
3) The references `requestInterceptors` and `responseInterceptors` point to the new `ArrayList` instances, because their values are initialized outside constructor.
4) Method `BaseBuilder#requestInterceptor` adds an interceptor into the list from the new builder object, but returns an old builder then:

```java
  /** Adds a single request interceptor to the builder. */
  public B requestInterceptor(RequestInterceptor requestInterceptor) {
    this.requestInterceptors.add(requestInterceptor);  // <-- interceptor added to new list
    return thisB;    // <-- returns an old builder object without added interceptor
  }
```

So here my questions:

1) @edudar why do we even need to catch value of `this` into instance field `thisB`? I removed it, by returning `this` from every builder method and it seems to works. At least, all tests pass.

2) @vitalijr2 why to use `Object#clone` that is going against all best practices of programming with modern Java? Using a copy constructor would be a more clean and relyable solution.

I can fix both of those problems, but I need a confirmation from you that I didn't miss something.

velo:
Thanks for the report. This project is community-maintained, and the maintainer will not implement fixes directly. If you want this to move forward, please open a pull request with automated tests that cover the change.

For bug reports, please also include (or confirm) a minimal reproduction, the affected version(s), and the full stack trace/logs so contributors can verify the issue.

PR #2783, opened 2025-02-23
REFACTOR: remove helper field BaseBuilder#thisB

hosuaby:
In anticipation of conversation https://github.com/OpenFeign/feign/issues/2782

Remove the field `BaseBuilder#thisB`. This field do not provide any value (appart of avoiding unchecked cast from builder methods), but it mixes badly with `Object#clone` and creates bugs very difficult to spot.


velo (changes requested):
I don't see the value of adding 20+ `@SuppressWarnings`

That was exactly the point of the 'thisB'

Feel free to reopen if you have a different idea without the extra warning suppression.

PR #2804, opened 2025-03-04
FIX: interceptors disapearing on BaseBuilder#clone

hosuaby:
Hello @velo 

I insist on this change that's why I am opening this PR again. This is not a mere style change, but a fix of a very ennoying bug that I have spotted during development of OAuth2 module. I added the test that reproduces the problem. If you revert `BaseBuilder.java` this test will fail.

### DIFF
--- a/core/src/test/java/feign/BaseBuilderTest.java
+++ b/core/src/test/java/feign/BaseBuilderTest.java
@@ -67,4 +67,93 @@ class BaseBuilderTest {
             .responseInterceptor((ic, c) -> c.next(ic)),
         12);
   }
+
+  @Test
+  void checkCloneDontLooseInterceptors() {
+    Feign.Builder originalBuilder =
+        Feign.builder()
+            .requestInterceptor(new FirstRequestInterceptor())
+            .addCapability(
+                new Capability() {
+                  @Override
+                  public RequestInterceptor enrich(RequestInterceptor requestInterceptor) {
+                    return new DecoratingRequestInterceptor(requestInterceptor);
+                  }
+                });
+
+    // There is one interceptor FirstRequestInterceptor
+    assertThat(originalBuilder.requestInterceptors)
+        .isNotNull()
+        .isNotEmpty()
+        .hasSize(1)
+        .first()
+        .isInstanceOf(FirstRequestInterceptor.class);
+
+    Feign.Builder enrichedBuilder = originalBuilder.enrich();
+
+    // Original builder should have one interceptor FirstRequestInterceptor
+    assertThat(originalBuilder.requestInterceptors)
+        .isNotNull()
+        .isNotEmpty()
+        .hasSize(1)
+        .first()
+        .isInstanceOf(FirstRequestInterceptor.class);
+
+    // enrichedBuilder should have one interceptor DecoratingRequestInterceptor
+    assertThat(enrichedBuilder.requestInterceptors)
+        .isNotNull()
+        .isNotEmpty()
+        .hasSize(1)
+        .first()
+        .isInstanceOf(DecoratingRequestInterceptor.class);
+
+    Feign.Builder enrichedBuilderWithInterceptor =
+        enrichedBuilder.requestInterceptor(new SecondRequestInterceptor());
+
+    // Original builder should have one interceptor FirstRequestInterceptor
+    assertThat(originalBuilder.requestInterceptors)
+        .isNotNull()
+        .isNotEmpty()
+        .hasSize(1)
+        .first()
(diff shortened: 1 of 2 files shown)


core/src/test/java/feign/BaseBuilderTest.java
```diff
@@ -67,4 +67,27 @@ void checkEnrichTouchesAllBuilderFields()
...
+
+  @Test
+  void checkCloneDontLooseInterceptors() {
+    Feign.Builder originalBuilder = Feign.builder().addCapability(new Capability() {});
+
+    // There is no any interceptors
+    assertThat(originalBuilder.requestInterceptors).isNotNull().isEmpty();
+
+    Feign.Builder enrichedBuilder = originalBuilder.enrich();
+    Feign.Builder enrichedBuilderWithInterceptor = enrichedBuilder.requestInterceptor((req) -> {});
```
velo, line 79: Well, by design `enrich` is meant to be the last thing you invoke.

After you set all your interceptors, encoders, decodes, you `build`, and build calls `enrich` witch will apply modifications to client, encoder and other components.

Doing this you are messing around with the Capability process.


If you wanna see my point, change you Capability on line 73 to do something with the `requestInterceptor`.

You will see that the capability will be ignored making very hard to anyone to understand why the capability is not working.

Now I don't think we can't have this change

velo (changes requested):

core/src/test/java/feign/BaseBuilderTest.java
hosuaby: Yes, @velo 

I know it. All interceptors added after `enrich` was invoked will be not processed by `Capability`. And this is exactly what my module `oauth2` does: https://github.com/hosuaby/feign/blob/feature/oauth2/oauth2/src/main/java/feign/auth/oauth2/OAuth2Authentication.java#L80

This `OAuth2Authentication` is a `Capability` that adds interceptor, retryer and errorDecoded just before `Feign` is built. So they are ignored by any other capability.

Even if you approved my PR already, please follow the link, and check once again. Do you agree with this design?

velo (approved):
Thanks! Clone/interceptor behavior fix is covered by tests, and I do not see compatibility or security concerns.

velo (commented):
Re-reviewed as part of an open-PR sweep. The fix is correct and the test is a good one — capturing `thisB` at construction time means it survives `Object#clone()` still pointing at the original builder, so every fluent call on the enriched clone mutates the original. Replacing it with `(B) this` at each call site is the right fix, and dropping the `thisB` filter from `getFieldsToEnrich` follows naturally.

`checkCloneDontLooseInterceptors` asserts exactly the right thing: that the original and the enriched builder diverge, in both directions.

The only blocker is that the branch now conflicts with master and CI has never run on it. Could you rebase onto master and push? Once checks are green I'll merge.

One heads-up: #2755 also touches `BaseBuilder`, so there will be a conflict between the two. I intend to land this one first.

hosuaby:
@velo Rebase is done

velo (changes requested):
I reproduced this locally rather than just reading the diff, and the fix doesn't actually work: applying the full PR diff on top of its own merge-base commit still fails with the same UnsupportedOperationException the new test (checkCloneDontLooseInterceptors) is meant to catch. CI confirms this too (ci/circleci: pr-build is currently failing on this PR).

The real bug is in `enrich()`: it reassigns clone.requestInterceptors (and the response/method equivalents) to RequestInterceptors#interceptors(), which returns Collections.unmodifiableList(...). Any later .requestInterceptor(...) call on the enriched builder throws. That wrapper predates this PR by years (#1626) and isn't touched here — thisB() was always just `return (B) this;`, with no state that could lose interceptors, so removing it can't be the fix (and isn't, in my repro).

On top of not fixing the bug, deleting thisB() is a straight quality regression on its own: it replaces one centrally-annotated helper (~28 call sites) with 20 scattered `(B) this` casts, each needing its own @SuppressWarnings("unchecked"). That's more surface to scan and maintain for zero behavioral gain.

The PR is also stale relative to master (mergeStateStatus DIRTY / CONFLICTING) — master's BaseBuilder.java has since grown with encoder/decoder and getFieldsToEnrich() machinery this PR predates.

Suggested path: keep thisB() as-is, and fix enrich() to wrap the RequestInterceptors/ResponseInterceptors/MethodInterceptors output in a fresh mutable ArrayList before assigning it to the clone's field (or otherwise give enriched builders a genuinely mutable interceptor list). Happy to take another look once the test actually passes against the real fix.

PR #2871, opened 2025-04-27
Add constructor to RetryableException for cases without retryAfter

### LINKED ISSUE #2458: Backward compatibility issue of overload resolution ambiguity
issue starts with https://github.com/OpenFeign/feign/pull/2170

Many code snippets are used this way.
```kotlin
RetryableException(
    response.status(),
    RATE_LIMIT_EXCEEDED,
    response.request().httpMethod(),
    null, // Date type (original)
    response.request(),
)
```

However, this code will not work with that change.

1. get a compile time error

```bash
Overload resolution ambiguity: 
public constructor RetryableException(p0: Int, p1: String!, p2: Request.HttpMethod!, p3: Date!, p4: Request!) defined in feign.RetryableException
public constructor RetryableException(p0: Int, p1: String!, p2: Request.HttpMethod!, p3: Long!, p4: Request!) defined in feign.RetryableException
```

2. ambiguous what to do when we don't want to use retry.
- Just looking at the code in the IDE didn't work, so I came to repository.

# suggest two things

1. fundamentally fix the overload resolution ambiguity.
  - need to ideation about how to fix it.
2. annotate how it should be written in the new constructor when don't want to retry.
  - What about adding it to a new constructor?

rojae:
### Add. two new constructors to RetryableException:
- A constructor without retryAfter and without cause
- A constructor without retryAfter but with cause

Standardized JavaDoc across all RetryableException constructors

These new constructors remove overload ambiguity when retryAfter is not needed.

### Add. a unit test to verify the new constructors.
- canCreateRetryableExceptionWithoutRetryAfter
- canCreateRetryableExceptionWithoutRetryAfterAndWithCause

Fixes #2458

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -24,7 +24,7 @@
     <extension>
         <groupId>com.gradle</groupId>
         <artifactId>common-custom-user-data-maven-extension</artifactId>
-        <version>2.0.1</version>
+        <version>2.0.2</version>
     </extension>
     <extension>
         <groupId>com.marvinformatics.jacoco</groupId>
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -32,7 +32,7 @@
     <jmh.version>1.37</jmh.version>
     <rx.netty.version>0.5.3</rx.netty.version>
     <rx.java.version>1.3.8</rx.java.version>
-    <netty.version>4.2.0.Final</netty.version>
+    <netty.version>4.2.1.Final</netty.version>
     <moditect.skip>true</moditect.skip>
   </properties>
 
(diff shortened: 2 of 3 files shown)


rojae:
Force-pushed to simplify the commit history and correct unintended changes caused by formatting.

velo (approved):

rojae:
Hi @velo 

Thanks approved.

But build is failed due to the `git-code-format-maven-plugin`.
➡️ [Build failure log](https://github.com/OpenFeign/feign/actions/runs/14693600504/job/41652590773#step:9:791)

So I ran the formatter locally using:
```bash
mvn git-code-format:format-code
```

Then I force-pushed the fix to simplify the commit history.
➡️ [Commit](https://github.com/OpenFeign/feign/pull/2871/commits/9c3bedf7e7938f2369f84cc302f06ebb24a8bb8d)

Apologies for missing that - thank you!

PR #2874, opened 2025-04-30
Add. hint message for missing @Param value

### LINKED ISSUE #2337: Add documentation/exception for required java compiler flag "parameters"
When using OpenFeign in the following way (no name attribute specified):

```java
@RequestLine("GET /getSomething?id={id}")
String getSomething(@Param String id);
```
You may encounter the following exception:

`... annotation was empty on param 0`

This happens because `@Param long catalogId` is compiled as `@Param long arg0`. OpenFeign will then fail to read the parameter name via reflection.

To fix this, you need to pass the `parameters` flag to the Java compiler:

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <configuration>
        <parameters>true</parameters>
    </configuration>
</plugin>
```

After hours of research and debugging I have found this solution. Just now I found a little hint in your documentation: `(if the code was compiled with -parameters flag)`. I would like to see that more visible or even in the exception..

rojae:
- Added a helpful hint in the exception message when @Param is missing a value and parameter name can't be resolved.
- Added a unit test to cover this case.

Files affected:
- Contract.java
- DefaultContractTest.java

Fixes #2337


velo (approved):
Excellent idea

rojae:
Hi @velo 

Since the **parent POM** uses the `-parameters` compiler flag during tests,  
the test I added was failing because parameter names are available at runtime.  
➡️ [Build failure log](https://github.com/OpenFeign/feign/actions/runs/14756611808/job/41652521622?pr=2874#step:9:453)

[Parent POM snippet]
```xml
<compilerArgs>
  <arg>-parameters</arg>
</compilerArgs>
```

I've updated the test to skip execution when this flag is present, which should resolve the build failure.
➡️ [Comparison with fix](https://github.com/OpenFeign/feign/compare/087c942cfd17c1b3a06af543503939f36de4d75d..cefb8c4daab5ee537aa6247eb6bc91a8a7e67e8e)

I missed this earlier because I only ran tests in the core module locally.
Let me know if anything else is needed — thank you!

Issue #2936, opened 2025-07-04
CI: build pipeline is failing


hicom150:
It seems that the CI build pipeline is failing in `Wikipedia Example with Spring Boot` module due to a profile that only activates on Java 17 [here](https://github.com/OpenFeign/feign/blob/master/example-wikipedia-with-springboot/pom.xml#L76) but the mail build pipeline uses Java 21 as default [here](https://github.com/OpenFeign/feign/blob/master/.github/workflows/build.yml#L71)

Does it make sense to remove this profile activation?

If it is OK, I can make a PR to remote it a fix the build pipeline 👍

Thank you!

velo:
Thanks for the report. This project is community-maintained, and the maintainer will not implement fixes or features directly. If you want this to move forward, please open a pull request with automated tests that cover the change.

PR #2937, opened 2025-07-05
Remove unnecessary profile activation

hicom150:
Removes unnecessary profile activation in `feign-example-wikipedia-with-springboot` module.
This should fix CI build pipeline as explained in #2936


hicom150:
@velo could be possible to give an approval to run the CI workflow so we can check that this really fixes the build pipeline? Thank you! 🙏

velo (approved):

velo (changes requested):
Thanks for the update. I can’t approve this state yet because required CI validation is not in a mergeable/green state for this PR branch. Please rebase/update the branch and ensure required checks are green, then I can re-review.

velo (commented):
Re-checked during an open-PR sweep — status is unchanged since my last review:

- `ci/circleci: pr-build` is failing
- the branch conflicts with master

Please rebase onto master and push so CI can run clean. Happy to re-review as soon as the checks are green.

velo (changes requested):
The intent here is right — the JDK-17 profile-activation gate was never a reliable signal for this module (toolchain selection is per-module via maven-toolchains-plugin, not driven by the host JVM invoking Maven), so deleting the conditional profile wrapper instead of patching its activation range is the correct move.

However, this branch is badly out of sync with master (last synced 2025-10-20) and merging it as-is would silently revert two unrelated fixes that have landed on this exact file since:
- #3088 (2025-10-21): extracted hardcoded plugin/dependency versions in this pom.xml into parent-POM properties
- the commit adding an explicit feign-form-spring dependency for parallel reactor build ordering, with an explanatory comment

Your diff still shows the old hardcoded versions and doesn't have the feign-form-spring dependency, so merging now would reintroduce both regressions.

Could you rebase onto current master and re-verify the profile removal is still correct against the current CI setup? Note the issue this PR cites (#2936) blames the old GitHub Actions workflow, which has since been removed — CI now runs on CircleCI with maven-toolchains-plugin, so the justification is worth re-confirming against .circleci/config.yml rather than the deleted workflow. Happy to merge a rebased version.

velo:
Apologies for the long silence on this one — you asked for a CI approval back in July 2025 and never got an answer.

I'm going to close it, and the reason isn't the idea. The branch has been conflicting for a while, and its copy of `example-wikipedia-with-springboot/pom.xml` still carries hardcoded plugin versions and is missing the `feign-form-spring` dependency. Merging it as-is would revert #3088 along with the reactor-ordering fix — master has since moved to `${springboot.version}` and `${really-executable-jar-maven-plugin.version}`.

The underlying change — unwrapping that unnecessary profile activation — is still worth doing, and it's about ten lines in one example pom. Rather than ask you to rebase 14 months of drift for that, I'll redo it directly on master.

Thanks for raising it, and sorry it sat this long.

PR #3126, opened 2025-11-19
build(deps): update Vert.x to 5.0.5 in vertx5-test module

velo:
Updates Vert.x from 5.0.4 to 5.0.5 in the feign-vertx5-test module only.

This PR correctly targets only the vertx5-test module which is specifically for testing with Vert.x 5.x. The vertx4-test module remains on Vert.x 4.x as intended.

Closes #3090 (Dependabot's PR tried to update both modules which was incompatible)


velo:
Superseded - vertx already at 5.0.7 on master

PR #3172, opened 2026-01-14
Fix Slf4jLogger and JavaLogger returning unbuffered response when logging disabled

### LINKED ISSUE #1336: Slf4jLogger returns different response depending on logging level
`
package feign.slf4j;

public class Slf4jLogger extends feign.Logger {
...

  @Override
  protected Response logAndRebufferResponse(String configKey,
                                            Level logLevel,
                                            Response response,
                                            long elapsedTime)
      throws IOException {
    if (logger.isDebugEnabled()) {
      return super.logAndRebufferResponse(configKey, logLevel, response, elapsedTime);
    }
    return response;
  }
...
}
`

It returns different response object depending on logging level.
When setting logging level above INFO, it returns just **response**, and it's not same as **logAndRebufferResponse**'s return value.
So when I try to access `response.body().toString()` in ErrorDecoder, I cannot get proper response value.
What I expected was json string value but I got `feign.okhttp.OkHttpClient$1@blabla`
It works fine with DEBUG/TRACE logging level.

Is it issue or am I missing something?

loadkrnis:
@velo 

### Problem

`Slf4jLogger` and `JavaLogger` currently behave differently depending on the logging level.

When the SLF4J level is set to `INFO` or higher (or `java.util.logging` is above `FINE`),  
`logAndRebufferResponse()` does not delegate to `super.logAndRebufferResponse()`.  
As a result, the response body is not rebuffered and remains a one-time `InputStream`.

This leads to several practical issues:

- `ErrorDecoder` implementations cannot reliably read the response body from error responses
- Calling `response.body().toString()` returns an instance reference (e.g. `feign@...`) instead of the actual content
- Application behavior varies across environments with different logging configurations

---

### Solution

Always delegate to `super.logAndRebufferResponse()`, regardless of the logger’s runtime level.

This ensures that response buffering follows Feign’s `Logger.Level` consistently and is no longer affected by SLF4J or JUL configuration differences.

---

### Result

- Response bodies are always rebuffered when `Logger.Level >= HEADERS`, regardless of SLF4J or JUL logging levels
- Consistent behavior is guaranteed across environments
- `ErrorDecoder` can reliably read response bodies

---

### Changes

- **`Slf4jLogger`**: Removed conditional logic in `logAndRebufferResponse()` and always delegate to `super`

---

### Behavioral Impact

**Memory usage**

- Response bodies are now buffered in memory when `Logger.Level` is `HEADERS` or `FULL`, even if the underlying SLF4J / JUL level is `INFO` or higher

---

All existing tests pass.

I’m happy to adjust this approach if there are alternative suggestions or concerns from the maintainers.

---

### Alternative Approach Considered

**Option: Lazy buffering with `LazyBufferedBody`**

As an alternative to eager buffering, we also considered introducing a `LazyBufferedBody` wrapper that would:

- Wrap non-repeatable streams when `Logger.Level >= HEADERS`
- Buffer the body into memory only on first access (for example, by an `ErrorDecoder`)
- Avoid allocating memory when the response body is never read

This approach could reduce memory usage in some cases, especially when response bodies are rarely accessed.

The current implementation is not meant to assert that eager rebuffering is the only or best solution.  
Rather, it serves as a concrete proposal to make the existing behavior consistent.

If maintainers prefer a different approach—such as lazy buffering—I’m happy to adjust the implementation based on that feedback.

Fixes #1336

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -230,10 +230,7 @@ public abstract class Logger {
     @Override
     protected Response logAndRebufferResponse(
         String configKey, Level logLevel, Response response, long elapsedTime) throws IOException {
-      if (logger.isLoggable(java.util.logging.Level.FINE)) {
-        return super.logAndRebufferResponse(configKey, logLevel, response, elapsedTime);
-      }
-      return response;
+      return super.logAndRebufferResponse(configKey, logLevel, response, elapsedTime);
     }
 
     @Override
--- a/core/src/test/java/feign/JavaLoggerTest.java
+++ b/core/src/test/java/feign/JavaLoggerTest.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright © 2012 The Feign Authors (feign@commonhaus.dev)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feign;
+
+import feign.Logger.JavaLogger;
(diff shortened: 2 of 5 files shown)


velo (approved):
LGTM — fix is correct: buffering should follow feign's Logger.Level, not the SLF4J/JUL backend's runtime level. Output suppression is still preserved via the guard in Slf4jLogger.log(). Merging.

PR #3237, opened 2026-02-23
Extract inner Default classes to standalone top-level classes

velo:
## Summary

- Extracts 9 inner `Default` implementation classes from their parent interfaces into standalone top-level classes (`DefaultClient`, `DefaultAsyncClient`, `DefaultContract`, `DefaultRetryer`, `DefaultEncoder`, `DefaultDecoder`, `DefaultErrorDecoder`, `DefaultInvocationHandlerFactory`, `DefaultQueryMapEncoder`)
- Deprecates the original inner `Default` classes, reducing them to thin subclasses that delegate to the new standalone classes
- Updates all references (~54 files across 16+ modules) to use the new standalone classes

## Test plan

- [x] `mvn test` passes across all modules
- [x] Deprecated inner classes still compile and work as thin wrappers for backward compatibility
- [x] `Client.Proxied` correctly extends `DefaultClient` instead of the deprecated `Client.Default`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


norrisjeremy:
What was the motivation for this PR?
And why isn't this change documented at all in the release notes?

velo:
did it broke something?

velo:
> And why isn't this change documented at all in the release notes?

`¯_(ツ)_/¯`
git hub ignored the PR when release notes are auto-generated.  Feel free to review released and suggest fix.

norrisjeremy:
You deprecated 9 classes with no explanation of what the benefit was.
From a user's perspective, you simply caused a bunch of new deprecation warnings in their codebase for no perceivable benefit.

PR #3308, opened 2026-04-12
fix: remove static ExecutorService in AsyncFeign to prevent ClassLoader leak

### LINKED ISSUE #3178: [Bug] Static ExecutorService in AsyncFeign causes ClassLoader Leak (Metaspace OOM)
Description
The AsyncFeign class utilizes a private static inner class (LazyInitializedExecutorService) to hold a singleton ExecutorService. There is no mechanism provided to shut down this thread pool.

Root Cause
1. Static Singleton: The executor is defined as private static final in LazyInitializedExecutorService (Lines 58-67).

2. No Termination: The AsyncFeign class does not expose any shutdown() method or hook to clean up this resource.

3. Container Context: While the threads are set as daemon, this only allows the JVM to exit. In a Servlet Container (like Tomcat), redeploying the application does not restart the JVM. The static thread pool retains strong references to the ContextClassLoader of the initial application, preventing it from being garbage collected.

Impact
·ClassLoader Leak: Each redeployment leaks the entire ClassLoader of the previous application version.

·Thread Leak: The number of active threads grows linearly with each redeployment.

·OOM: Eventually leads to Metaspace OutOfMemoryError.

Relevant Code
feign/AsyncFeign.java

private static class LazyInitializedExecutorService {
    private static final ExecutorService instance =
        Executors.newCachedThreadPool(
            r -> {
              final Thread result = new Thread(r);
              result.setDaemon(true); // Daemon is insufficient for container redeployments
              return result;
            });
}

Suggested Fix
Avoid using a static singleton for the ExecutorService. Instead, manage the executor as an instance field within AsyncFeign, allowing it to be garbage collected when the client instance is discarded, or allow the user to provide their own ExecutorService via the builder.

anuragg-saxenaa:
## Problem

`AsyncFeign` held a static singleton `ExecutorService` via `LazyInitializedExecutorService`:

```java
private static class LazyInitializedExecutorService {
  private static final ExecutorService instance =
      Executors.newCachedThreadPool(r -> {
        final Thread result = new Thread(r);
        result.setDaemon(true);
        return result;
      });
}
```

In servlet containers (Tomcat, Jetty, etc.), redeploying an application does not restart the JVM. The static thread pool holds a strong reference to the `ContextClassLoader` of the first deployment. On each redeployment:
- The old `ClassLoader` is never GC'd (retained by the static thread pool's threads)
- Thread count grows linearly
- Metaspace eventually exhausts → `OutOfMemoryError`

Setting threads as daemon only helps at JVM exit — it does not prevent ClassLoader leaks during hot-redeploy.

## Fix

Remove `LazyInitializedExecutorService` entirely. `AsyncBuilder.client` is now initialized with a fresh `Executors.newCachedThreadPool(...)` per builder instance. This means:
- Each `AsyncFeign.builder()` creates its own independent executor
- When the client is discarded, the executor becomes eligible for GC
- Users who need lifecycle control or sharing can supply their own via `.client(AsyncClient)`

## Testing

Added `asyncFeignHasNoStaticExecutorService()` to `AsyncFeignTest` — uses reflection to assert that no `static` field of type `ExecutorService` exists in any inner class of `AsyncFeign`.

Fixes #3178

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -55,23 +55,18 @@ public final class AsyncFeign<C> {
     return builder();
   }
 
-  private static class LazyInitializedExecutorService {
-
-    private static final ExecutorService instance =
-        Executors.newCachedThreadPool(
-            r -> {
-              final Thread result = new Thread(r);
-              result.setDaemon(true);
-              return result;
-            });
-  }
-
   public static class AsyncBuilder<C> extends BaseBuilder<AsyncBuilder<C>, AsyncFeign<C>> {
 
     private AsyncContextSupplier<C> defaultContextSupplier = () -> null;
     private AsyncClient<C> client =
         new DefaultAsyncClient<>(
-            new DefaultClient(null, null), LazyInitializedExecutorService.instance);
+            new DefaultClient(null, null),
+            Executors.newCachedThreadPool(
+                r -> {
+                  final Thread result = new Thread(r);
+                  result.setDaemon(true);
+                  return result;
+                }));
     private MethodInfoResolver methodInfoResolver = MethodInfo::new;
 
     @Deprecated
--- a/core/src/test/java/feign/AsyncFeignTest.java
+++ b/core/src/test/java/feign/AsyncFeignTest.java
@@ -1377,4 +1377,23 @@ public class AsyncFeignTest {
   void afterEachTest() throws IOException {
     server.close();
   }
+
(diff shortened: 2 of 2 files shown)


yvasyliev:
isn't it already covered in #3277?

anuragg-saxenaa:
Closing as duplicate of #3277 which addresses the same issue with a more complete fix (also covering CoroutineFeign). Thanks for the pointer @yvasyliev!

Issue #1487, opened 2021-08-17
ErrorDecoder not invoked when using Apache Http client


tszlek:
Hi, I have discovered issue when using `feign.httpclient.ApacheHttpClient`.
In some cases Apache client throws `ProtocolException` which is a checked exception. This happens for example when server returned 303 HTTP code but there was no location header.
In such case Feign proxy throws `UndeclaredThrowableException` instead of invoking ErrorDecoder.

Vaibhav-tech22:
Hey @velo, is this issue closed. If not I would like to give it a try

velo:
Still needs work.

Probably some tests showing the issue and handling the exception at Client code

young0264:
@velo 
Hi, I'd like to give this issue a try and follow up on the previously suggested approaches.  
I'll check the exception handling inside ApacheHttpClient and evaluate how to pass root cause info to ErrorDecoder.  
Let me know if there's any preference for the direction.

gitshabh:
Hello @velo , I've created a PR for this. Can you please have a look at it and review?
https://github.com/OpenFeign/feign/pull/3338

young0264:
@velo Following up on my earlier comment here: https://github.com/OpenFeign/feign/issues/1487#issuecomment-2875074758

I traced the exception flow further and ended up addressing this in the core UNWRAP propagation path rather than in ApacheHttpClient itself.

I opened a PR here: https://github.com/OpenFeign/feign/pull/3359

The change keeps undeclared checked causes wrapped as RetryableException, still unwraps declared checked causes, and adds coverage for declared, undeclared, assignable, and Apache redirect-without-Location cases.

PR #3338, opened 2026-05-01
[Issue 1487] Fix ErrorDecoder not invoked for Apache HttpClient checked exceptions

gitshabh:
Problem

When Apache HttpClient throws checked exceptions like ProtocolException (e.g., missing Location header on 303 redirect), the ErrorDecoder was never invoked because the exception occurred during [client.execute()](vscode-file://vscode-app/c:/Users/Rishabh/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) before a Response object was created.

Root Cause

The ErrorDecoder interface requires a Response object: [Exception decode(String methodKey, Response response)]. When execution fails with checked exceptions, there's no Response to pass, causing the exception to propagate uncaught and potentially get wrapped in UndeclaredThrowableException.

Solution:

Build Response with appropriate status codes in case of ProtocolException instead of throwing exception so that error decoder can be invoked.

### DIFF
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -23,6 +23,8 @@ import feign.Feign;
 import feign.Feign.Builder;
 import feign.FeignException;
 import feign.Request.Options;
+import feign.RetryableException;
+import feign.Retryer;
 import feign.client.AbstractClientTest;
 import feign.jaxrs.JAXRSContract;
 import java.nio.charset.StandardCharsets;
@@ -87,6 +89,21 @@ public class ApacheHttpClientTest extends AbstractClientTest {
         assertThrows(FeignException.class, () -> testInterface.withOptions(options));
     assertThat(feignException.status()).isEqualTo(302);
     assertThat(server.takeRequest().getPath()).isEqualTo("/withOptions");
+  }
+
+  @Test
+  void redirectWithoutLocationHeader() {
+    JaxRsTestInterface api =
+        Feign.builder()
+            .contract(new JAXRSContract())
+            .retryer(Retryer.NEVER_RETRY)
+            .client(new ApacheHttpClient(HttpClientBuilder.create().build()))
+            .target(JaxRsTestInterface.class, "http://localhost:" + server.getPort());
+
+    server.enqueue(new MockResponse().setResponseCode(303));
+    RetryableException exception =
+        assertThrows(RetryableException.class, () -> api.withoutBody("foo"));
+    assertThat(exception.getMessage()).contains("org.apache.http.client.ClientProtocolException");
   }
 
   private JaxRsTestInterface buildTestInterface() {
--- a/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
+++ b/httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
@@ -34,6 +34,8 @@ import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.RecordedRequest;
+import org.apache.http.ProtocolException;
+import org.apache.http.client.ClientProtocolException;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.junit.jupiter.api.Test;
 
@@ -43,6 +45,25 @@ public class ApacheHttpClientTest extends AbstractClientTest {
   @Override
   public Builder newBuilder() {
     return Feign.builder().client(new ApacheHttpClient());
+  }
+
+  @Test
+  void redirectWithoutLocationHeaderKeepsRetryableExceptionWhenPropagationPolicyIsUnwrap() {
+    JaxRsTestInterface api =
+        Feign.builder()
+            .contract(new JAXRSContract())
+            .exceptionPropagationPolicy(feign.ExceptionPropagationPolicy.UNWRAP)
+            .retryer(Retryer.NEVER_RETRY)
+            .client(new ApacheHttpClient(HttpClientBuilder.create().build()))
+            .target(JaxRsTestInterface.class, "http://localhost:" + server.getPort());
+
+    server.enqueue(new MockResponse().setResponseCode(303));
+
+    RetryableException exception =
+        assertThrows(RetryableException.class, () -> api.withoutBody("foo"));
+
+    assertThat(exception.getCause()).isInstanceOf(ClientProtocolException.class);
+    assertThat(exception.getCause()).hasCauseInstanceOf(ProtocolException.class);
   }
 
   @Test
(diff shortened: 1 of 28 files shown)


gitshabh:
@velo
Hi, we can see the "Redirection but no location header" in the FeignException instead of UndeclaredThrowableException. Do you think this solution is correct?

velo (changes requested):
Thanks for picking this up, but I don't think the synthetic-response approach is the right direction. Concerns:

- **Misleading status codes**: `300` for "missing Location header" and `500` for arbitrary checked exceptions misrepresent what actually happened. ErrorDecoders keying off status will misclassify, and these synthetic codes can collide with real server responses the user already handles.
- **Silent retry loss**: callers previously got `RetryableException` (going through `Retryer`); now they get a non-retryable `FeignException`. That is a backwards-incompatible behavior change.
- **Brittle detection**: `cause.getClass().getSimpleName().equals("ProtocolException")` should be `instanceof org.apache.http.ProtocolException`. String comparison also collides with `java.net.ProtocolException`.
- **Lossy**: original cause is reduced to body bytes via `getMessage().getBytes()` (platform default charset). ErrorDecoders can't tell what actually went wrong.
- **Catches too broadly**: `!IOException && !RuntimeException && !Error` swallows *any* checked-exception cause into a fake 500, far wider than "ProtocolException on redirect".
- **`reason` field misuse**: that field is the HTTP reason phrase, not an exception class name.
- **Cross-client inconsistency**: only HC4 gets this treatment; hc5/OkHttp/JDK/Java11 still behave differently for the same scenario.
- **Test coverage**: one test, only the redirect path; the generic 500 branch is untested.

Posting better directions in a separate comment.

velo:
@gitshabh thanks for the work — to your question, I don't think this is the right approach. Some better directions:

**1. Minimal fix at the boundary.** Ensure non-`IOException` causes get wrapped in `IOException` so the existing `errorExecuting` path in `SynchronousMethodHandler` (`core/src/main/java/feign/SynchronousMethodHandler.java:115`) handles them. This addresses the original `UndeclaredThrowableException` complaint without inventing fake responses or changing retry semantics:

```java
try {
  HttpResponse httpResponse = client.execute(httpUriRequest);
  return toFeignResponse(httpResponse, request);
} catch (IOException e) {
  throw e;
} catch (Exception e) {
  throw new IOException(e);
}
```

That said — `client.execute()` already declares `throws IOException`, and Apache wraps `ProtocolException` in `ClientProtocolException` (which extends `IOException`). For the reported case the existing path *should* already fire and produce a `RetryableException`. Worth confirming the actual observed exception chain before patching, because the original `UndeclaredThrowableException` symptom may have a different root cause.

**2. If you genuinely want `ErrorDecoder` to participate in transport-level failures**, that's a `Client`-interface design discussion, not a per-client patch. The 2021 commits linked from #1487 explored extending `ErrorDecoder` to receive the original exception — that's a more honest model than fabricating an HTTP response. It should be applied consistently across all client modules (OkHttp/hc5/JDK/Java11), otherwise you create per-client behavior divergence.

**3. Where the change belongs.** If anything is centralized, it belongs in core (e.g., `SynchronousMethodHandler` broadening its catch and wrapping non-`IOException` causes), not duplicated in every `Client` implementation. Replicating the synthetic-response trick across hc5/OkHttp/JDK/Java11 would multiply the misleading-status-code and retry-loss problems across modules.

The fundamental issue with the current PR: synthetic HTTP statuses (`300`/`500`) are a lie about what happened, and they silently disable retries. Even just for HC4, those footguns outweigh the nicer error message.

gitshabh:
@velo thanks for the detailed clarification. I agree that the synthetic-response approach is a bad decision. It can lead to loss of retries and potential ambiguity.

I have another solution in mind. We can stop unwrapping undeclared checked exceptions to avoid UndeclaredThrowableException by keeping them wrapped in RetryableException.

* Feign wraps client exceptions (e.g., IOException from HTTP clients) in RetryableException for retry logic.
* When retries fail, the UNWRAP policy unwraps the original exception to avoid hiding the root cause.

```java
try {
  retryer.continueOrPropagate(e);
} catch (RetryableException th) {
  Throwable cause = th.getCause();
  if (methodHandlerConfiguration.getPropagationPolicy() == UNWRAP && cause != null) {
    throw cause;
  } else {
    throw th;
  }
}
``` 

However when the cause is not declared by the target method's throws clause, throwing it results in UndeclaredThrowableException.

**Solution:**
Don't unwrap undeclared checked exceptions. We throw RetryableException which contains high level information of the actual cause which is better than having an UndeclaredThrowableException.

Updated Unwrapping Condition:
```java
if (methodHandlerConfiguration.getPropagationPolicy() == UNWRAP && cause != null 
    && (cause instanceof RuntimeException || cause instanceof Error 
        || isDeclaredCheckedException(cause, methodHandlerConfiguration.getMetadata().method()))) {
  throw cause;
} else {
  throw th;  // Keep wrapped
}
```
* Unwraps unchecked exceptions (as before).
* Additionally unwraps checked exceptions if they match the interface method's throws clause.

Helper Method:
```java
private static boolean isDeclaredCheckedException(Throwable cause, Method method) {
  Class<?>[] exceptionTypes = method.getExceptionTypes();
  for (Class<?> exType : exceptionTypes) {
    if (exType.isAssignableFrom(cause.getClass())) {
      return true;
    }
  }
  return false;
}
```
* Checks if the exception type (or a subclass) is declared in the method's throws clause.
* Uses isAssignableFrom to handle inheritance (e.g., IOException covers ClientProtocolException).

Do you think this approach can work for us? I've updated the PR according to it. Please have a look at it if possible.

httpclient/src/test/java/feign/httpclient/ApacheHttpClientTest.java
```diff
@@ -89,6 +91,21 @@ void notFollowRedirectIsRespected() throws InterruptedException {
...
+        Feign.builder()
+            .contract(new JAXRSContract())
+            .retryer(Retryer.NEVER_RETRY)
+            .client(new ApacheHttpClient(HttpClientBuilder.create().build()))
+            .target(JaxRsTestInterface.class, "http://localhost:" + server.getPort());
+
+    server.enqueue(new MockResponse().setResponseCode(303));
+    RetryableException exception =
+        assertThrows(RetryableException.class, () -> api.withoutBody("foo"));
+    assertThat(exception.getMessage()).contains("org.apache.http.client.ClientProtocolException");
```
gitshabh: Prior to the change in ApacheHttpClient the exception message was
"null executing PUT http://localhost:59799/withoutBody?foo=foo"

After adding try catch new exception message is
"org.apache.http.client.ClientProtocolException executing PUT http://localhost:33095/withoutBody?foo=foo"

velo (approved):
Thanks for addressing this; the behavior change is covered by tests and the implementation looks compatible and secure.

gitshabh:
Hello I think you merged another PR which was exactly the same. I thought I did it first. It's fine anyways, maybe this PR can be closed now.

PR #3359, opened 2026-05-14
Only unwrap declared checked causes with UNWRAP

### LINKED ISSUE #1487: ErrorDecoder not invoked when using Apache Http client

young0264:
## Summary

This updates `UNWRAP` propagation so that Feign only rethrows:

- `RuntimeException` and `Error`
- checked exceptions declared by the target method, including assignable supertypes

Undeclared checked causes now remain wrapped in `RetryableException`.

## Why

Previously, `UNWRAP` would rethrow any non-null cause from `RetryableException`.

That could rethrow checked exceptions that are not declared by the target method, which can lead to inconsistent exception propagation at the proxy boundary.

This keeps the fix at the core boundary instead of adding client-specific handling or synthesizing responses.

## Tests

Added coverage for:

- undeclared checked causes staying wrapped with `UNWRAP`
- declared checked causes being unwrapped with `UNWRAP`
- assignable checked causes being unwrapped with `UNWRAP`
- the Apache HttpClient redirect-without-Location case staying wrapped as `RetryableException`

Fixes #1487

### DIFF
--- a/core/src/main/java/feign/SynchronousMethodHandler.java
+++ b/core/src/main/java/feign/SynchronousMethodHandler.java
@@ -76,10 +76,17 @@ final class SynchronousMethodHandler implements MethodHandler {
         } catch (RetryableException th) {
           Throwable cause = th.getCause();
           if (methodHandlerConfiguration.getPropagationPolicy() == UNWRAP && cause != null) {
-            throw cause;
-          } else {
-            throw th;
+            if (cause instanceof RuntimeException || cause instanceof Error) {
+              throw cause;
+            }
+            for (Class<?> exceptionType :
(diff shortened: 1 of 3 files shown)


velo (approved):
Thanks for the fix — this change is focused, includes targeted test updates, and looks backward-compatible and safe.

Issue #1193, opened 2020-03-18
Make FeignException truly serializable


Nikolas-Charalambidis:
The class `FeignException` implements `Serializable` inherited from the `Exception` hierarchy, however, it contains the object `feign.Request` which is not `Serializable` itself, therefore further serialization into either JSON or Camunda process variable is impossible (my case), etc... 

Is possible to either change the internals of `feign.Request` to be serializable itself or amend `FeignException` to hold information about the failed request in another way?

Version: 10.4.0

kdavisk6:
@Nikolas-Charalambidis 

I think we would ok with either option.  The simplest would be to investigate what it would take to make `Request` `Serializable`.  Would you be willing to look into that for us?

Nikolas-Charalambidis:
I have already looked into the internals of `feign.Request` (and its `feign.template.Template`) which use objects right from JDK such as `java.nio.charset.Charset` that is not `Serializable` itself...

I think there are two ways:

1. Make `Request` as `transient` in the `FeignException`.
2. Make `Request` implement `Serializable` and all instance field classes `Serializable`/transient`. 

The object can never be fully `Serializable` when it uses `java.nio.charset.Charset` as an instance field internally.

kdavisk6:
@Nikolas-Charalambidis 

I think it's reasonable enough for use to change the charset instance variable type, so long as the method interfaces don't change.  Are there any other instance variables that pose a problem?

Nikolas-Charalambidis:
@kdavisk6 

I have noticed no more problematic classes aside from `java.nio.charset.Charset`. We can substitute it with either a String literal or an enumeration from which we would be able to construct `Charset` where needed and keeping the interfaces untouched. If I propose a pull request, would you check it up?

kdavisk6:
Closed via #1262

PR #3360, opened 2026-05-18
feat!: make `feign.Request.Body` streaming-ready

yvasyliev:

### Summary

This PR is the first step toward resolving #2734 — enabling request body streaming in Feign. It builds on the foundational work from @trumpetinc in #2754, which had become too outdated to merge directly.

The core idea is to change `feign.Request.Body` from a `byte[]`-backed structure to a streaming-ready abstraction. This means request bodies are **no longer cached in memory** unless explicitly needed — enabling Feign to send large files and streams without buffering.

> [!IMPORTANT]
> This change does not introduce public streaming API in Feign. It only makes the Feign client streaming-ready. Public streaming API is planned to be implemented in a follow-up PR.

> [!WARNING]
> This is a **breaking change** and is intended to be released as `v14.rc.1`.

---

#### Core changes

- **`feign.Request.Body`** is now an **interface** with:
  - `writeTo(OutputStream)` — the primary write mechanism
  - `contentLength()` — returns `-1` for unknown/streaming bodies
  - `isRepeatable()` — `false` for non-repeatable bodies
  - `writeToByteArray()` and `writeToString(Charset)` — helper methods for repeatable bodies (use with caution)
  - `Body.of(...)` — factory methods for `String`, `byte[]` inputs
- **`feign.Request.BodyImpl`** is the default implementation for repeatable, non-streaming bodies. It overrides `toString()` for human-readable logging.
- **`feign.RequestTemplate`** now exposes a single `body(Request.Body)` setter. The old `body(byte[], Charset)` overload is `@Deprecated` for backward compatibility with `spring-cloud-openfeign-core`.
- **`feign.Request.body()`** now returns `Optional<Request.Body>` instead of `byte[]`.
- **`feign.Request#length()`** removed — use `Request.Body#contentLength()` instead.
- **`feign.RequestTemplate#requestBody()`** returns `Optional<Request.Body>`.
- Removed `feign.RequestTemplate#charset` instance variable — charset should be read from `Content-Type` headers.

#### HTTP client updates

All `feign.Client` implementations updated to stream bodies via `Request.Body#writeTo`:

| Client | Change |
|---|---|
| `DefaultClient` | `body.get().writeTo(out)` |
| `ApacheHttp5Client` | New `FeignBodyEntity` wrapping `Request.Body` |
| `AsyncApacheHttp5Client` | Migrated to `ClassicRequestBuilder` + `ClassicToAsyncRequestProducer` for true streaming |
| `OkHttpClient` | New anonymous `RequestBody` delegating to `feign.Request.Body` |
| `Http2Client` | `BodyPublishers.ofInputStream` via `PipedInputStream`/`PipedOutputStream` |
| `GoogleHttpClient` | New `FeignBodyContent` inner class |
| `JAXRSClient` | Uses `StreamingOutput` |
| `VertxHttpClient` | Uses new `OutputToReadStream` bridge (see below) |

#### Encoder updates

All encoder implementations updated to call `template.body(Request.Body.of(...))`:
- `GsonEncoder`, `Jackson3Encoder`, `JAXBEncoder`, `JacksonJaxbJsonEncoder`, `Fastjson2Encoder`, `MoshiEncoder`, `SOAPEncoder`, `DefaultEncoder`, `JsonEncoder`, `Fastjson2Encoder`

#### Vert.x streaming bridge

A new `feign.vertx.OutputToReadStream` class bridges Java blocking `OutputStream` to Vert.x `ReadStream<Buffer>`. It is adapted from [`io.cloudonix:vertx-java.io`](https://github.com/cloudonix/vertx-java.io) (MIT License) with a local compatibility fix to support **both Vert.x 4 and Vert.x 5** (using `onComplete` instead of `andThen` to avoid a runtime linkage to `io.vertx.core.Completable` absent in Vert.x 4).

An issue has been filed upstream: [cloudonix/vertx-java.io#8](https://github.com/cloudonix/vertx-java.io/issues/8). Once fixed upstream, `OutputToReadStream` can be removed from this repo.

`VertxFeign.Builder` now requires `.vertx(Vertx)` in addition to `.webClient(WebClient)`. A `NullPointerException` with a descriptive message is thrown if either is missing.

#### `FeignException` changes

- `FeignException.errorReading(...)` now passes `null` as the body (instead of `request.body()`), since the request body may be a non-repeatable stream and should not be cached.
- Related test assertions updated to expect `isEmpty()`.

#### Logging

- `Logger` updated to use `Request.Body#toString()` (provided by `BodyImpl`) for repeatable bodies.

#### Metrics

- `dropwizard-metrics4/5` `MeteredEncoder`: uses `body.contentLength()`.
- `micrometer` `MeteredEncoder`: reads `Content-Length` header for recording.

#### `mock` module

- `RequestKey` no longer stores or compares `Charset`.
- `RequestKey.Builder#charset(Charset)` removed.

---

### Known limitations / future work

1. **`spring-cloud-openfeign-core` compatibility**: `RequestTemplate#body(byte[], Charset)` kept as `@Deprecated`. Once the Spring team migrates to `body(Request.Body)`, this can be removed.
2. **Charset extraction**: There is no trivial vanilla Java way to parse charset from `Content-Type` headers. `UTF-8` is used as a fallback in `writeToString`. A separate issue/PR may be needed.
3. **`FeignException` design**: Response bodies are still cached as `byte[]` in `FeignException`. Reconsidering this is deferred to a future PR.
4. All tests pass, but the build should be triggered with `-Djapicmp.skip=true` property provided.

---

### Credits

Special thanks to @trumpetinc whose original #2754 laid the groundwork for this change.

---

### Related

- Partially implements #2734
- Based on #2754 by @trumpetinc
- Follow-up PRs planned:
  - Step 2: Make `feign.form.MultipartFormContentProcessor` streamable (introduce `class MultipartBody implements feign.RequestBody`)
  - Step 3: Make Response Body streamable


jackson-jaxb/src/main/java/feign/jackson/jaxb/JacksonJaxbJsonEncoder.java
```diff
@@ -46,7 +46,7 @@ public void encode(Object object, Type bodyType, RequestTemplate template)
     try {
       jacksonJaxbJsonProvider.writeTo(
           object, bodyType.getClass(), null, null, APPLICATION_JSON_TYPE, null, outputStream);
-      template.body(outputStream.toByteArray(), Charset.defaultCharset());
+      template.body(Request.Body.of(outputStream.toByteArray()));
```
trumpetinc, line 49: Instead of creating a temporary byte array, could we do something like this?

```
template.body( Request.Body.of(os -> jacksonJaxbJsonProvider.writeTo(
          object, bodyType.getClass(), null, null, APPLICATION_JSON_TYPE, null, os) ) );
```

This would require a Body.of method that would take an interface that can write body content to an arbitrary stream, so I'm thinking that maybe what I'm suggesting would be some future improvement?  If so, then I understand why it is done this way for now.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -237,6 +235,30 @@ private Builder newRequestBuilder(Request request, Options options) throws URISy
...
+        BodyPublishers.ofInputStream(
+            () -> {
+              PipedInputStream inputStream = new PipedInputStream();
+              try {
+                PipedOutputStream outputStream = new PipedOutputStream(inputStream);
+                CompletableFuture.runAsync(
+                    () -> {
+                      try (outputStream) {
+                        body.writeTo(outputStream);
+                      } catch (IOException ignored) {
```
trumpetinc: Is it a good idea to ignore here?  It seems like we may need to rethrow an unchecked exception....

trumpetinc:
I've read over the core changes, and I had one smallish input on this change:

"Removed feign.RequestTemplate#charset instance variable — charset should be read from Content-Type headers."

I think that this should be carefully considered from a backwards compatibility viewpoint.  Some REST servers do not properly set the charset header - in those cases, it is probably important that the Feign user be able to set an override using the @Headers annotation.

I have not reviewed the code in sufficient detail to know whether this type of override is still possible, but that should probably be checked.

We should also be aware that existing Feign users that rely on the old default UTF-8 content type behavior (for example, when interacting with servers that send the wrong content type header in the response), could have problems.  I don't really see a good way around this.

Personally, I think that this is a breaking change that we should be ok with - the library should have always used the headers for content type determination, and it is better to fix this now, even if it introduced breaking changes for some misconfigured servers.

jackson-jaxb/src/main/java/feign/jackson/jaxb/JacksonJaxbJsonEncoder.java
yvasyliev: While I like the elegance of that suggestion, I find three reasons to keep it as is:

1. In vast majority of cases, XML/JSON encoders operate with small chunk of data fitting in-memory.
2. It appears that some clients, servers and the `Logger` might benefit from `Content-Length` header. And to calculate the length, the data should be eagerly encoded into tangible bytes one way or another.
3. If there's retry feature enabled, `jacksonJaxbJsonProvider.writeTo` will be executed two or more times which is not efficient.

Additionally, I like your idea from the other thread to add support for raw `OutputStream` and `Writer` bodies (not implemented in this PR). If there's a really big piece of XML or JSON document, users could leverage raw body supplier to stream the large payload:

```java
interface Client {
    @RequestLine("POST /api/v1/send")
    @Header("Content-Type: application/json; charset=UTF-8")
    void send(ThrowingConsumer<OutputStream, IOException> body);
}

client.send(outputStream -> jacksonJaxbJsonProvider.writeTo(..., outputStream));
```

Alternatively, users can provide custom `Encoder` implementation with lazy encoding configured.

I'd be happy to hear maintainers' opinion too, and I'm open to any idea.

TL;DR I prefer to keep existing encoders as is and let users enable streaming on demand if needed.

java11/src/main/java/feign/http2client/Http2Client.java
yvasyliev: Good catch! Created `PropagatingPipedInputStream` to propagate IO errors from child thread to upstream.

Also, I replaced the hardcoded `CompletableFuture.runAsync` with injectable `Executor` in case if a user wants to utilize virtual threads or something.

yvasyliev:
> I've read over the core changes, and I had one smallish input on this change:
> 
> "Removed feign.RequestTemplate#charset instance variable — charset should be read from Content-Type headers."
> 
> I think that this should be carefully considered from a backwards compatibility viewpoint. Some REST servers do not properly set the charset header - in those cases, it is probably important that the Feign user be able to set an override using the @headers annotation.
> 
> I have not reviewed the code in sufficient detail to know whether this type of override is still possible, but that should probably be checked.
> 
> We should also be aware that existing Feign users that rely on the old default UTF-8 content type behavior (for example, when interacting with servers that send the wrong content type header in the response), could have problems. I don't really see a good way around this.
> 
> Personally, I think that this is a breaking change that we should be ok with - the library should have always used the headers for content type determination, and it is better to fix this now, even if it introduced breaking changes for some misconfigured servers.

I believe this is out of #2734 scope. From my experience, there's only a handful of cases where the client might not want to trust the server's `Content-Type` response header. For those scenarios, I would consider adding a `charset` instance variable to `Decoder` implementations so that the users can hardcode the desired charset via Feign builder. But again, I would do this as part of a separate initiative

If we're talking about the scenario where the server doesn't return `charset` or `Content-Type` values, the fallback remains `UTF-8`

jackson-jaxb/src/main/java/feign/jackson/jaxb/JacksonJaxbJsonEncoder.java
trumpetinc: I agree with you about most json encoders working on relatively small chunks.  And I agree that there is really no need to make changes to the existing encoders.

You make a *very* good point about logging - Feign users rely on body logging, and having the body content in memory makes logging a lot simpler.  I actually explored logging with streams in my initial prototypes for this project using buffered streams and mark/reset, but it was overly complicated and did not add sufficient value.  Your design is much better.

Note that there *is* a "streaming" encoder for Feign that looks like it is designed to handle very large json streams, and they might benefit from this (I haven't dug into that encoder in depth), but for the general case, I am in agreement that we should leave the existing encoders alone until there is a compelling reason to change them.

velo (changes requested):
Thanks for the contribution. I can’t approve this in the current state:\n\n1) Required checks are failing (), so this cannot be approved/merged yet.\n2) This changes core public request-body behavior and API surface (, ) with a  breaking-change signal. For a compatibility-sensitive library, this needs an explicit compatibility plan and migration path (or clear major-release targeting) before approval.\n\nPlease address CI and provide explicit compatibility/migration details in the PR.

velo (changes requested):
Thanks for the contribution. I can’t approve this in the current state:

1) Required checks are failing (ci/circleci: pr-build), so this cannot be approved/merged yet.
2) This changes core public request-body behavior and API surface (core/src/main/java/feign/Request.java, core/src/main/java/feign/RequestTemplate.java) with a breaking-change signal (feat!). For a compatibility-sensitive library, this needs an explicit compatibility plan and migration path (or clear major-release targeting) before approval.

Please address CI and provide explicit compatibility/migration details in the PR.

trumpetinc:
@yvasyliev @velo 

I will take a stab at the compatibility plan and migration path (I don't have edit rights on this PR but can include a proposal in comments for @yvasyliev to incorporate).

 @velo  IMPORTANT: The failing circleCI test is because of the breaking changes required for this PR.  What is the standard project process for allowing CI to run when there are intentional breaking changes?  I've looked all over and can't find a process document that covers this.  Thanks.

Here's the CI error:

`[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.25.6:cmp (default) on project feign-core: There is at least one incompatibility: 
feign.Request.body():METHOD_RETURN_TYPE_CHANGED,
feign.Request.body():METHOD_RETURN_TYPE_GENERICS_CHANGED,
feign.Request.charset():METHOD_REMOVED,
feign.Request.create(java.lang.String,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset):METHOD_REMOVED,
feign.Request.create(feign.Request$HttpMethod,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset):METHOD_REMOVED,
feign.Request.create(feign.Request$HttpMethod,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset,feign.RequestTemplate):METHOD_REMOVED,
feign.Request.isBinary():METHOD_REMOVED,feign.Request.length():METHOD_REMOVED,
feign.Request.method():METHOD_REMOVED,
java.io.Serializable[java.io.Serializable]:INTERFACE_REMOVED,
feign.RequestTemplate.body(java.lang.String):METHOD_REMOVED,
feign.RequestTemplate.body():METHOD_REMOVED,
feign.RequestTemplate.requestBody():METHOD_RETURN_TYPE_CHANGED,
feign.RequestTemplate.requestBody():METHOD_RETURN_TYPE_GENERICS_CHANGED,
feign.RequestTemplate.requestCharset():METHOD_REMOVED
 -> [Help 1]
`

trumpetinc:
@yvasyliev here is a first draft of a compatibility/migration plan.  Can you please review then add to the PR if you are good with it?

# Compatibility

This change includes breaking changes in how Request bodies are handled.  For most users, there will be no impact, but any code that interacts directly with request bodies will be impacted.

# Migration

## feign.Request

### body()

Instead of returning a byte[], this method now returns an Optional<Body>.  If accessing the underlying byte[] is still necessary, call `request.body().get().writeToByteArray()`

### charset()

No longer supported. Charset should be an internal implementation detail of the body, not the request.

### create(...)

The following methods have been removed:

- `create(feign.Request$HttpMethod,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset)`
- `create(java.lang.String,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset)`
- `create(feign.Request$HttpMethod,java.lang.String,java.util.Map,byte[],java.nio.charset.Charset,feign.RequestTemplate)`

Replace these calls with:

```
feign.Request.create(
      HttpMethod httpMethod,
      String url,
      Map<String, Collection<String>> headers,
      Body body,
      RequestTemplate requestTemplate)
```

For simple byte[] bodies, use `Body.of(byte[] content, Charset charset)`.  For simple String bodies with UTF-8 encoding, use `Body.of(String content)`


### isBinary()

No longer supported.  The Body implementation is now responsible for encoding for non-binary content if applicable.

### length()

Use `request.body().map(Body::length).orElse(0)`

### method()

This was a deprecated method that has now been removed.  Use `request.httpMethod()` instead.

### No longer implements java.io.Serializable

feign.Request is no longer Serializable.


## feign.RequestTemplate

### body(String)

use `body(Body.of(String))`

### body()

use `requestBody()` - please note that requestBody is itself deprecated and will be removed in later releases.

### requestBody()

Use `requestBody().get().writeToByteArray()`

Note that this alternative method is deprecated and will be removed in later releases.

### requestCharset()

No longer supported. Charset should be an internal implementation detail of the body, not the request or request template.

yvasyliev:
Thank you @trumpetinc !

---

Hey @velo ,

1. Like @trumpetinc mentioned, the build was failing because of `com.github.siom79.japicmp:japicmp-maven-plugin`. It checks for breaking changes which are intentionally featured in this PR. The easiest fix I found is to exclude `Request`, `RequestTemplate`, `RequestKey` & `VertxHttpClient` classes from the plugin scope. I guess, once the major version is released, we can remove the exclusions.
2. I added the [MIGRATION-v14.md](https://github.com/OpenFeign/feign/pull/3360/changes#diff-ce2cccf277a3f780600f3dbc33ad580b28105595dabf696b7d3a92b4e29a2092) for migration guide.

The [ci/circleci: pr-build](https://circleci.com/gh/OpenFeign/feign/11839) has passed. Could you please review the PR?

Thank you!

velo (changes requested):
Thanks for the updates and for adding migration notes. I still can’t approve this PR in its current form.

Gate results:
1) Test coverage: PASS (extensive test updates are present).
2) Backwards compatibility: FAIL.
3) Security: PASS.

Blocking issue (compatibility):
- This PR intentionally removes/changes core public API in `core/src/main/java/feign/Request.java` and `core/src/main/java/feign/RequestTemplate.java` (e.g., `Request.Body` shape and `Request.create(...)`/body access patterns), and adds `MIGRATION-v14.md` describing the break.
- For a compatibility-sensitive library, this cannot be approved on a normal PR track unless it is explicitly handled as a major-release-only change with clear release governance.

Please split this into a major-version release plan (or stage it behind a compatibility-preserving approach) before it can be approved.

yvasyliev:
>Please split this into a major-version release plan (or stage it behind a compatibility-preserving approach) before it can be approved.

Sorry @velo I'm not sure I understand what action items are pending on this PR to have the green light. Do I need to bump the library version to `14.0`?

trumpetinc:
@kdavisk6 I wanted to pull you into this discussion. The proposal you and I discussed last year has gotten solid traction, and @yvasyliev has a PR that makes the changes necessary for Feign to support streaming data while minimizing impact to regular users.

As we discussed last year, this will require a breaking change - but the impact is restricted to library users who are interacting directly with Request. Users that use Feign "normally" using annotations, etc will not be impacted. 

If you have any feedback on how we can effectively navigate the PR process for a breaking change like this i would appreciate it!

Thanks, 

K

core/src/main/java/feign/Request.java
```diff
@@ -19,21 +19,25 @@
...
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 /** An immutable request to an http server. */
-public final class Request implements Serializable {
+public final class Request {
```
trumpetinc: I guess I should ask why we are dropping Serializable?  Personally, I don't think it should have been there in the first place, but removing it adds more to the breaking change footprint.  Is there any disadvantage to leaving it?

core/src/main/java/feign/Request.java
```diff
@@ -513,78 +425,141 @@ public RequestTemplate requestTemplate() {
...
-      this.data = data;
+    /**
+     * Creates a new {@link Body} instance from the provided byte array.
+     *
+     * @param content the byte array representing the body content
+     * @return a new {@link Body} instance
+     * @apiNote It's assumed that the byte array can be converted to a string using {@link
+     *     StandardCharsets#UTF_8} charset.
+     */
+    static Body of(byte[] content) {
```
trumpetinc: You may want to consider not including this method right now.  It is a little confusing having a pure byte oriented call that implicitly involves charsets.

I couldn't find any callers of this, and given potential for confusion, it is probably safest to just not include it.

core/src/main/java/feign/Request.java
```diff
@@ -513,78 +425,141 @@ public RequestTemplate requestTemplate() {
...
-      return data != null ? data.length : 0;
+    /**
+     * Creates a new {@link Body} instance from the provided byte array, using the specified
+     * charset.
+     *
+     * @param content the byte array representing the body content
+     * @param charset the content charset
+     * @return a new {@link Body} instance
+     */
+    static Body of(byte[] content, Charset charset) {
```
trumpetinc: I have a similar question about the need to expose this as public API.  I think the of(xxx) methods are primarily about backwards compatibility for character based request bodies, in which case only `of(String content, Charset charset)` and `of(String content)` are needed.

I would move the `new Request.BodyImpl(...)` code  into the of(String, Charset) method and remove both of(byte[]....) calls.

core/src/main/java/feign/Request.java
```diff
@@ -513,78 +425,141 @@ public RequestTemplate requestTemplate() {
...
+    void writeTo(OutputStream outputStream) throws IOException;
+
+    /**
+     * Writes the body content to a string using the specified charset for decoding.
+     *
+     * @param charset the charset to be used for decoding the body content
+     * @return a string representation of the body content
+     * @throws IOException if an I/O error occurs while writing the body content to a string
+     */
+    default String writeToString(Charset charset) throws IOException {
```
trumpetinc: very small thing, and probably not worth worrying about, but I prefer `asString(chaset)` and `asByteArray()`.

Usually writeToXXX() implies that we will pass XXX in as a target to the method (which isn't the case here or in the writeToByteArray() method).

core/src/main/java/feign/Request.java
```diff
@@ -513,78 +425,141 @@ public RequestTemplate requestTemplate() {
...
 
-    public String asString() {
-      return !isBinary() ? new String(data, encoding) : "Binary data";
+    /**
+     * Writes the body content to a byte array.
+     *
+     * @return a byte array containing the body content
+     * @throws IOException if an I/O error occurs while writing the body content to a byte array
+     */
+    default byte[] writeToByteArray() throws IOException {
```
trumpetinc: consider byte[] asByteArray()

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -278,4 +318,35 @@ private String[] asString(Map<String, Collection<String>> headers) {
                     .flatMap(List::stream))
         .toArray(String[]::new);
   }
+
+  private static class PropagatingPipedInputStream extends PipedInputStream {
```
trumpetinc, line 328: I know this is a private class, but some javadocs would be helpful to have...

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -83,12 +94,21 @@ public Http2Client() {
             .build());
   }
 
+  public Http2Client(HttpClient client) {
+    this(client, ForkJoinPool.commonPool());
```
trumpetinc, line 98: This seems like it is more than just backwards compatibility.  Can you expand on why these changes are being made?  I'm not opposed - I would just like to understand what's going on.

MIGRATION-v14.md
```diff
@@ -0,0 +1,419 @@
...
+`feign.Request.Body` has been redesigned from a `byte[]`-backed concrete class into a **streaming-ready interface**.
+Request bodies are no longer eagerly buffered in memory unless you explicitly use the `byte[]`/`String` factory methods.
+
+This is a **breaking change** affecting:
+
+- Custom `Encoder` implementations
+- Custom `Client` implementations
+- Any code that directly reads `Request.body()`, `Request.length()`, `Request.charset()`, or `RequestTemplate.body()`/
+  `RequestTemplate.requestBody()`
+
```
trumpetinc, line 21: Consider adding the following:

What is NOT breaking:  Regular use of Feign via annotations and Feign.builder().target().  As such, most users will not be impacted by this change.

MIGRATION-v14.md
```diff
@@ -0,0 +1,419 @@
...
+```java
+boolean binary = request.isBinary();
+```
+
+**After:**
+
+```java
+// isRepeatable() does not throw, so Optional.map is safe here.
+boolean repeatable = request.body()
+    .map(Request.Body::isRepeatable)
```
trumpetinc, line 141: I'm not sure this is a direct replacement...  It may be better to say "No direct replacement. In some situations, isRepeatable may be a useful replacement, depending on why you were checking isBinary."

trumpetinc (commented):
I did another full review and have added a few minor items.

I also reviewed the migration guide in detail and provided some feedback.  All in all, this is an excellent document (probably one of the best I've seen across many projects I've worked on).

core/src/main/java/feign/Request.java
yvasyliev: It's in use by some encoders (`DefaultEncoder`, `JacksonEncoder`, `JacksonJaxbJsonEncoder`)

JSON encoders should be fine with implicit `UTF-8` charset. Otherwise, I need to update the encoding logic (encode to String) which I'm reluctant to do tbh

`BodyImpl` needs the charset for `toString()`/logging purposes only. If a given `byte[]` is not a human-readable sequence of bytes, then the caller shouldn't even bother logging the body (=providing any charset)

core/src/main/java/feign/Request.java
yvasyliev: I have no clue why it was `Serializable` without a `serialVersionUID`. I don't find any benefit of keeping it since `feign.Request#body` is no longer serializable

I hope the maintainers will let us know if it's necessary to keep it

core/src/main/java/feign/Request.java
yvasyliev: It actually was `asString` at some point. I decided to move to `writeToString` to emphasize that this is an IO operation, so that `asString` is not confused with `toString`, especially considering that the body might be non-repeatable.

I'm now thinking that `writeAsString` might be a good alternative

Please let me know what option you prefer and I update accordingly

java11/src/main/java/feign/http2client/Http2Client.java
yvasyliev: So JDK HTTP Client takes `InputStream` as request body content. In order to adapt `feign.Request.Body` to `InputStream`, I need to leverage a combination of piped input/output streams.

If I just make

```java
BodyPublisher publisher =
        BodyPublishers.ofInputStream(
            () -> {
              PipedInputStream inputStream = new PipedInputStream();
              try (PipedOutputStream outputStream = new PipedOutputStream(inputStream)) {
                body.writeTo(outputStream);
              } catch (IOException e) {
                throw new UncheckedIOException(e);
              }
              return inputStream;
            });
```

then:

1. The whole request body will be cached in-memory.
2. `return inputStream` will get blocked until the body is fully written to the `outputStream`.

To kill two birds with one stone, we need to execute `body.writeTo(outputStream)` asynchronously and therefore `PropagatingPipedInputStream` class & `feign.http2client.Http2Client#executor` instance variable were introduced.

`Http2Client(HttpClient)` constructor must be sufficient for most users. `ForkJoinPool#commonPool` is created & managed by JVM (no user intervention is required). By default, the same pool is used when executing `CompletableFuture#runAsync(Runnable)` while offering `CompletableFuture#runAsync(Runnable, Executor)` alternative.

Similarly, in case if anyone wants to supply & manage the executor on their own, they can utilize `Http2Client(HttpClient, Executor)` constructor.

Thus, I kept the constructor to provide both backward compatibility and convenience

java11/src/main/java/feign/http2client/Http2Client.java
yvasyliev: The Javadoc is added

MIGRATION-v14.md
yvasyliev: Added the clarification on [Line 14](https://github.com/OpenFeign/feign/pull/3360/changes#diff-ce2cccf277a3f780600f3dbc33ad580b28105595dabf696b7d3a92b4e29a2092R14)

MIGRATION-v14.md
yvasyliev: Added the comment clarifying the same

core/src/main/java/feign/Request.java
trumpetinc: oh weird - my "Callers" search seems to be failing.  Sorry about that.  Totally agree with leaving it as-is.  It is a point of confusion, but not worth changing right now.

core/src/main/java/feign/Request.java
trumpetinc: I completely agree in principle (that should never have passed code review when it was originally written).

My concern is that this might create more friction to us getting the PR accepted - and it's not relevant to the big picture we are trying to achieve.

On the flip side, if you are making a breaking change, it's best to break everything that needs to be broken at the same time :-)  We'll see what they say.

core/src/main/java/feign/Request.java
trumpetinc: Ah - good point - the potential for it being an IO operation (and throwing IOException) does make it not quite as clear...

I'm fine with either naming - this was a minor item and I don't feel that strongly about it...

java11/src/main/java/feign/http2client/Http2Client.java
trumpetinc: I just did a dive into HttpClient BodyPublisher, and I understand now.  It's unfortunate that they made the design decision they did (it's a similar decision to the original Feign design that we are trying to fix with this PR).

Thanks for the explanation.  It does make sense that we'd need to do this to properly support streaming.

trumpetinc:
@yvasyliev my code review is complete and all of your changes look good to me.  Thank you for your hard work on this - I'm hoping we can get approval shortly.

core/src/main/java/feign/Request.java
yvasyliev: @trumpetinc I found this thread - #1193

It seems that we need to either redesign `FeignException` as part of the major release or keep `Request` serializable (does it have any value without the body?)

velo (changes requested):
Thanks for the updated migration notes. I still cannot approve this PR in its current form.

Gate results:
1) Test coverage: PASS.
2) Backwards compatibility: FAIL.
3) Security: PASS.

Blocking issue:
- This still intentionally changes/removes core public API in core/src/main/java/feign/Request.java and core/src/main/java/feign/RequestTemplate.java, including request body access and factory methods. The migration guide documents the break, but it does not change the compatibility gate result.

Please split this into an explicit major-version release plan, or stage it behind a compatibility-preserving approach, before it can be approved.

velo:
ok, I will move main to `14.0-alpha1-SNAPSHOT` and then we can merge

velo (approved):

core/src/main/java/feign/Request.java
trumpetinc: Ugh, Java serialization was such a broken design. 

And the fact that Feign uses the *response* body field in FeignException to store *request* bodies is super awkward. FeignException should have been an abstract base class with different subclasses for requests vs responses.

But that is not our problem to solve.

Here is my proposal - do you think this is a reasonable compromise?

Make FeignExceptiom serializable again (and include a comment about explaining why with link back to the issue you found).

Change the "request" constructor so it accepts Body. It will then call toString() like the Feign logger does. Then convert the string to byte[] using UTF_8 encoding.

And include comments explaining that we know this is weird, but we are having to do this to work around FeignException not having an abstract base class with a separate class for request exceptions vs response exceptions.

yvasyliev: Decided not to break the backward compatibility for `FeignException` for now. Made the `Request` serializable again, but the `Request#body` & `RequestTemplate#body` are transient now. I'd suggest revising the exception handling in a separate PR

yvasyliev:
@velo I had to make the `Request` serializable back again to avoid regression issues. Please review one more time🙏

PR #3382, opened 2026-06-02
Preserve delegate content type for multipart parameters

### LINKED ISSUE #2813: Wrong content-type for multipart requests on JSON parts (text/plain instead of application/json)
I have the following request mapping defined which includes 2 parts:

- metadata: POJO (application/json)
- files: multipart

```java
    @RequestMapping(
        method = RequestMethod.POST,
        value = "/v1/files/bulk",
        produces = { "application/json" },
        consumes = "multipart/form-data"
    )
    ResponseEntity<Files> saveFiles(
        @Parameter(name = "metadata", description = "Array of file attributes", required = true) @Valid @RequestPart(value = "metadata", required = true) List<@Valid Metadata> metadata,
        @Parameter(name = "files", description = "Array of Files") @RequestPart(value = "files", required = false) List<MultipartFile> files
    );
```
which produces the following body:

```
--195903cbf58
Content-Disposition: form-data; name="metadata"
Content-Type: text/plain; charset=UTF-8

[{"fileName":"somefileName","hash":"somehash"}]
--195903cbf58
Content-Disposition: form-data; name="files"; filename=""
Content-Type: application/octet-stream
Content-Transfer-Encoding: binary

--195903cbf58--
```

It shows that the content type of the JSON part is set to **text/plain** instead of **application/json**.
This is due to the fact that **feign.form.multipart.DelegateWriter#parameterWriter** created in **feign.form.MultipartFormContentProcessor#MultipartFormContentProcessor** is set to **feign.form.multipart.SingleParameterWriter** which is using a hard-coded content-type of **text/plain**.

I have noticed that the **RequestTemplate** that is used to write the JSON data actually has the correct headers set ("Content-Type: application/json"). However, this is ignored as the hard-coded value is used.

I would expect that the content type is taken from the **RequestTemplate**.

The only workaround that I found is to do a String replace on the body like this:

```java
public class SpringFormEncoderWithContentTypeCorrection extends SpringFormEncoder {

    private static final String CONTENT_TYPE_STRING_TO_REPLACE = "Content-Type: text/plain; charset=";
    private static final String NEW_CONTENT_TYPE_STRING = "Content-Type: application/json; charset=";

    public SpringFormEncoderWithContentTypeCorrection(Encoder delegate) {
        super(delegate);
    }

    @Override
    public void encode(Object object, Type bodyType, RequestTemplate template) throws EncodeException {
        super.encode(object, bodyType, template);
        if (template.body() != null) {
            // modify content type from 'text/plain' to 'application/json' for json parts
            String modifiedContent = (new String(template.body())).replace(CONTENT_TYPE_STRING_TO_REPLACE, NEW_CONTENT_TYPE_STRING);
            template.body(modifiedContent);
        }
    }
}
```

Configuration:
```java
    @Bean
    @Primary
    @Scope("prototype")
    public Encoder feignFormEncoder(final ObjectFactory<HttpMessageConverters> messageConverters) {
        return new SpringFormEncoderWithContentTypeCorrection(new SpringEncoder(messageConverters));
    }
```

Versions used:
* openfeign: 13.5
* spring-cloud-openfeign-core: 4.2.0
* Spring Boot: 3.4.2

seonwooj0810:
Fixes #2813

## Problem
When a multipart request contains a POJO/JSON part, `DelegateWriter` encodes the value through the delegate `Encoder` (e.g. the Jackson encoder), which correctly sets `Content-Type: application/json` on the throwaway `RequestTemplate`. That header was then discarded, and `SingleParameterWriter` hard-coded `Content-Type: text/plain`. As a result, JSON parts were emitted as `text/plain` instead of `application/json` (see the report in #2813).

## Change
- `DelegateWriter` now reads the `Content-Type` produced by the delegate and forwards it to a new `SingleParameterWriter.writeWithContentType(output, key, value, contentType)` method.
- When the delegate sets no content type, the writer falls back to the previous `text/plain; charset=<charset>` behaviour, so plain single parameters are unaffected.
- The existing `write(output, key, value)` template method is preserved and simply delegates with a `null` content type.

## Tests
Added `feign.form.multipart.DelegateWriterTest`:
- `usesContentTypeFromDelegate` — asserts a delegate that sets `application/json` produces `Content-Type: application/json` (and not `text/plain`).
- `fallsBackToTextPlainWhenDelegateSetsNoContentType` — asserts the `text/plain; charset=UTF-8` fallback when the delegate sets no content type.

## Test evidence
Ran the `form` module test suite (and its upstream dependencies) locally:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- feign.form.multipart.DelegateWriterTest
... full feign-core + feign-form suites green ...
BUILD SUCCESS
```

Verification done: confirmed the hard-coded `text/plain` still present on `master` in `SingleParameterWriter#write`; confirmed no in-flight PR referencing the issue and no active self-claim (maintainer invited a PR with tests on 2026-05-27); fix touches only `.java` files; full `form` module test suite passes after the change.

### DIFF
--- a/form/src/main/java/feign/form/multipart/DelegateWriter.java
+++ b/form/src/main/java/feign/form/multipart/DelegateWriter.java
@@ -48,6 +48,18 @@ public class DelegateWriter extends AbstractWriter {
     delegate.encode(value, value.getClass(), fake);
     val bytes = fake.body();
     val string = new String(bytes, output.getCharset()).replaceAll("\n", "");
-    parameterWriter.write(output, key, string);
+    parameterWriter.writeWithContentType(output, key, string, contentType(fake));
+  }
+
+  private static String contentType(RequestTemplate template) {
+    val headers = template.headers().get("Content-Type");
+    if (headers != null) {
+      for (val header : headers) {
+        if (header != null && !header.isEmpty()) {
+          return header;
+        }
+      }
+    }
+    return null;
   }
 }
--- a/form/src/main/java/feign/form/multipart/SingleParameterWriter.java
+++ b/form/src/main/java/feign/form/multipart/SingleParameterWriter.java
@@ -34,14 +34,31 @@ public class SingleParameterWriter extends AbstractWriter {
 
   @Override
   protected void write(Output output, String key, Object value) throws EncodeException {
+    writeWithContentType(output, key, value, null);
+  }
+
+  /**
+   * Writes a single parameter using the given content type.
+   *
+   * @param output output writer.
+   * @param key name for piece of data.
+   * @param value piece of data.
+   * @param contentType the content type of the part. May be {@code null}, in which case {@code
+   *     text/plain} with the output charset is used.
+   * @throws EncodeException in case of write errors
+   */
+  protected void writeWithContentType(Output output, String key, Object value, String contentType)
+      throws EncodeException {
+    val contentTypeHeader =
+        contentType != null ? contentType : "text/plain; charset=" + output.getCharset().name();
     val string =
         new StringBuilder()
             .append("Content-Disposition: form-data; name=\"")
             .append(key)
             .append('"')
             .append(CRLF)
-            .append("Content-Type: text/plain; charset=")
-            .append(output.getCharset().name())
+            .append("Content-Type: ")
+            .append(contentTypeHeader)
             .append(CRLF)
             .append(CRLF)
             .append(value.toString())
--- a/form/src/test/java/feign/form/multipart/DelegateWriterTest.java
+++ b/form/src/test/java/feign/form/multipart/DelegateWriterTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2012 The Feign Authors (feign@commonhaus.dev)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feign.form.multipart;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import feign.codec.Encoder;
+import java.nio.charset.StandardCharsets;
+import org.junit.jupiter.api.Test;
+
+class DelegateWriterTest {
+
+  private static final String BOUNDARY = "boundary";
+
+  private static final String KEY = "metadata";
+
+  @Test
+  void usesContentTypeFromDelegate() throws Exception {
+    Encoder delegate =
+        (object, bodyType, template) -> {
+          template.header("Content-Type", "application/json");
+          template.body("{\"hash\":\"somehash\"}");
+        };
+
+    assertThat(write(delegate))
+        .contains("Content-Type: application/json")
+        .doesNotContain("Content-Type: text/plain");
+  }
+
+  @Test
+  void fallsBackToTextPlainWhenDelegateSetsNoContentType() throws Exception {
+    Encoder delegate = (object, bodyType, template) -> template.body("plain");
+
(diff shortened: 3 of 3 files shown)


velo (changes requested):
Thanks for the focused fix and for adding coverage around the delegated part content type. I cannot approve this while the required CI check is failing.\n\nGate results:\n1) Test coverage: PASS.\n2) Backwards compatibility: PASS.\n3) Security: PASS.\n\nBlocking issue:\n- Required check  is currently failing for this PR. Please get the build green before this can be approved or merged.

seonwooj0810:
Thanks for the review @velo. The failing `pr-build` was `feign.vertx.ConnectionsLeakTests.http2NoConnectionLeak` in `feign-vertx5-test` — unrelated to this change (this PR only touches `feign-form` multipart). That flakiness was fixed on master by 6769d9ad ("Fix flaky vertx http2NoConnectionLeak test by using h2c prior knowledge"), so I rebased onto the latest master to pick it up and retrigger CI.

seonwooj0810:
Quick follow-up: CI is now fully green on the rebased branch — `ci/circleci: pr-build`, `ci/circleci: setup-environment`, and `security/snyk` all pass. The earlier failure was the unrelated flaky `feign-vertx5-test` `http2NoConnectionLeak` test, which was fixed on master and is no longer present after the rebase. This PR only touches `feign-form` multipart handling. Happy to make any further adjustments if needed.

velo (approved):
Thanks for the update. The new regression tests cover both delegated content-type propagation and the existing fallback; the change preserves the protected write path and public behavior, and introduces no security issue. Required checks are green. The branch still conflicts with the base branch, so it cannot be merged until the conflict is resolved.

seonwooj0810:
Thanks for the approval @velo. I've rebased the branch onto the latest `master` to resolve the conflict in `DelegateWriter` — `master` had refactored body extraction to use `RequestTemplate#requestBody()` / `Request.Body#writeToString`, so I adapted the content-type propagation to route the delegate's `Content-Type` through `SingleParameterWriter#writeWithContentType` on top of that new API. The fix and its regression tests are otherwise unchanged. The branch now reports mergeable (head `ba1cf15a`).

seonwooj0810:
The current `pr-build` failure (CircleCI 12024) is a test-compilation error in `feign-core` that is pre-existing on `master`, not introduced by this PR: `JavaLoggerTest` and `LoggerRebufferTest` still call the `Request.create(..., byte[], Charset)` overload that was removed in f9159cfa ("Charset cannot be converted to RequestTemplate"). Master's own snapshot builds fail with the identical error (CircleCI 11938 and 11953), and those two test files are byte-identical to `master` on this branch — the build never reaches `feign-form`. cc @velo in case the master fix isn't already in flight.

While reproducing this locally I noticed the rebased `DelegateWriterTest` also still used the removed `RequestTemplate#body(String)`, which would have failed compilation once `feign-core` is fixed. I've pushed d9d46e92 updating it to `template.body(Request.Body.of(...))` to match the new streaming body API. With that change all `feign-form` tests pass locally (31/31, including `DelegateWriterTest`). Once master compiles again, a CI re-run here should go green.

velo (changes requested):
Thanks for adapting the regression test to the current Request.Body API. The test coverage, backwards-compatibility, and security gates pass, but the required ci/circleci: pr-build check is failing, so this cannot be approved or merged until CI is green.

seonwooj0810:
To unblock the CI gate here, I've opened #3406 fixing the `feign-core` test-compilation error on `master` (`JavaLoggerTest`/`LoggerRebufferTest` still using the `Request.create` overload removed in f9159cfa). Once that lands, a re-run of `pr-build` on this branch should go green — the failure never reaches `feign-form`, and all 31 `feign-form` tests pass locally on this branch.

seonwooj0810:
Update: #3406 has been merged, so the `feign-core` test-compilation error that was failing `ci/circleci: pr-build` is now fixed on `master`. I've rebased this branch onto the latest `master` (no conflicts) so the gate builds against that fix.

Verified locally: `feign-core` test sources now compile and `DelegateWriterTest` passes. Should be green now — ready for another look whenever you have a moment, @velo. Thanks!

velo (approved):
Thanks for carrying the delegate's Content-Type through the multipart writer. The fallback keeps existing plain-parameter behavior intact, DelegateWriterTest covers both delegated and fallback paths, and the required checks are green now.

Issue #2998, opened 2025-08-25
Is it safe to use RequestTemplate#MethodMeta?


Dainerx:
Hello, 

My goal is to get the url template without the host and port. 

> Given: https://{{host}}:{{port}}/v1/api/resource/{id}}
> Wanted: /v1/api/resource/{id} 

Today, this can be achieved by calling methodMetadata. However, it is annotated with `Experimental`

```java
@Experimental
public RequestTemplate methodMetadata(MethodMetadata methodMetadata) {
    this.methodMetadata = methodMetadata;
    return this;
}
``` 

I am using it, in production code, as follow: 

`templatedRouted = request.requestTemplate().methodMetadata().template().path();` 

-- 

First things first, I would have preferred to use `requestTemplate.uriTemplate()` but uriTemplate field is not exposed as of today. Moreover `requestTemplate.url()` appends the path (https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/RequestTemplate.java#L546). 

Is it safe to use `methodMetadata()`? Have you gathered enough data to make a decision? Finally, should I check against nil?

Thanks!

velo:
Thanks for the report. This project is community-maintained, and the maintainer will not implement fixes or features directly. If you want this to move forward, please open a pull request with automated tests that cover the change.

Dainerx:
@velo PTAL https://github.com/OpenFeign/feign/pull/3393

PR #3393, opened 2026-06-05
[issue #2998] Expose uriTemplate in RequestTemplate

Dainerx:
PR proposal for #2998


velo (approved):
Thanks, this is a small additive API change with focused coverage for the template path behavior. I do not see compatibility or security concerns.

PR #3394, opened 2026-06-06
Avoid ClassLoader leak from static default async executor (#3178)

### LINKED ISSUE #3178: [Bug] Static ExecutorService in AsyncFeign causes ClassLoader Leak (Metaspace OOM)

seonwooj0810:
Fixes #3178

## Problem
`AsyncFeign` held its default `ExecutorService` in a private **static** singleton (`LazyInitializedExecutorService`) that is never shut down:

```java
private static class LazyInitializedExecutorService {
  private static final ExecutorService instance =
      Executors.newCachedThreadPool(r -> { Thread t = new Thread(r); t.setDaemon(true); return t; });
}
```

Inside a servlet container, marking the threads as daemon only lets the JVM exit — it does not help on redeploy. The static cached thread pool retains a strong reference to the initial application's `ContextClassLoader`, so each redeployment leaks the previous application's `ClassLoader` and its threads, eventually leading to a Metaspace `OutOfMemoryError`.

## Change
- Replace the static singleton with an **instance-scoped** default executor created per built client (`defaultExecutorService()`). Once the built client is discarded, the executor and its (daemon, idle-timing-out) threads become eligible for garbage collection, releasing the `ClassLoader`.
- Add `AsyncBuilder.executorService(ExecutorService)` so callers can supply and own a managed, shut-downable executor — the recommended way to control this lifecycle explicitly. It is ignored when a custom `client(AsyncClient)` is provided.

Default behavior is preserved for callers who don't configure anything: a cached daemon thread pool is still used; it is simply no longer a process-wide static singleton.

Note: `feign.kotlin.CoroutineFeign` contains the same static-singleton pattern. I kept this PR scoped to `AsyncFeign` (the reported class); happy to follow up on the Kotlin module if you'd like it addressed here too.

## Tests
Added to `core/src/test/java/feign/AsyncFeignTest.java`:
- `usesProvidedExecutorService` — builds an `AsyncFeign` with a spied `ExecutorService`, performs a request, and verifies the provided executor actually ran the work (`submit(...)` invoked).
- `defaultExecutorServiceStillExecutesRequests` — guards that the default (no executor configured) path still works end-to-end after removing the static singleton.

## Test evidence
`./mvnw -pl core -am test -Dtest=AsyncFeignTest` → `Tests run: 54, Failures: 0, Errors: 0, Skipped: 0`. Code format validated with `git-code-format:validate-code-format` (BUILD SUCCESS).

Verification done: confirmed no in-flight PR referencing #3178; reproduced the static-singleton pattern still present on `master` (`AsyncFeign.java`); maintainer invited a PR with automated tests on the issue; fix touches only `.java` files; build + targeted tests + format check pass locally on JDK 25.

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -56,23 +56,28 @@ public final class AsyncFeign<C> {
     return builder();
   }
 
-  private static class LazyInitializedExecutorService {
-
-    private static final ExecutorService instance =
-        Executors.newCachedThreadPool(
-            r -> {
-              final Thread result = new Thread(r);
-              result.setDaemon(true);
-              return result;
-            });
+  /**
+   * Creates the default {@link ExecutorService} used when neither a custom {@link AsyncClient} nor
+   * a custom executor is provided. The executor is created per built client (instance scoped)
+   * rather than being held in a static singleton, so that it - together with its daemon threads -
+   * becomes eligible for garbage collection once the owning client is discarded. This avoids the
+   * {@code ClassLoader} leak that a static singleton causes on application redeployments inside
+   * servlet containers (see gh-3178).
+   */
+  private static ExecutorService defaultExecutorService() {
+    return Executors.newCachedThreadPool(
+        r -> {
+          final Thread result = new Thread(r);
+          result.setDaemon(true);
+          return result;
+        });
   }
 
   public static class AsyncBuilder<C> extends BaseBuilder<AsyncBuilder<C>, AsyncFeign<C>> {
 
     private AsyncContextSupplier<C> defaultContextSupplier = () -> null;
-    private AsyncClient<C> client =
-        new DefaultAsyncClient<>(
-            new DefaultClient(null, null), LazyInitializedExecutorService.instance);
+    private AsyncClient<C> client;
+    private ExecutorService executorService;
     private MethodInfoResolver methodInfoResolver = MethodInfo::new;
 
     @Deprecated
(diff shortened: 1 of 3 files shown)


velo (changes requested):
Thanks for the fix and the executor coverage. I cannot approve this while the required CircleCI pr-build check is failing.\n\nGate results:\n1. Test coverage: PASS.\n2. Backwards compatibility: PASS.\n3. Security: PASS.\n\nBlocking issue:\n- Required check ci/circleci: pr-build is currently failing for this PR. Please get the build green before this can be approved or merged.

seonwooj0810:
I tracked the failing `ci/circleci: pr-build` down to `feign.BaseBuilderTest.checkEnrichTouchesAllAsyncBuilderFields`. The PR added the `AsyncFeign.AsyncBuilder.executorService` field, but the test still expected 12 enrichable fields, while reflection now correctly sees 13.

I pushed ea6c1db62bd5824f199c5f027cf4792d33f77af8 to update the expected field count. CI has been retriggered on the new head.

seonwooj0810:
Follow-up: the first push only exposed the next assertion in the same test. The final fix is 268d8e941055393fbefed2aecc98a83f7fd91ee0: `executorService` is now excluded from `BaseBuilder#getFieldsToEnrich()` because it is a caller-owned lifecycle resource and `Capability` does not enrich `ExecutorService`. The expected async builder field count remains 12.

velo (approved):
Thanks for updating the branch. The per-client default executor and caller-managed executor option are covered by regression tests, preserve existing API behavior, and introduce no security concern. Required checks are green.

PR #3396, opened 2026-06-07
feat: add streaming support for request bodies in `DefaultEncoder`

### LINKED ISSUE #220: Streaming support

yvasyliev:
### Overview

A second PR in a series related to #2734 feature implementation.

`DefaultEncoder` previously accepted only `String` and `byte[]` body types, forcing callers to buffer file contents or stream data into a `byte[]` before encoding. This change extends `DefaultEncoder` with native support for `File`, `Path`, `InputStream`, and `Request.Body`, enabling streaming without in-memory buffering.

### What changed

**`DefaultEncoder`** — four new `instanceof` branches dispatch to the appropriate `Request.Body` implementation. `File` is normalized to `Path` before dispatch, so both go through `PathBody`.

**`Request.PathBody`** (new) — a `Body` backed by a filesystem path. Opens a fresh `InputStream` on each `writeTo()` call, so `isRepeatable()` returns `true` and retry-capable clients can re-send the body without special handling. `contentLength()` returns `Files.size(path)`, falling back to `-1` on I/O error, allowing clients to emit a `Content-Length` header when the file is accessible.

**`Request.InputStreamBody`** (new) — a `Body` backed by a caller-supplied `InputStream`. Because an `InputStream` can only be consumed once, `isRepeatable()` returns `false` (interface default) and `contentLength()` returns `-1` (interface default). Callers are responsible for stream lifecycle.

**`Util.copy()`** — visibility widened from `private` to `public` so that `PathBody` and `InputStreamBody` in `Request.java` (the `feign-api` module) can share the same buffered copy loop.

**`Encoder.java` Javadoc** — the example updated to use `Request.Body.of(...)` consistent with the v14 API.

**`MIGRATION-v14.md`** — adds a non-breaking change entry for the new `DefaultEncoder` types.

### Not changed

This is entirely additive. Existing `DefaultEncoder` behaviour for `String`, `byte[]`, and `null` is untouched. No interface changes.

### Tests

- `DefaultEncoderTest` — unit tests verify `contentLength`, `isRepeatable`, and round-trip body content for all four new types.
- `FeignRequestBodyStreamingTest` (new) — end-to-end integration tests against `MockWebServer` confirm the encoded body arrives over the wire correctly for `File`, `Path`, `InputStream`, and `Request.Body`.

---

Closes #220

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### Version 14.0
+
+* `DefaultEncoder` now supports streaming request bodies for `File`, `Path`, `InputStream`, and `Request.Body` types,
+  avoiding in-memory buffering. New `Request.PathBody` and `Request.InputStreamBody` implementations are provided for
+  these cases.
+
 ### Version 13.12
 
 * `UrlencodedFormContentProcessor` now honors `CollectionFormat` from `@RequestLine`/`RequestTemplate` for array and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,12 @@
 
 * `DefaultEncoder` now supports streaming request bodies for `File`, `Path`, `InputStream`, and `Request.Body` types,
   avoiding in-memory buffering. New `Request.PathBody` and `Request.InputStreamBody` implementations are provided for
-  these cases.
+  these cases. (https://github.com/OpenFeign/feign/pull/3396)
 
 ### Version 13.12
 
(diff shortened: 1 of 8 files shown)


velo (approved):
Thanks for the focused addition. The streaming body support is additive, covered by both encoder-level and request-level tests, and I do not see compatibility or security concerns.

PR #3406, opened 2026-06-10
Fix feign-core test compilation broken by removed Request.create overload

seonwooj0810:
## What
`master` currently fails to build: several test classes still use APIs removed in f9159cfa ("prevent Content-Length smuggling, log format-string injection, and common-pool starvation in streaming bodies"), so test compilation fails with:

```
[ERROR] incompatible types: Charset cannot be converted to RequestTemplate
```

This breaks master's own snapshot builds (e.g. CircleCI 11938, 11953) and every PR branch rebased onto current master (e.g. #3382, CircleCI 12024/12030).

## How
Test-only changes, adapting the affected call sites to the current API (the CI reactor stops at the first failing module, so these surfaced one module at a time):

- **feign-core** — `JavaLoggerTest`, `LoggerRebufferTest`: switch 10 call sites from the removed `Request.create(HttpMethod, String, Map, byte[], Charset)` overload to the current `(Body, RequestTemplate)` signature with `null` body/template, matching the existing usage in `RetryableExceptionTest`. The logger tests only exercise response rebuffering, so the request body/charset is irrelevant to their assertions.
- **feign-slf4j** — `Slf4jLoggerTest`: same removed `Request.create` overload, plus a missing `feign.Util` import.
- **feign-validation / feign-validation-jakarta** — `BeanValidationMethodInterceptorTest`: replace the removed `RequestTemplate#body(String)` with `body(Request.Body.of(...))`.

I also grepped the whole repo for the remaining removed APIs (`RequestTemplate#body(String)`, `requestCharset()`, the `byte[]`-returning `body()`, `ThrowingConsumer`, `FeignBodyEntity`) — no further usages outside `src/main`.

No production code is touched.

### DIFF
--- a/core/src/test/java/feign/JavaLoggerTest.java
+++ b/core/src/test/java/feign/JavaLoggerTest.java
@@ -43,8 +43,7 @@ public class JavaLoggerTest {
         Response.builder()
             .status(200)
             .reason("OK")
-            .request(
-                Request.create(HttpMethod.GET, "/api", Collections.emptyMap(), null, Util.UTF_8))
+            .request(Request.create(HttpMethod.GET, "/api", Collections.emptyMap(), null, null))
             .headers(Collections.<String, Collection<String>>emptyMap())
             .body("{\"error\":\"test\"}", Util.UTF_8)
             .build();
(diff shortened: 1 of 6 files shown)


velo (changes requested):
The code changes are test-only and look compatible, but the required CircleCI pr-build check timed out before completing. Per project policy, this cannot be approved until the required check completes successfully. Please rerun CI and ensure it is green.

velo (approved):

PR #3407, opened 2026-06-10
Open observation scope and stop on non-FeignException errors

### LINKED ISSUE #2406: Feign observation span is not in "scope" so that loggers do not see its spanId
Despite new observation is started around feign call, MDC still contains "parent" spanId.

Code to reproduce:

https://github.com/norbson/feign-tracing-issue

run and call http://localhost:8080/test

Result:
Screenshot from 2024-05-05 18-04-03

### LINKED ISSUE #2566: [BUG] MicrometerObservationCapability not reporting timeouts

seonwooj0810:
## Summary

`MicrometerObservationCapability` had two related defects that left users without correct tracing or metrics when their feign-instrumented call did anything other than complete successfully or fail with a `FeignException`:

1. **The new observation was never made current** (#2406). The capability called `observation.start()` but never `observation.openScope()`. Any `ObservationHandler` that looks at the current observation — most notably the tracing handler that copies `traceId`/`spanId` into the MDC — saw the **parent** observation while the Feign call ran. Log lines emitted from inside the call were tagged with the wrong span id, breaking distributed-tracing log correlation.
2. **Only `FeignException` was caught** (#2566). The actual clients Feign delegates to throw client-specific types directly: JDK `HttpURLConnection` throws `SocketTimeoutException` / `ConnectException`, Apache HC5 throws `HttpHostConnectException`, OkHttp throws `IOException`. None of those are `FeignException`. They flew past the `catch`, the observation was never stopped, `observation.error(...)` was never called — so the timer was never committed, the trace stayed open, and one failed request quietly leaked an observation context.

Both fixes are small and live in the same handful of lines, so this PR addresses them together.

## What changed

In `MicrometerObservationCapability`:

- The synchronous and asynchronous `enrich` paths now open the observation as a `Scope` (`try (Observation.Scope scope = observation.openScope()) { … }`). The observation is the current observation while the underlying client runs, so MDC-propagating handlers see the Feign span — not the parent.
- The `catch` clause widened from `FeignException` to `Throwable`. Anything thrown by the underlying client is now captured by `observation.error(...)` and the observation is always stopped in `finally`. On the async path the same widening is applied to the synchronous portion of `execute` and to the `whenComplete` callback.
- The async path now unwraps `CompletionException` before recording, so the error stamped on the observation is the underlying cause rather than the executor wrapper.
- `finalizeObservation` collapsed into the inline logic now that the lifecycle is symmetrical.

## Why this shape

`Observation.Scope` is the documented mechanism in Micrometer for making an observation visible to downstream code — handlers, propagators, baggage. `start()` only marks the observation as in-flight; without `openScope()` the registry's `getCurrentObservation()` keeps returning the parent. Opening the scope inside a `try-with-resources` keeps the scope tied to the call boundary, so a thrown exception still closes it cleanly and the parent observation is correctly restored afterwards.

Catching `Throwable` rather than `Exception` is deliberate: the contract here is "the observation must be closed no matter what happens inside this block." `finally { observation.stop(); }` is what makes that true, and `catch (Throwable)` is what makes sure `observation.error(...)` gets the right cause attached before stop.

## Tests

Added `MicrometerObservationCapabilityTest` (uses `TestObservationRegistry` from `micrometer-observation-test`, already on the classpath via `micrometer-test`) covering:

- `scopeIsOpenDuringClientExecution` — `observationRegistry.getCurrentObservation()` returns the Feign observation while the underlying `Client` runs, and is `null` again afterwards. Locks in the fix for #2406.
- `recordsNonFeignExceptionThrownByClient` / `recordsRuntimeExceptionThrownByClient` — when the `Client` throws `SocketTimeoutException` / `RuntimeException`, the observation is stopped and the underlying error is recorded. Locks in #2566. (`Retryer.NEVER_RETRY` is used so the test asserts on the single observation produced, not on retry copies.)
- `asyncScopeIsOpenDuringClientExecution` — same for the async path.
- `asyncRecordsNonFeignExceptionFromFailedFuture` — a future completing exceptionally with `IOException` records the original `IOException` on the observation (not a `CompletionException` wrapper).
- `asyncRecordsSynchronousExceptionFromClient` — synchronous throws from inside `AsyncClient.execute` also stop the observation.
- `parentObservationIsRestoredAfterCall` — confirms the scope correctly returns to the caller's observation, so we aren't accidentally clobbering ambient tracing context.
- `scopeReceivesObservationCarryingFeignContext` — verifies the `FeignContext` is the context delivered to `ObservationHandler#onScopeOpened`, which is what tracing handlers actually read.

Existing tests still pass: `mvn -pl micrometer test` → 26 tests, 0 failures across `MicrometerCapabilityTest`, `MicrometerObservationRegistryCapabilityTest`, `FeignHeaderInstrumentationTest`, and the new class.

Fixes #2406
Fixes #2566

### DIFF
--- a/micrometer/src/main/java/feign/micrometer/MicrometerObservationCapability.java
+++ b/micrometer/src/main/java/feign/micrometer/MicrometerObservationCapability.java
@@ -18,10 +18,10 @@ package feign.micrometer;
 import feign.AsyncClient;
 import feign.Capability;
 import feign.Client;
-import feign.FeignException;
 import feign.Response;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
+import java.util.concurrent.CompletionException;
 
 /** Wrap feign {@link Client} with metrics. */
 public class MicrometerObservationCapability implements Capability {
@@ -55,13 +55,15 @@ public class MicrometerObservationCapability implements Capability {
                   this.observationRegistry)
               .start();
 
-      try {
+      try (Observation.Scope scope = observation.openScope()) {
         Response response = client.execute(request, options);
-        finalizeObservation(feignContext, observation, null, response);
+        feignContext.setResponse(response);
         return response;
-      } catch (FeignException ex) {
-        finalizeObservation(feignContext, observation, ex, null);
+      } catch (Throwable ex) {
+        observation.error(ex);
         throw ex;
+      } finally {
+        observation.stop();
       }
     };
   }
@@ -80,24 +82,29 @@ public class MicrometerObservationCapability implements Capability {
                   this.observationRegistry)
               .start();
 
-      try {
+      try (Observation.Scope scope = observation.openScope()) {
         return client
             .execute(feignContext.getCarrier(), options, context)
-            .whenComplete((r, ex) -> finalizeObservation(feignContext, observation, ex, r));
-      } catch (FeignException ex) {
-        finalizeObservation(feignContext, observation, ex, null);
-
+            .whenComplete(
+                (response, ex) -> {
+                  feignContext.setResponse(response);
+                  if (ex != null) {
+                    observation.error(unwrap(ex));
+                  }
+                  observation.stop();
+                });
+      } catch (Throwable ex) {
+        observation.error(ex);
+        observation.stop();
         throw ex;
       }
     };
   }
 
-  private void finalizeObservation(
-      FeignContext feignContext, Observation observation, Throwable ex, Response response) {
-    feignContext.setResponse(response);
-    if (ex != null) {
-      observation.error(ex);
+  private static Throwable unwrap(Throwable ex) {
+    if (ex instanceof CompletionException && ex.getCause() != null) {
+      return ex.getCause();
     }
-    observation.stop();
+    return ex;
   }
 }
--- a/micrometer/src/test/java/feign/micrometer/MicrometerObservationCapabilityTest.java
+++ b/micrometer/src/test/java/feign/micrometer/MicrometerObservationCapabilityTest.java
@@ -1,0 +1,302 @@
+/*
+ * Copyright © 2012 The Feign Authors (feign@commonhaus.dev)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feign.micrometer;
+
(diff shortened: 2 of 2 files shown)


velo (approved):
The production change is covered by focused synchronous and asynchronous observation tests, preserves the public API, and introduces no security concerns. Required checks are green. Thanks!

PR #3415, opened 2026-06-15
Return response body in FeignException from errorReading (#2618)

### LINKED ISSUE #2618: FeignException::responseBody() returns request body instead of response body

seonwooj0810:
Fixes #2618

## Problem
`FeignException.errorReading(...)` builds the exception by passing `request.body()` and `request.headers()` into the constructor parameters named `responseBody` / `responseHeaders`:

```java
return new FeignException(
    response.status(),
    format(...),
    request,
    cause,
    request.body(),     // <- responseBody slot
    request.headers()); // <- responseHeaders slot
```

As reported in #2618, this means that when an `IOException` is thrown while reading/decoding the response (see `InvocationContext.decode` and `ResponseHandler.handleResponse`), `FeignException.responseBody()` / `content()` / `contentUTF8()` / `responseHeaders()` return the **request** body and headers instead of the response. The maintainer confirmed in the thread that this looks like a bug.

## Fix
Read the response body into a `byte[]` and pass the **response** headers, mirroring the existing `errorStatus(...)` method. Reading is wrapped in a `try/catch (IOException ignored)` so a streamed or already-consumed body degrades gracefully to an empty body rather than throwing (consistent with `errorStatus`).

## Test evidence
Strengthened the existing `FeignExceptionTest.canCreateWithRequestAndResponse` to assert the *content*: the exception now carries the response body (`"response"`) rather than the request body (`"data"`), and exposes the response headers. Before the fix this assertion fails (`contentUTF8()` returned `"data"`).

```
./mvnw -pl core test -Dtest=FeignExceptionTest
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Verification done: (1) no in-flight PR (`gh pr list --search "responseBody"/"errorReading"` empty); (2) the only self-claim is a 2024-11-03 question that never produced a PR (>21 days stale); (3) code-focused change in `FeignException.java` + test; (4) confirmed the `request.body()` pattern still present on current master; (5) maintainer acknowledged the bug in-thread. Commit is DCO signed-off; google-java-format validation passes.

### DIFF
--- a/core/src/main/java/feign/FeignException.java
+++ b/core/src/main/java/feign/FeignException.java
@@ -177,13 +177,20 @@ public class FeignException extends RuntimeException {
   }
 
   static FeignException errorReading(Request request, Response response, IOException cause) {
+    byte[] body = {};
+    try {
+      if (response.body() != null) {
+        body = Util.toByteArray(response.body().asInputStream());
+      }
+    } catch (IOException ignored) { // NOPMD
+    }
     return new FeignException(
         response.status(),
         format("%s reading %s %s", cause.getMessage(), request.httpMethod(), request.url()),
         request,
         cause,
-        request.body(),
-        request.headers());
+        body,
+        response.headers());
   }
 
   public static FeignException errorStatus(String methodKey, Response response) {
--- a/core/src/test/java/feign/AsyncFeignTest.java
+++ b/core/src/test/java/feign/AsyncFeignTest.java
@@ -621,7 +621,8 @@ public class AsyncFeignTest {
     } catch (FeignException e) {
       assertThat(e.getMessage())
           .contains("timeout reading POST http://localhost:" + server.getPort() + "/");
-      assertThat(e.contentUTF8()).isEqualTo("Request body");
+      // After #2618 the FeignException carries the response body, not the request body.
+      assertThat(e.contentUTF8()).isEqualTo("success!");
       return;
     }
     fail("");
--- a/core/src/test/java/feign/FeignExceptionTest.java
+++ b/core/src/test/java/feign/FeignExceptionTest.java
@@ -43,16 +43,22 @@ class FeignExceptionTest {
             StandardCharsets.UTF_8,
             null);
 
+    Map<String, Collection<String>> responseHeaders =
+        Collections.singletonMap("content-type", Collections.singletonList("application/json"));
     Response response =
         Response.builder()
             .status(400)
             .body("response".getBytes(StandardCharsets.UTF_8))
+            .headers(responseHeaders)
             .request(request)
             .build();
 
     FeignException exception =
         FeignException.errorReading(request, response, new IOException("socket closed"));
     assertThat(exception.responseBody()).isNotEmpty();
+    // the exception must carry the response body, not the request body (gh-2618)
+    assertThat(exception.contentUTF8()).isEqualTo("response");
+    assertThat(exception.responseHeaders()).containsKeys("content-type");
     assertThat(exception.hasRequest()).isTrue();
     assertThat(exception.request()).isNotNull();
   }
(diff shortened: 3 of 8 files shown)


velo (changes requested):
The response-body/header fix in core/src/main/java/feign/FeignException.java has focused coverage in FeignExceptionTest, and I do not see a compatibility or security concern in the patch itself. I still cannot approve this head while the required ci/circleci: pr-build check is failing. Please get that check green, then I can re-review.

seonwooj0810:
Thanks @velo. The failing `ci/circleci: pr-build` was a pre-existing `feign-core` test-compilation error on `master` (`JavaLoggerTest`/`LoggerRebufferTest` calling the `Request.create(..., byte[], Charset)` overload that had been removed), not anything introduced by this PR — this change only touches `FeignException`. That error has since been fixed on `master` by #3406 (commit 18fe771), so I've rebased this branch onto the latest `master` to pick it up and retrigger CI (head now `a8e3ac0d`).

Verified locally on the rebased branch: `feign-core` test sources compile and `FeignExceptionTest` passes (14/14). Should be green now — ready for another look whenever you have a moment. Thanks!

velo (changes requested):
The current FeignException.errorReading change has a focused test for preserving the response body and headers, and I do not see a compatibility or security concern in the diff itself. I still cannot approve this head because the required ci/circleci: pr-build check is failing. Please get that check green, then I can take another pass.

seonwooj0810:
Thanks @velo. I tracked down the red `ci/circleci: pr-build`: it was three pre-existing tests that asserted the *old* behavior this PR fixes. `throwsFeignExceptionIncludingBody` / `throwsFeignExceptionWithoutBody` in `FeignTest`, `AsyncFeignTest` and `FeignUnderAsyncTest` expected `FeignException.contentUTF8()` from the `errorReading` path to return the **request** body — which is exactly the #2618 bug. Now that `errorReading` carries the **response** body, I updated those assertions to expect the response body (`"success!"`), and changed the without-body case to enqueue an empty response so it still verifies the empty-body path.

Full `feign-core` suite is green locally now: `./mvnw -pl core test` → `Tests run: 627, Failures: 0, Errors: 0`. Pushed as feb8c31a (DCO signed-off). Could you re-trigger the build when you get a chance?

velo (changes requested):
The added updates in core/src/test/java/feign/FeignTest.java, AsyncFeignTest.java, and FeignUnderAsyncTest.java are in the right area for the errorReading behavior, but the required ci/circleci: pr-build check is still failing on this head. I can't approve while that required check is red. Please get the branch back to green, then I'll re-review the current diff.

seonwooj0810:
Thanks @velo. Found the remaining red in `ci/circleci: pr-build`: `feign-core` was green, but the **async client modules** (okhttp, java11, hc5) still had their own `throwsFeignExceptionIncludingBody` asserting the pre-#2618 *request* body. After this fix the `FeignException` carries the *response* body, so those three assertions now expect the enqueued response (`success!`), matching what I already did in core.

Pushed `11c38764`. Targeted runs are green: `OkHttpClientAsyncTest` 42/0/0, `Http2ClientAsyncTest` 42/0/0, `AsyncApacheHttp5ClientTest` 43/0/0. Mind re-triggering the build?

velo (approved):

PR #3417, opened 2026-06-15
escape crlf and quotes in multipart Content-Disposition headers

alhudz:
Repro: send a multipart request whose `FormData.fileName` (or a `File` name / field name) contains `"` or a CRLF, e.g. `evil"\r\nX-Injected: 1`.
Cause: `AbstractWriter.writeFileMetadata` and `SingleParameterWriter` append the field name and file name straight into the quoted `Content-Disposition` parameters, so the value can close the quote and start a new header line or part.
Fix: percent-encode `\r`, `\n` and `"` in those parameters inside the writer, matching the WHATWG `multipart/form-data` encoding rules. Added `AbstractWriterTest` covering an injected file name and field name.

### DIFF
--- a/form/src/main/java/feign/form/multipart/AbstractWriter.java
+++ b/form/src/main/java/feign/form/multipart/AbstractWriter.java
@@ -64,10 +64,14 @@ public abstract class AbstractWriter implements Writer {
     val contentDespositionBuilder =
         new StringBuilder()
             .append("Content-Disposition: form-data; name=\"")
-            .append(name)
+            .append(escapeHeaderParameter(name))
             .append("\"");
     if (fileName != null) {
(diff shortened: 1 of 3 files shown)


yvasyliev:
This needs to be backported to https://github.com/OpenFeign/feign/pull/3414

yvasyliev (approved):

alhudz:
Backported to `14.x` in #3418 - same escaping on `AbstractPartEncoder.createHeaders`, with a regression test.

velo (approved):
This covers the non-streaming multipart writer path well: parameter names and filenames are escaped before they reach Content-Disposition, with regression tests for the injection cases. Checks are green and the change is compatible.

PR #3418, opened 2026-06-15
escape crlf and quotes in streaming multipart part headers

alhudz:
Backport of #3417 to the streaming `MultipartFormEncoder` added in #3414, as requested there.

Repro: send a streamed multipart request whose part `name` or `FormData` `filename` contains `"` or a CRLF, e.g. `evil"\r\nX-Injected: 1`.
Cause: `AbstractPartEncoder.createHeaders` appends the field name and file name straight into the quoted `Content-Disposition` parameters, so the value can close the quote and start a new header line or part. Same issue as the legacy writers fixed in #3417, on the new streaming path.
Fix: percent-encode `\r`, `\n` and `"` in those parameters (`escapeHeaderParameter`), matching the WHATWG `multipart/form-data` encoding rules. Added `AbstractPartEncoderTest` covering an injected name and file name.

### DIFF
--- a/form/src/main/java/feign/form/multipart/AbstractPartEncoder.java
+++ b/form/src/main/java/feign/form/multipart/AbstractPartEncoder.java
@@ -42,11 +42,15 @@ public abstract class AbstractPartEncoder implements PartEncoder {
     var disposition =
         new StringBuilder("form-data; name=")
             .append(DOUBLE_QUOTE)
-            .append(name)
+            .append(escapeHeaderParameter(name))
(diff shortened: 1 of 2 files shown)


yvasyliev (approved):

velo (approved):
Thanks for tightening the streaming multipart header generation. The escaping is covered for both field name and filename, it fixes a real header-injection risk, and I do not see a compatibility issue with the percent-encoded quoted parameter values.

PR #3419, opened 2026-06-16
Honor jdk.httpclient.allowRestrictedHeaders in Http2Client (#2975)

### LINKED ISSUE #2975: Http2Client doesn't allow setting Host header (or others restricted by default)
New Java HttpClient by default disalows setting some restricted headers - this can be disabled by using a system property jdk.httpclient.allowRestrictedHeaders with a lowercased header list.

I added such system property `-Djdk.httpclient.allowRestrictedHeaders=host`, but feign (logging that it sends my `Host` headere, still sends request with a different one (taken from the url, in my case it is `localhost:1234`).

When looking at the code of Http2Client I see Feign also keeps a list of disaloved headers that mimick the ones in jdk client - in that case it should also mimick the reading of `jdk.httpclient.allowRestrictedHeaders` and allow setting those that are on the list.

Relevant code from JDK:
```
   private static final Set<String> DISALLOWED_HEADERS_SET = getDisallowedHeaders();

    private static Set<String> getDisallowedHeaders() {
        Set<String> headers = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
        headers.addAll(Set.of("connection", "content-length", "expect", "host", "upgrade"));

        String v = getNetProperty("jdk.httpclient.allowRestrictedHeaders");
        if (v != null) {
            // any headers found are removed from set.
            String[] tokens = v.trim().split(",");
            for (String token : tokens) {
                headers.remove(token);
            }
            return Collections.unmodifiableSet(headers);
        } else {
            return Collections.unmodifiableSet(headers);
        }
    }
```
(Implemented in https://bugs.openjdk.org/browse/JDK-8213189 - it was backported to JDK 11)

seonwooj0810:
Fixes #2975

## Problem
The JDK `HttpClient` lets callers opt in to sending otherwise restricted headers (e.g. `Host`, `Connection`) by setting the `jdk.httpclient.allowRestrictedHeaders` system property. However `Http2Client` keeps its **own** hard-coded `DISALLOWED_HEADERS_SET` and strips those headers in `filterRestrictedHeaders(...)` *before* the request reaches the JDK client. As a result, even with `-Djdk.httpclient.allowRestrictedHeaders=host` the `Host` header is silently dropped and Feign sends the host derived from the URL instead.

## Change
`Http2Client` now mirrors `jdk.internal.net.http.common.Utils#getDisallowedHeaders()`: any header listed (comma-separated) in the `jdk.httpclient.allowRestrictedHeaders` system property is removed from the disallowed set, so opting in at the JDK level is no longer defeated by Feign's own filter. Default behavior (property unset) is unchanged — all of `connection, content-length, expect, host, upgrade` remain filtered.

The header-set construction was extracted into a package-private `disallowedHeaders(String)` method to keep it pure and unit-testable.

## Test evidence
Added `Http2ClientHeadersTest` covering the default set, single-header opt-in, and multi-header/case-insensitive/whitespace handling:

```
./mvnw -pl java11 -am test -Dtest=Http2ClientHeadersTest
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```
Code format validated via `git-code-format-maven-plugin:validate-code-format` (BUILD SUCCESS).

Verification done: confirmed bug pattern still on master (static `DISALLOWED_HEADERS_SET` strips headers unconditionally); no in-flight PR or linked branch for #2975; no self-claim in thread (maintainer invited a PR with tests on 2026-06-04); change is code-only in `.java`; commit carries DCO sign-off.

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -242,13 +242,26 @@ public class Http2Client implements Client, AsyncClient<Object> {
    *
    * @see jdk.internal.net.http.common.Utils.DISALLOWED_HEADERS_SET
    */
-  private static final Set<String> DISALLOWED_HEADERS_SET;
+  private static final Set<String> DISALLOWED_HEADERS_SET =
+      disallowedHeaders(System.getProperty("jdk.httpclient.allowRestrictedHeaders"));
 
-  static {
+  /**
+   * Builds the set of headers the underlying JDK HttpClient refuses to send. Mirrors {@code
+   * jdk.internal.net.http.common.Utils#getDisallowedHeaders()}: headers listed (comma separated) in
+   * the {@code jdk.httpclient.allowRestrictedHeaders} system property are removed from the set, so
+   * that callers who opt in at the JDK level (e.g. to set {@code Host}) are not silently filtered
+   * out here as well.
+   */
+  static Set<String> disallowedHeaders(String allowRestrictedHeaders) {
     // A case insensitive TreeSet of strings.
     final TreeSet<String> treeSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
     treeSet.addAll(Set.of("connection", "content-length", "expect", "host", "upgrade"));
-    DISALLOWED_HEADERS_SET = Collections.unmodifiableSet(treeSet);
+    if (allowRestrictedHeaders != null) {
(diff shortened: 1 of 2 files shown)


velo (approved):
This is a clean fix for the Http2Client restricted-header filtering. The new tests cover the default restricted set and the opt-in system property behavior, and the change stays internal to the Java 11 client without compatibility or security concerns.

PR #3422, opened 2026-06-17
Make Expressions max length configurable (fixes #2916)

### LINKED ISSUE #2916: MAX_EXPRESSION_LENGTH should not be fixed to 10000 but custom
feign.template.Expressions limit of MAX_EXPRESSION_LENGTH = 10000 added by commit 8d9fe72 should be configurable to a custom value and/or being able to be disabled by setting it to 0.

seonwooj0810:
Fixes #2916

### Problem
`feign.template.Expressions` hard-codes the maximum length of a single template expression to `MAX_EXPRESSION_LENGTH = 10000`. Any expression longer than that fails with `IllegalArgumentException: expression is too long. Max length: 10000`, with no way to raise or disable the limit. As noted in the issue, applications that legitimately need longer expressions have no escape hatch.

### Change
- The limit is now read from the `feign.template.expression.maxLength` system property, defaulting to `10000` (unchanged behavior).
- Setting the property to `0` (or any non-positive value) **disables** the length check entirely, as requested in the issue.

The constant was used in a single place (`Expressions.create`), so the change is self-contained and the default keeps existing behavior intact.

### Tests
Added `ExpressionsTest` cases:
- `tooLongExpressionFailsWithDefaultLimit` — default 10000 limit still rejects over-long expressions.
- `maxExpressionLengthIsConfigurable` — a custom limit is honored (`Max length: 5`).
- `lengthCheckCanBeDisabled` — with the property set to `0`, an expression beyond the default limit is accepted.

Verification done: built and ran `./mvnw -pl core -am test -Dtest=ExpressionsTest` (7/7 pass) and validated formatting with the project's `git-code-format-maven-plugin:validate-code-format`. All commits are DCO signed-off.

### DIFF
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -25,7 +25,14 @@ import java.util.regex.Pattern;
 
 public final class Expressions {
 
-  private static final int MAX_EXPRESSION_LENGTH = 10000;
+  /**
+   * System property controlling the maximum allowed length of a single expression. Defaults to
+   * {@link #DEFAULT_MAX_EXPRESSION_LENGTH}. Setting it to {@code 0} (or any non-positive value)
+   * disables the length check entirely.
+   */
+  static final String MAX_EXPRESSION_LENGTH_PROPERTY = "feign.template.expression.maxLength";
+
+  private static final int DEFAULT_MAX_EXPRESSION_LENGTH = 10000;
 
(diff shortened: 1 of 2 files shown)


velo (approved):
The default expression length guard stays unchanged, and the new system property is additive for users who need to tune or disable it explicitly. The ExpressionsTest additions cover the default, configured, and disabled cases, and the patch does not introduce a compatibility or security concern.

PR #3427, opened 2026-06-18
fall back to utf-8 for invalid content-type charset

alhudz:
Repro: a server response with `Content-Type: text/plain; charset=made-up-99` (or an illegal name such as `charset=@`) reaches `Response.charset()`, which is called by every JSON/text decoder (`jackson`, `json`, `fastjson2`, `jackson-jr`, `jackson3`) through `response.body().asReader(response.charset())`.
Expected: `charset()` returns its documented UTF-8 default.
Actual: `UnsupportedCharsetException` / `IllegalCharsetNameException` escapes the decoder on an otherwise valid response.
Cause: `Response.charset()` hands the charset token parsed out of the header straight to `Charset.forName` with no guard. `feign-form`'s `FormEncoder.getCharset` has the same unguarded call on the request `Content-Type`. `FeignException.getResponseCharset` already guards the identical parse.
Fix: catch `IllegalCharsetNameException`/`UnsupportedCharsetException` in both helpers and fall back to UTF-8. Regression tests cover both an illegal and an unsupported charset name.

### DIFF
--- a/core/src/main/java/feign/Response.java
+++ b/core/src/main/java/feign/Response.java
@@ -20,7 +20,9 @@ import static feign.Util.*;
 import feign.Request.ProtocolVersion;
 import java.io.*;
 import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.StandardCharsets;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.*;
 
 /** An immutable response to an http invocation which only returns string content. */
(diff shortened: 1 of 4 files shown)


velo (approved):
This is a focused fix: Response.charset() and FormEncoder now both fall back to UTF-8 for illegal or unsupported charset tokens, and the new tests cover the two failure paths in core and feign-form. I do not see a compatibility or security issue here.

PR #3429, opened 2026-06-19
generate multipart boundary from SecureRandom

alhudz:
Repro: `MultipartBoundaryTest` drives `MultipartFormContentProcessor.process` and reads the boundary back out of the generated `Content-Type` header; on the current code it equals `Long.toHexString(System.currentTimeMillis())` for the millisecond the request was built.
Cause: the boundary is derived from the wall clock, so it is low-entropy and guessable. RFC 2046 needs the boundary to never appear in a part, and against attacker-controlled body bytes the only guarantee is that the delimiter is unpredictable. A caller forwarding untrusted data into a field or filename value can embed `\r\n--<boundary>` and inject or truncate parts, which the `Content-Disposition` escaping in #3417 cannot cover because part bodies are written verbatim.
Fix: build the boundary from a shared `SecureRandom` as 128-bit lowercase hex.

### DIFF
--- a/form/src/main/java/feign/form/MultipartFormContentProcessor.java
+++ b/form/src/main/java/feign/form/MultipartFormContentProcessor.java
@@ -33,6 +33,7 @@ import feign.form.multipart.SingleParameterWriter;
 import feign.form.multipart.Writer;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.security.SecureRandom;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


velo (approved):
Good security hardening for multipart boundaries. The switch in form/src/main/java/feign/form/MultipartFormContentProcessor.java keeps the existing wire format simple while removing the predictable clock-derived boundary, and MultipartBoundaryTest covers the regression. CI is green and I do not see a compatibility issue here.

PR #3431, opened 2026-06-20
cap Content-Length before narrowing to int in Http2Client

alhudz:
1. `Http2Client.toFeignResponse` narrows the response `Content-Length` with `(int) length.getAsLong()` and no range check.
2. A server sending `Content-Length: 2147483648` (2^31) wraps to `-2147483648`, and `Content-Length: -1` is forwarded as-is.

`Response.Body.length()` is documented to be `null` when the length is unknown or greater than `Integer.MAX_VALUE`, and `okhttp`, `httpclient`, `hc5` and `googlehttpclient` already cap before narrowing. The wrapped negative length also slips past the `response.body().length() <= MAX_RESPONSE_BUFFER_SIZE` check in `InvocationContext` that decides whether to buffer the whole response into memory. Capped so the `int` is returned only for `0 <= length <= Integer.MAX_VALUE`, otherwise `null`.

`repro`: `Http2ClientContentLengthTest` builds an `HttpResponse` with a given `Content-Length` and reads `body().length()`.
- `Content-Length: 2147483648` expected `null`, before `-2147483648`.
- `Content-Length: -1` expected `null`, before `-1`.
- `Content-Length: 1024` stays `1024`.

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -130,6 +130,10 @@ public class Http2Client implements Client, AsyncClient<Object> {
 
   protected Response toFeignResponse(Request request, HttpResponse<InputStream> httpResponse) {
     final OptionalLong length = httpResponse.headers().firstValueAsLong("Content-Length");
+    final Integer contentLength =
+        length.isPresent() && length.getAsLong() >= 0 && length.getAsLong() <= Integer.MAX_VALUE
+            ? (int) length.getAsLong()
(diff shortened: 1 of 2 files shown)


velo (approved):
This fixes a real edge case in the HTTP/2 adapter without changing the API surface. The range check in java11/src/main/java/feign/http2client/Http2Client.java matches the Response.Body contract, and Http2ClientContentLengthTest covers the oversized, negative, and normal-length paths. Checks are green, so this looks good to me.

PR #3432, opened 2026-06-20
strip crlf from multipart Content-Type header

alhudz:
Repro: encode a `FormData` whose content type carries `CR`/`LF`, e.g. `new FormData("text/plain\r\nX-Injected: 1", "f.txt", bytes)`.
Cause: `AbstractWriter.writeFileMetadata` and `SingleParameterWriter.writeWithContentType` append the part `Content-Type` value verbatim, so a forwarded upload content type breaks out of the header and injects extra part headers. #3417 escaped the `Content-Disposition` `name`/`filename` parameters but left this header value alone.
Fix: strip `CR` and `LF` from the content type in the writers before it reaches the header. The value is unquoted so it cannot be percent-encoded like the quoted parameters without corrupting a legitimate media type.

### DIFF
--- a/form/src/main/java/feign/form/multipart/AbstractWriter.java
+++ b/form/src/main/java/feign/form/multipart/AbstractWriter.java
@@ -89,7 +89,7 @@ public abstract class AbstractWriter implements Writer {
             .append(contentDespositionBuilder.toString())
             .append(CRLF)
             .append("Content-Type: ")
-            .append(fileContentType)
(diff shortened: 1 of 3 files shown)


velo (approved):
This is a clean, focused fix for the multipart Content-Type header path. The production change is narrow, the new tests cover both FormData and single-parameter writers, and I do not see a compatibility or security downside in stripping CR/LF before writing the header.

PR #3433, opened 2026-06-21
encode soap request body with the configured charset

alhudz:
Repro: configure `SOAPEncoder` with `withCharsetEncoding(UTF-16)` and encode an object whose body holds a non-ASCII character (e.g. `Café`).
Expected: the request body bytes match the `UTF-16` declared in the XML prolog the marshaller wrote.
Actual: the body is mangled. `SOAPEncoder.encode` writes the message to a `ByteArrayOutputStream` in `charsetEncoding`, then the soap module does `new String(bos.toByteArray())` and soap-jakarta does `bos.toString()`. Both decode those bytes with the JVM platform default charset, and `RequestTemplate.body(String)` re-encodes as `UTF-8`, so the configured encoding is dropped and any non-ASCII content is corrupted (pure ASCII survives by accident, which is why the existing tests miss it).
Fix: hand the marshalled bytes to `template.body(bos.toByteArray(), charsetEncoding)` in both modules, the same way `UrlencodedFormContentProcessor` already passes bytes plus charset.

Regression test in both `SOAPCodecTest` files encodes non-ASCII content under `UTF-16`.

### DIFF
--- a/soap-jakarta/src/main/java/feign/soap/SOAPEncoder.java
+++ b/soap-jakarta/src/main/java/feign/soap/SOAPEncoder.java
@@ -131,7 +131,7 @@ public class SOAPEncoder implements Encoder {
       } else {
         soapMessage.writeTo(bos);
       }
-      template.body(bos.toString());
+      template.body(bos.toByteArray(), charsetEncoding);
     } catch (SOAPException
         | JAXBException
         | ParserConfigurationException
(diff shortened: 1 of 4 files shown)


velo (approved):
This is the right fix for the SOAP charset path: keep the marshalled bytes and pass the configured charset through to RequestTemplate. The UTF-16 non-ASCII tests in both SOAP modules cover the regression well.

PR #3436, opened 2026-06-22
encode collection values in urlencoded form bodies

alhudz:
Repro: post an `application/x-www-form-urlencoded` form whose value is a collection or array, e.g. a `tags` field of `["a&b=c", "d"]`. The body comes out as `tags=a&b=c&tags=d`, smuggling an extra `b=c` field into the request the downstream service parses.
Cause: `UrlencodedFormContentProcessor.createKeyValuePair` collects collection and array values with `Object::toString` and passes them to `CollectionFormat.join`, which encodes the field name and the separator but writes each value verbatim. The scalar path already runs values through `encode`, so only collection-valued fields leak reserved characters.
Fix: encode each collection value with the same `encode(value, charset)` helper before the join, matching the scalar path. `QueryTemplate` is the only other `CollectionFormat.join` caller and already passes pre-encoded values, so `join` itself is left untouched. Added regression tests for the array and collection paths.

### DIFF
--- a/form/src/main/java/feign/form/UrlencodedFormContentProcessor.java
+++ b/form/src/main/java/feign/form/UrlencodedFormContentProcessor.java
@@ -106,7 +106,10 @@ public class UrlencodedFormContentProcessor implements ContentProcessor {
   private CharSequence createKeyValuePair(
       CollectionFormat collectionFormat, String key, Stream<?> values, Charset charset) {
     val stringValues =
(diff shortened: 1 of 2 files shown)


velo (approved):
Good catch on the collection-valued form path. Encoding the values before handing them to CollectionFormat matches the scalar behavior, and the array/collection regressions cover the failure mode clearly.

Issue #2739, opened 2025-01-22
When using FeignClient, an error occurs when a value with a special structure is passed in the header map


tonyKe-Kazi:
feign-core:11.8
When the content of my header map is as follows, the following exception is throw
headerMap：
{
 "x-test":{"test":["123:861"]}
}
error:
java.util.regex.PatternSyntaxException: Unclosed character class near index 4
["123
    ^
	at java.base/java.util.regex.Pattern.error(Pattern.java:2204)
	at java.base/java.util.regex.Pattern.clazz(Pattern.java:2869)
	at java.base/java.util.regex.Pattern.sequence(Pattern.java:2315)
	at java.base/java.util.regex.Pattern.expr(Pattern.java:2245)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1945)
	at java.base/java.util.regex.Pattern.<init>(Pattern.java:1576)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1101)
	at feign.template.Expression.lambda$new$0(Expression.java:35)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at feign.template.Expression.<init>(Expression.java:35)
	at feign.template.Expressions$SimpleExpression.<init>(Expressions.java:106)
	at feign.template.Expressions.create(Expressions.java:86)
	at feign.template.Template.parseFragment(Template.java:218)
	at feign.template.Template.parseTemplate(Template.java:202)
	at feign.template.Template.<init>(Template.java:61)
	at feign.template.HeaderTemplate.<init>(HeaderTemplate.java:82)
	at feign.template.HeaderTemplate.create(HeaderTemplate.java:48)
	at feign.RequestTemplate.lambda$appendHeader$3(RequestTemplate.java:748)
	at java.base/java.util.TreeMap.callRemappingFunctionWithCheck(TreeMap.java:792)
	at java.base/java.util.TreeMap.compute(TreeMap.java:722)
	at feign.RequestTemplate.appendHeader(RequestTemplate.java:746)
	at feign.RequestTemplate.header(RequestTemplate.java:708)
	at feign.ReflectiveFeign$BuildTemplateByResolvingArgs.addHeaderMapHeaders(ReflectiveFeign.java:294)
	at feign.ReflectiveFeign$BuildTemplateByResolvingArgs.create(ReflectiveFeign.java:243)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:84)
[...]

PR #3437, opened 2026-06-22
treat invalid expression value modifier as a literal

alhudz:
Values supplied through `@HeaderMap` are parsed as URI templates, so a value that contains `{...:...}` is treated as an expression and the text after the `:` is compiled as a regular expression in `Expression`. A value like `{range:[1:10}` (brackets plus a colon, easy to hit with JSON-ish header data) is not a valid pattern, so `Pattern.compile` throws an uncaught `PatternSyntaxException` from `Expressions.create` and the whole request fails to build. Reported in #2739.

`Expressions.create` already returns `null` for chunks it cannot validate as an expression, and `Template.parseFragment` keeps those as literals. The fix catches `PatternSyntaxException` there and falls back to that same literal path, so a malformed value modifier is emitted verbatim instead of aborting expansion. Keeping the guard in the parser covers every template type (header, query, uri) and leaves valid `{name:regex}` expressions untouched.

### DIFF
--- a/core/src/main/java/feign/template/Expressions.java
+++ b/core/src/main/java/feign/template/Expressions.java
@@ -22,6 +22,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 public final class Expressions {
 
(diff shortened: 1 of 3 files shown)


velo (approved):
This keeps the parser behavior nicely contained: invalid regex modifiers now fall back to the existing literal path, and the tests cover both header-map expansion and valid modifiers. Looks good to me.

PR #3440, opened 2026-06-23
Revert "feat: create streamable multipart body encoder (#3414)" 

yvasyliev:
Reverts #3414 because it has both bad architecture & implementation. #3418 is also reverted transitively. Further discussion continues in #2734.

Target branch: `14.x`


velo (approved):
Reverting the streaming multipart work from the 14.x line is reasonable here given the architectural concerns, and CI is clean after removing the related docs, APIs, and tests. I’m fine with taking this out before it ships.

PR #3441, opened 2026-06-23
honour requested charset in okhttp response body reader

alhudz:
1. `OkHttpClient`'s response `Body.asReader(Charset)` ignores the `charset` argument and delegates to the deprecated `asReader()`, which returns okhttp's `ResponseBody.charStream()` (charset taken from `Content-Type`, defaulting to UTF-8).
2. decoders read a non-buffered response via `response.body().asReader(response.charset())`, so the resolved charset is dropped on the okhttp client only; `httpclient`, `hc5` and the core `InputStreamBody`/`ByteArrayBody` all decode with the requested charset.

Repro: a `text/plain` response (no `charset` parameter) whose body is `café` as ISO-8859-1 (`63 61 66 E9`), read via `asReader(ISO_8859_1)`.
Expected: `café`.
Actual: `caf�` (decoded as UTF-8).
Fix: decode with the requested charset using `new InputStreamReader(asInputStream(), charset)`, matching the other clients.

### DIFF
--- a/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
+++ b/okhttp/src/main/java/feign/okhttp/OkHttpClient.java
@@ -20,8 +20,10 @@ import static feign.Util.enumForName;
 import feign.AsyncClient;
 import feign.Client;
 import feign.Request.ProtocolVersion;
+import feign.Util;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

velo (approved):

PR #3442, opened 2026-06-23
encode string request body with utf-8 not the platform default

alhudz:
Repro: build a body with `Request.Body.create(String)` on a JVM whose default charset is not UTF-8 (`-Dfile.encoding=ISO-8859-1`) and send a non-ASCII payload such as `"é"`; the bytes on the wire are the ISO-8859-1 encoding, not the UTF-8 the rest of feign and the declared `Content-Type` assume.
Cause: `create(String)` encodes with `data.getBytes()`, the platform-default charset, and leaves the body `encoding` null so it is flagged binary. Every other body path defaults to UTF-8 (`RequestTemplate.body(String)` and the `create(String, Charset)` overload), so the result is deployment-dependent and the dropped charset means `asString()`/`charset()` cannot recover the text.
Fix: delegate to `create(data, UTF_8)` so the bytes are deterministic UTF-8 and the charset is recorded, matching the sibling overload.

### DIFF
--- a/core/src/main/java/feign/Request.java
+++ b/core/src/main/java/feign/Request.java
@@ -554,7 +554,7 @@ public final class Request implements Serializable {
     }
 
     public static Body create(String data) {
-      return new Body(data.getBytes());
+      return create(data, Util.UTF_8);
     }
 
     public static Body create(String data, Charset charset) {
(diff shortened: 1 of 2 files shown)


alhudz:
any update?

velo (approved):

PR #3444, opened 2026-06-24
strip crlf from resolved header template values

alhudz:
Repro: resolve a templated header (`@Headers("X-Custom: {value}")`, or a `@HeaderMap` entry) with `value` set to `legit\r\nX-Injected: evil`.
Expected: one `X-Custom` header.
Actual: the `CR`/`LF` survives into the header value, breaking the line and injecting an `X-Injected` request header.
Cause: `HeaderTemplate` expands header values with `EncodingOptions.NOT_REQUIRED`, so `expand` (templated values, via `RequestTemplate.resolve`) and `getValues` (`@HeaderMap` values added after `resolve` and read back when the `Request` is built) emit them verbatim. Query and URI templates pct-encode their values, so headers are the only un-encoded expansion path.
Fix: strip `\r` and `\n` from each value in both producers, matching the multipart `Content-Type` strip in #3432.

### DIFF
--- a/core/src/main/java/feign/template/HeaderTemplate.java
+++ b/core/src/main/java/feign/template/HeaderTemplate.java
@@ -141,7 +141,10 @@ public final class HeaderTemplate {
 
   public Collection<String> getValues() {
     return Collections.unmodifiableList(
-        this.values.stream().map(Template::toString).collect(Collectors.toList()));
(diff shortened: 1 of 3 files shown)


alhudz:
any update?

velo (approved):

PR #3445, opened 2026-06-24
clear response body length after gzip/deflate decode

alhudz:
Repro: return a response with `Content-Encoding: gzip` (or `deflate`) and a `Content-Length` (the compressed size); `response.body().length()` then reports that compressed length even though feign has already decompressed the body.
Cause: `Http2Client.toFeignResponse` and `DefaultClient.convertResponse` wrap the stream in `GZIPInputStream`/`InflaterInputStream` but keep the on-the-wire `Content-Length`, so the stale value no longer matches the readable body and still gates the `MAX_RESPONSE_BUFFER_SIZE` check in `InvocationContext.disconnectResponseBodyIfNeeded` that decides whether to buffer the whole body into a `byte[]`.
Fix: report the length as `null` once a decompressing wrapper is applied, since the decompressed length is unknown per the `Body.length()` contract. The other clients leave decompression to their underlying library, so only these two build a `Response` from a manually decompressed stream.

### DIFF
--- a/core/src/main/java/feign/DefaultClient.java
+++ b/core/src/main/java/feign/DefaultClient.java
@@ -122,8 +122,11 @@ public class DefaultClient implements Client {
     }
     if (stream != null && this.isGzip(headers.get(CONTENT_ENCODING))) {
       stream = new GZIPInputStream(stream);
+      // the body is now decompressed, the Content-Length described the compressed bytes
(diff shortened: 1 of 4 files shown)


alhudz:
any update?

velo (approved):

PR #3446, opened 2026-06-24
Let capabilities supply response interceptors

### LINKED ISSUE #2935: Response interceptor not getting picked from Capability
Hi,

We recently upgraded our Feign client from version 12.5 to 13.5 and have encountered an issue post-upgrade. Specifically, the method `public ResponseInterceptor enrich(ResponseInterceptor responseInterceptor)` within our custom capability is no longer being executed and hence the response interceptor is not getting used at all. This method was functioning as expected prior to the upgrade.

I suspect this has something to do with the changes introduced as part of supporting multiple ResponseInterceptors (https://github.com/OpenFeign/feign/pull/1829), but not sure. Please provide some insights or guidance on resolving this issue.

Attaching the code for reference:
```
@Slf4j
public class MyInterceptor implements ResponseInterceptor, Retryer {

    @Override
    public Object intercept(InvocationContext invocationContext, Chain chain) throws Exception {
        try (Response ignored = invocationContext.response()) {
            log.info("Inside intercept method of MyInterceptor");
            return chain.next(invocationContext);
        }
    }

    @Override
    public void continueOrPropagate(RetryableException e) {
        log.info("Inside continueOrPropagate method of MyInterceptor");
        throw e;
    }

    @Override
    public Retryer clone() {
        return this;
    }
}
```
```
public class MyCapability implements Capability {

    private final MyInterceptor myInterceptor;

    public MyCapability() {
        this.myInterceptor = new MyInterceptor();
    }

    public ResponseInterceptor enrich(ResponseInterceptor responseInterceptor) {
        return myInterceptor;
    }

    public Retryer enrich(Retryer retryer) {
        return myInterceptor;
    }
}
```
```
public class MyConfig {

    @Bean
    public MyCapability myCapability() {
        return new MyCapability();
    }
}
```
```
@FeignClient(name = "myClient", url = "${myUrl}", configuration = {MyConfig.class})
public interface PartyDetailsClient {

    @PostMapping
    MyResponse getPartyDetails(@RequestBody MyRequest myRequest);
}

```

hutiefang76:
## Summary

`Capability.enrich(ResponseInterceptor)` was only called for configured response interceptors. If a capability wanted to supply the response interceptor itself, there was no component for it to enrich, so the interceptor was never installed.

This keeps the existing per-interceptor enrichment path, and adds a default pass-through response interceptor only for the empty-list case. If a capability replaces that default interceptor, the replacement is installed before the response interceptor list is enriched as a whole.

## Tests

- `./mvnw -pl core -Dtoolchain.skip=true -Dtest=feign.BaseBuilderTest#capabilityCanProvideResponseInterceptorWhenNoneConfigured test`
- `./mvnw -pl core -Dtoolchain.skip=true -Dtest=feign.BaseBuilderTest test`
- `./mvnw -pl core -Dtoolchain.skip=true test`
- `git diff --check`

Fixes #2935

### DIFF
--- a/core/src/main/java/feign/BaseBuilder.java
+++ b/core/src/main/java/feign/BaseBuilder.java
@@ -316,10 +316,21 @@ public abstract class BaseBuilder<B extends BaseBuilder<B, T>, T> implements Clo
       // enrich each response interceptor, then enrich the list as a whole
       ResponseInterceptor[] responseArray =
           clone.responseInterceptors.toArray(new ResponseInterceptor[0]);
-      for (int i = 0; i < responseArray.length; i++) {
-        responseArray[i] =
+      if (responseArray.length == 0) {
+        ResponseInterceptor defaultResponseInterceptor = (context, chain) -> chain.next(context);
+        ResponseInterceptor enrichedResponseInterceptor =
             (ResponseInterceptor)
-                Capability.enrich(responseArray[i], ResponseInterceptor.class, capabilities);
+                Capability.enrich(
+                    defaultResponseInterceptor, ResponseInterceptor.class, capabilities);
+        if (enrichedResponseInterceptor != defaultResponseInterceptor) {
+          responseArray = new ResponseInterceptor[] {enrichedResponseInterceptor};
+        }
+      } else {
+        for (int i = 0; i < responseArray.length; i++) {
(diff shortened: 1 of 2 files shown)


velo (approved):

PR #3447, opened 2026-06-24
honour response charset in gson and jackson iterator decoders

alhudz:
Repro: a server response with `Content-Type: application/json; charset=ISO-8859-1` and an `ISO-8859-1` encoded body decoded through `GsonDecoder`, `JacksonIteratorDecoder` or `Jackson3IteratorDecoder`.
Cause: those three read the body with `asReader(UTF_8)`, dropping `response.charset()`, so non-ASCII bytes come back as `U+FFFD`. The sibling `JacksonDecoder`, `Jackson3Decoder`, `Fastjson2Decoder`, `JacksonJrDecoder` and `JsonDecoder` already pass `response.charset()`.
Fix: pass `response.charset()` in the three decoders, matching the siblings. `response.charset()` already defaults to UTF-8, so plain UTF-8 responses are unchanged. Regression test added per module.

### DIFF
--- a/gson/src/main/java/feign/gson/GsonDecoder.java
+++ b/gson/src/main/java/feign/gson/GsonDecoder.java
@@ -15,7 +15,6 @@
  */
 package feign.gson;
 
-import static feign.Util.UTF_8;
 import static feign.Util.ensureClosed;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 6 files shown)


alhudz:
gentle ping

velo (approved):

PR #3448, opened 2026-06-24
Fix duplicate Content-Length handling in DefaultClient

### LINKED ISSUE #2862: feign.core v13.2.1 Repeatedly writing "Content-Length"
I have two applications, A and B. B invokes A through the interface exposed by A. However, an issue occurred where A failed to decode the HTTP message on the Tomcat side, resulting in an HTTP status code of 400. Through source code analysis, I found that in the `feign.Client.Default#convertAndSend` method, there is a section that adds request headers. When gzip or deflate compression of the request body is not enabled, "content-length" is written once, but it is written again later. See lines 179 to 198 of the `feign.Client` class. 
The `feign.RequestTemplate#body(feign.Request.Body)` will calculate the length of the byte array and add it to the header. In `tomcat-embed-core (v9.0.46)`, there is such a processing in `org.apache.tomcat.util.http.MimeHeaders#getUniqueValue`. If the existing header information is obtained, an exception will be thrown. 

My current solutions are as follows:
1. Inject a feign.RequestInterceptor instance to remove duplicate header information.
2. Enable compression and manually calculate the request body length, but this involves modifications to other applications.

hutiefang76:
Fixes #2862.

`RequestTemplate.body(...)` records the request body length as a `Content-Length` header. `DefaultClient.convertAndSend(...)` used that value in two ways for non-compressed request bodies: it kept the length for fixed-length streaming mode and also added it as a normal request property. With `HttpURLConnection`, the body/output path can manage the request length itself, so adding the header manually can produce duplicate `Content-Length` values when restricted headers are allowed.

This change keeps the parsed content length for the existing `disableRequestBuffering` / fixed-length streaming path, but stops adding `Content-Length` through `addRequestProperty(...)` for request bodies.

Tests:

```bash
mvn -pl core -Dtoolchain.skip=true -Dtest=ClientTest#testConvertAndSendDoesNotAddContentLengthHeaderForBody test
mvn -pl core -Dtoolchain.skip=true -Dtest=ClientTest test
mvn -pl core -Dtoolchain.skip=true -Dtest=DefaultClientTest test
mvn -pl core -Dtoolchain.skip=true test
git diff --check
```

Note: I used `-Dtoolchain.skip=true` locally because this machine does not have a Maven JDK 8 toolchain configured; the core module still compiled and ran its test suite successfully.

### DIFF
--- a/core/src/main/java/feign/DefaultClient.java
+++ b/core/src/main/java/feign/DefaultClient.java
@@ -170,7 +170,6 @@ public class DefaultClient implements Client {
         if (field.equals(CONTENT_LENGTH)) {
           if (!gzipEncodedRequest && !deflateEncodedRequest) {
             contentLength = Integer.valueOf(value);
-            connection.addRequestProperty(field, value);
           }
         }
         // Avoid add "Accept-encoding" twice or more when "compression" option is enabled
--- a/core/src/test/java/feign/ClientTest.java
+++ b/core/src/test/java/feign/ClientTest.java
@@ -17,12 +17,19 @@ package feign;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.Mockito.anyString;
+import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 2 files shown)


velo (approved):

PR #3449, opened 2026-06-24
Fix fluent methods on cloned builders

### LINKED ISSUE #2782: BaseBuilder not cloned properly

hutiefang76:
Fixes #2782.

`BaseBuilder` cached `this` in a final `thisB` field. When a builder is cloned, that final field is copied as-is, so fluent methods invoked on the clone still return the original builder instance. This can make chained configuration on cloned builders silently continue on the wrong object.

This keeps the existing self-type helper pattern but makes it resolve `this` dynamically through one private helper method instead of storing it in a clone-stale field. That avoids the broader refactor from #2783 and keeps the unchecked cast centralized.

Validation:

```bash
mvn -pl core -Dtoolchain.skip=true -Dtest=BaseBuilderTest#clonedBuilderFluentMethodsReturnClone test
mvn -pl core -Dtoolchain.skip=true -Dtest=BaseBuilderTest test
git diff --check
mvn -pl core -Dtoolchain.skip=true test
```

The full `core` test run passed 635 tests with 0 failures/errors and 3 skipped tests.

### DIFF
--- a/core/src/main/java/feign/BaseBuilder.java
+++ b/core/src/main/java/feign/BaseBuilder.java
@@ -40,8 +40,6 @@ import java.util.stream.Collectors;
 
 public abstract class BaseBuilder<B extends BaseBuilder<B, T>, T> implements Cloneable {
 
-  private final B thisB;
-
   protected List<RequestInterceptor> requestInterceptors = new ArrayList<>();
   protected List<ResponseInterceptor> responseInterceptors = new ArrayList<>();
   protected List<MethodInterceptor> methodInterceptors = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


velo (approved):

PR #3453, opened 2026-06-26
honour response charset in StringDecoder

alhudz:
Repro: a `String`-returning method against a server that replies `Content-Type: text/plain; charset=ISO-8859-1` with body byte `0xE9`.
Expected: `é`.
Actual: `U+FFFD`, because the byte is run through a UTF-8 reader.
Cause: `StringDecoder.decode` reads the body with `asReader(Util.UTF_8)`, so the charset `response.charset()` already resolves from `Content-Type` is dropped. `DefaultDecoder` extends it, so this is the default behaviour for every non-UTF-8 `String` response.
Fix: decode with `response.charset()`, which already falls back to UTF-8 when none is declared, matching the JSON decoders (`jackson`, `json`, `fastjson2`, `jackson-jr`).

### DIFF
--- a/core/src/main/java/feign/codec/StringDecoder.java
+++ b/core/src/main/java/feign/codec/StringDecoder.java
@@ -31,7 +31,7 @@ public class StringDecoder implements Decoder {
       return null;
     }
     if (String.class.equals(type)) {
-      return Util.toString(body.asReader(Util.UTF_8));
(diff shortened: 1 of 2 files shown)


alhudz:
any update?

velo (approved):

PR #3454, opened 2026-06-27
Add HTTP QUERY method support (RFC 10008)

kumarprabhashanand:
RFC 10008 defines QUERY as a safe, idempotent, cacheable HTTP method that carries a request body describing query criteria - a body carrying counterpart to GET.

### Core

- `Request.HttpMethod.QUERY(true)` registered as a first-class method
- `RequestLine` Javadoc updated to list QUERY as a valid method name

### Mock client

- `feign.mock.HttpMethod` gains a `QUERY` constant; without it, `RequestKey.create(Request)` threw `IllegalArgumentException` on any QUERY request routed through `MockClient`
- `MockClientTest` adds `queryMock()` to verify end-to-end routing

### Client tests

- `AbstractClientTest` adds a `query()` test verifying that QUERY transmits a body with the correct `Content-Type` and `Content-Length`
- `DefaultClientTest` overrides `query()` to assert `RetryableException` wrapping `ProtocolException` - `HttpURLConnection.setRequestMethod` does not accept QUERY (mirrors the existing PATCH pattern)
- `GoogleHttpClientTest` and `AbstractJAXRSClientTest` override `query()` for the same reason (both delegate to `HttpURLConnection` under the hood)

### HTTP cache

- `HttpCacheInterceptor` includes QUERY in its default cacheable set
- Cache key incorporates `Arrays.hashCode(body)` when a request body is present to reduce cross-body collisions
- `HttpCacheInterceptorTest` verifies that distinct bodies produce separate cache entries and that same-body repeats trigger conditional revalidation

**Out of scope:** `Accept-Query` response header (RFC 10008 §6) is not implemented; servers indicate QUERY support via `Allow` already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### Version 13.14
+
+* Add support for the HTTP QUERY method (RFC 10008) — safe, idempotent, and cacheable with a
+  request body. `HttpCacheInterceptor` includes QUERY in its default cacheable set and
+  incorporates a body hash into the cache key to reduce cross-body collisions.
+
 ### Version 13.12
 
 * `UrlencodedFormContentProcessor` now honors `CollectionFormat` from `@RequestLine`/`RequestTemplate` for array and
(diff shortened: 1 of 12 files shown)


velo (approved):

PR #3462, opened 2026-07-01
Enforce Java 11 response body timeouts

### LINKED ISSUE #3068: Http2Client with native HttpClient does not enforce response correctly timeout
There is a flaw in Java's interface design that makes it very easy to introduce bug where application stalls when using native HttpClient. Request timeout given applies only to point when response headers are received. Timeout does not apply to receiving body.

Official jdk issue: https://bugs.openjdk.org/browse/JDK-8208693

Easiest way to address the problem is to use async sending with `orTimeout`. The far more complex alternative is to create customer body handler / subscriber that is able to handle timeout correctly.

Considering that writing customer body handler is a non-trivial task and that orTimeout does address the problem, I would recommend changing feign implementation to use asyncSend. Java's internal synchronous implementation is a delegate to async call, so change to async would not change threading behaviour.

goutamadwant:
## Summary

This updates the Java 11 `Http2Client` response handling so the configured Feign read timeout also applies while the response body is being read.

The existing request timeout still covers waiting for the response headers, but `BodyHandlers.ofInputStream()` can hand back a response before the body has arrived. The response body stream is now wrapped with a read-timeout guard, preserving the existing streaming response behavior while failing stalled body reads with `HttpTimeoutException`.

## Tests

- `mvn --batch-mode -Dtoolchain.skip=true -pl java11 -am -Dtest=Http2ClientTest#timeoutReadingResponseBody,Http2ClientAsyncTest#timeoutReadingResponseBody -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn --batch-mode -Dtoolchain.skip=true -pl java11 -am -Dtest=Http2ClientTest,Http2ClientAsyncTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn --batch-mode -Dtoolchain.skip=true -pl java11 -am -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn --batch-mode -Dtoolchain.skip=true -pl java11 -am -DskipTests validate`
- `mvn --batch-mode -Dtoolchain.skip=true -pl java11 -am git-code-format:validate-code-format`

Fixes #3068

### DIFF
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -38,6 +38,7 @@ import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpRequest.Builder;
 import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
+import java.net.http.HttpTimeoutException;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collection;
@@ -52,12 +53,25 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
 
 public class Http2Client implements Client, AsyncClient<Object> {
+
+  private static final ScheduledExecutorService BODY_READ_TIMEOUT_EXECUTOR =
+      Executors.newSingleThreadScheduledExecutor(
+          runnable -> {
+            Thread thread = new Thread(runnable, "feign-http2client-body-timeout");
+            thread.setDaemon(true);
+            return thread;
+          });
 
   private final HttpClient client;
 
@@ -109,7 +123,7 @@ public class Http2Client implements Client, AsyncClient<Object> {
       throw new IOException(e);
     }
 
-    return toFeignResponse(request, httpResponse);
+    return toFeignResponse(request, httpResponse, options);
   }
 
   @Override
@@ -125,17 +139,22 @@ public class Http2Client implements Client, AsyncClient<Object> {
     HttpClient clientForRequest = getOrCreateClient(options);
     CompletableFuture<HttpResponse<InputStream>> future =
         clientForRequest.sendAsync(httpRequest, HttpResponse.BodyHandlers.ofInputStream());
-    return future.thenApply(httpResponse -> toFeignResponse(request, httpResponse));
+    return future.thenApply(httpResponse -> toFeignResponse(request, httpResponse, options));
   }
 
   protected Response toFeignResponse(Request request, HttpResponse<InputStream> httpResponse) {
+    return toFeignResponse(request, httpResponse, null);
+  }
+
+  private Response toFeignResponse(
+      Request request, HttpResponse<InputStream> httpResponse, Options options) {
     final OptionalLong length = httpResponse.headers().firstValueAsLong("Content-Length");
     final Integer contentLength =
         length.isPresent() && length.getAsLong() >= 0 && length.getAsLong() <= Integer.MAX_VALUE
             ? (int) length.getAsLong()
             : null;
 
-    InputStream body = httpResponse.body();
+    InputStream body = withReadTimeout(httpResponse.body(), options);
 
     if (httpResponse.headers().allValues(CONTENT_ENCODING).contains(ENCODING_GZIP)) {
       try {
@@ -154,6 +173,110 @@ public class Http2Client implements Client, AsyncClient<Object> {
         .status(httpResponse.statusCode())
         .headers(castMapCollectType(httpResponse.headers().map()))
         .build();
+  }
+
+  private static InputStream withReadTimeout(InputStream body, Options options) {
+    if (body == null || options == null || options.readTimeout() <= 0) {
+      return body;
+    }
+    return new TimeoutInputStream(body, options.readTimeout(), options.readTimeoutUnit());
+  }
+
+  private static final class TimeoutInputStream extends InputStream {
+
+    private final InputStream delegate;
+    private final long timeout;
+    private final TimeUnit timeoutUnit;
+
+    private TimeoutInputStream(InputStream delegate, long timeout, TimeUnit timeoutUnit) {
+      this.delegate = delegate;
+      this.timeout = timeout;
+      this.timeoutUnit = timeoutUnit;
+    }
+
+    @Override
+    public int read() throws IOException {
+      return readWithTimeout(delegate::read);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+      return readWithTimeout(() -> delegate.read(b, off, len));
+    }
+
+    @Override
+    public int available() throws IOException {
+      return delegate.available();
+    }
+
+    @Override
+    public void close() throws IOException {
+      delegate.close();
+    }
+
+    private int readWithTimeout(BodyRead read) throws IOException {
+      final AtomicBoolean completed = new AtomicBoolean(false);
+      final AtomicBoolean timedOut = new AtomicBoolean(false);
+      final ScheduledFuture<?> timeoutFuture =
+          BODY_READ_TIMEOUT_EXECUTOR.schedule(
+              () -> {
+                if (completed.compareAndSet(false, true)) {
+                  timedOut.set(true);
+                  try {
+                    delegate.close();
+                  } catch (IOException ignored) {
+                  }
+                }
+              },
+              timeout,
+              timeoutUnit);
+
+      try {
+        final int result = read.read();
+        if (completed.compareAndSet(false, true)) {
+          timeoutFuture.cancel(false);
+          return result;
+        }
+        throw timeoutException(null);
+      } catch (IOException e) {
+        if (completed.compareAndSet(false, true)) {
+          timeoutFuture.cancel(false);
+        }
+        final HttpTimeoutException timeoutException = findTimeoutException(e);
+        if (timedOut.get() || timeoutException != null) {
+          throw timeoutException == null ? timeoutException(e) : timeoutException;
+        }
+        throw e;
+      } catch (RuntimeException e) {
+        if (completed.compareAndSet(false, true)) {
+          timeoutFuture.cancel(false);
+        }
+        throw e;
+      }
+    }
+
+    private static HttpTimeoutException timeoutException(IOException cause) {
+      final HttpTimeoutException exception = new HttpTimeoutException("response timed out");
+      if (cause != null) {
+        exception.initCause(cause);
+      }
+      return exception;
+    }
+
+    private static HttpTimeoutException findTimeoutException(Throwable throwable) {
+      Throwable current = throwable;
+      while (current != null) {
+        if (current instanceof HttpTimeoutException) {
+          return (HttpTimeoutException) current;
+        }
+        current = current.getCause();
+      }
+      return null;
+    }
+  }
+
+  private interface BodyRead {
+    int read() throws IOException;
   }
 
   private HttpClient getOrCreateClient(Options options) {
--- a/java11/src/main/java/feign/http2client/Http2Client.java
+++ b/java11/src/main/java/feign/http2client/Http2Client.java
@@ -32,6 +32,7 @@ import java.net.URISyntaxException;
 import java.net.http.HttpClient;
 import java.net.http.HttpClient.Redirect;
 import java.net.http.HttpClient.Version;
+import java.net.http.HttpHeaders;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublisher;
 import java.net.http.HttpRequest.BodyPublishers;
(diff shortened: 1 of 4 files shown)


velo:
This feels like it belong on a interceptor or capability.... it is an orthogonal feature that we could apply to all clients.

velo (commented):
Good problem to solve — the Java 11 client genuinely ignores `readTimeout` once headers arrive, and I like that you covered both the sync and async paths with tests. Two things need fixing before this can go in.

### 1. Subclass overrides of `toFeignResponse` are silently bypassed

`execute()` now calls the new **private** three-arg `toFeignResponse(Request, HttpResponse, Options)`. The existing `protected Response toFeignResponse(Request, HttpResponse)` is left in place but is no longer on the execution path — so any subclass overriding it (a documented extension point today) keeps compiling and keeps passing its own tests, while its code stops being called in production. That's the worst kind of breaking change: invisible.

`Http2ClientContentLengthTest` hides this, because it calls the two-arg form directly and therefore exercises a path production no longer uses.

Please make the three-arg overload `protected` and reduce the two-arg one to a delegating shim (`return toFeignResponse(request, httpResponse, null);`), so overriding either signature still works.

### 2. One scheduled task per `read()` call

`readWithTimeout` schedules a `ScheduledFuture` and cancels it on **every** invocation, including the single-byte `read()`. For a consumer reading unbuffered, that's a schedule+cancel round-trip on a shared single-threaded executor per byte. Two allocations of `AtomicBoolean` per byte on top.

Schedule one deadline for the stream (or at least per `read(byte[], int, int)` batch) and check remaining time, rather than arming a fresh timer per read.

### Smaller notes

- `TimeoutInputStream` doesn't delegate `skip`, `mark`, `reset`, `markSupported`. `InputStream#skip` routes through `read`, so it's correct, but `markSupported` silently degrades to `false` for a delegate that supported it.
- `BODY_READ_TIMEOUT_EXECUTOR` is a process-wide static that's never shut down. Daemon thread makes that survivable, but it means every `Http2Client` in the JVM shares one timer thread — worth a comment at minimum.
- In `readWithTimeout`, the `throw timeoutException == null ? timeoutException(e) : timeoutException;` line reads awkwardly. `if (timeoutException != null) throw timeoutException; throw timeoutException(e);` is clearer.
- Please make sure the new tests use `Retryer.NEVER_RETRY` consistently — the async one doesn't set it, so a retry would multiply the 1s body delay.

goutamadwant:
Thanks for your review @velo. Addressed all your comments and suggestions. Can you pls take a look again at the changes? thanks!

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -52,13 +54,27 @@
...
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
+import javax.net.ssl.SSLSession;
 
 public class Http2Client implements Client, AsyncClient<Object> {
 
+  // Shared by all clients so response-body deadlines do not create one thread per client.
+  private static final ScheduledExecutorService BODY_READ_TIMEOUT_EXECUTOR =
```
velo, line 70: Two problems with this executor.

**Cancelled tasks are never removed from the queue.** `ScheduledThreadPoolExecutor` defaults `removeOnCancelPolicy` to `false`, so a task cancelled in `close()`/at EOF stays in the `DelayedWorkQueue` until its *original* deadline elapses. Every response arms one of these, and the happy path cancels essentially all of them. At 1k req/s with a 60s `readTimeout` that's ~60k dead queue nodes resident at steady state, each holding a strong reference to a response `InputStream` and its buffers. This is the one I'd consider a blocker.

Fix: build it as a `ScheduledThreadPoolExecutor` and call `setRemoveOnCancelPolicy(true)`.

**Static, never shut down.** The thread is created lazily on first `schedule()`, inheriting the context classloader of whichever caller happens to trigger it — the classic webapp/OSGi classloader pin on undeploy. Worth at least making it lazily initialised behind a holder so the class doesn't own a live thread for the lifetime of the JVM.

(`CompletableFuture.delayedExecutor` would avoid creating a thread at all, but note it gives you no handle to cancel with, so you'd trade the queue-retention problem for a different shape of it. The `removeOnCancelPolicy` fix is the simpler one.)

velo (commented):
Thanks for picking this up — #3068 is a real gap and the guard-stream idea is a reasonable way in. But I have two design-level objections plus a handful of concrete issues before this can go in.

**1. This silently redefines what `readTimeout` means.** Everywhere else in Feign — including `Client.Default`, where it maps to `SO_TIMEOUT` — `readTimeout` is an *idle* timeout: time between reads. Here it becomes a hard deadline on the entire body transfer. A 30s `readTimeout` now kills a perfectly healthy 31s download, and streaming responses (users returning `feign.Response` and consuming `body().asInputStream()` themselves) stop working with any finite `readTimeout`.

I know the issue suggests `orTimeout`, which has the same semantics, and explicitly calls the correct fix (a custom body subscriber) non-trivial. I'm not necessarily against shipping the deadline version — but it has to be a deliberate, documented decision, not a side effect. At minimum: javadoc on the behaviour, a note in the changelog, and ideally a way to opt out.

**2. The clock starts at the wrong moment.** The timer is armed when the `HttpResponse` object is constructed, not when the caller starts reading. In the async path `thenApply(...)` runs on the HttpClient executor, so the response can sit through the decoder/retryer before a single byte is read. Under load you get spurious timeouts on responses that never stalled — the opposite of what the option is for.

Inline comments cover the rest: an unbounded cancelled-task queue on a shared static scheduler, and a restructuring that would delete ~90 lines of this PR.

Marking this as a comment rather than request-changes, but points 1, 2 and the `setRemoveOnCancelPolicy` one all need resolving before merge.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -159,6 +175,194 @@ protected Response toFeignResponse(Request request, HttpResponse<InputStream> ht
         .build();
   }
 
+  protected Response toFeignResponse(
```
velo, line 178: This new `protected` overload, `TimeoutHttpResponse` below, and the two `*ArgumentOverrideClient` tests all exist for one reason: you're wrapping the *JDK* `HttpResponse` so the existing 2-arg `toFeignResponse` can be reused unchanged.

Wrap the *Feign* `Response` instead and all of it goes away:

```java
Response response = toFeignResponse(request, httpResponse);
return withReadTimeout(response, options);
// withReadTimeout: response.toBuilder().body(wrapped, response.body().length())
```

That's no new protected API to maintain, no 60-line `HttpResponse` mirror, no tests needed to prove the overload plumbing dispatches correctly — and as a bonus the guard sits *outside* the gzip/inflate layer, so it fires on the outermost read rather than on the compressed bytes.

Also: this is the point where the deadline is armed, which is my objection (2) in the review body — for the async path this runs on the HttpClient executor, potentially well before the caller reads anything.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -159,6 +175,194 @@ protected Response toFeignResponse(Request request, HttpResponse<InputStream> ht
...
+  }
+
+  private static InputStream withReadTimeout(InputStream body, Options options) {
+    if (body == null || options == null || options.readTimeout() <= 0) {
+      return body;
+    }
+    return new TimeoutInputStream(body, options.readTimeout(), options.readTimeoutUnit());
+  }
+
+  private static final class TimeoutInputStream extends InputStream {
```
velo, line 194: For the record, the semantics encoded here: a single `schedule()` at construction that fires unconditionally, regardless of whether data has been flowing. That's a total-transfer deadline, not a read timeout.

If you want to keep the class but get idle-timeout semantics, the shape is a re-armed deadline: track `lastReadNanos`, and have the scheduled task compare against it and reschedule itself for the remainder instead of closing when the stream has made progress since it was armed. Costs one volatile write per read and behaves the way `SO_TIMEOUT` does everywhere else in Feign.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -159,6 +175,194 @@ protected Response toFeignResponse(Request request, HttpResponse<InputStream> ht
...
+        return afterRead(delegate.read());
+      } catch (IOException e) {
+        throw translateException(e);
+      }
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+      try {
+        return afterRead(delegate.read(b, off, len));
```
velo, line 227: Nit: if the timer fires between `delegate.read(b, off, len)` returning `n` and `checkTimedOut()`, those `n` bytes are already in the caller's buffer but the call throws and the count is lost. Practically harmless since it's a terminal timeout, but `read` isn't exception-safe as written — checking the flag *before* delegating would be.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -159,6 +175,194 @@ protected Response toFeignResponse(Request request, HttpResponse<InputStream> ht
...
+        final long skipped = delegate.skip(n);
+        checkTimedOut();
+        return skipped;
+      } catch (IOException e) {
+        throw translateException(e);
+      }
+    }
+
+    @Override
+    public int available() throws IOException {
```
velo, line 245: `available()`, `mark()`, `reset()` and `markSupported()` all bypass the `timedOut` check. After the timer has closed the delegate, `available()` will either return 0 or throw a raw `IOException` depending on the underlying stream, rather than the `HttpTimeoutException` the rest of the class is careful to produce. `reset()` in particular can hand back a stream that then times out mid-read.

java11/src/main/java/feign/http2client/Http2Client.java
```diff
@@ -159,6 +175,194 @@ protected Response toFeignResponse(Request request, HttpResponse<InputStream> ht
...
+        if (current instanceof HttpTimeoutException) {
+          return (HttpTimeoutException) current;
+        }
+        current = current.getCause();
+      }
+      return null;
+    }
+  }
+
+  private static final class TimeoutHttpResponse implements HttpResponse<InputStream> {
```
velo, line 315: 60 lines of pure delegation to swap out one method. This whole class disappears if the wrapping moves to the Feign `Response` — see my comment on the 3-arg `toFeignResponse` above.

java11/src/test/java/feign/http2client/Http2ClientContentLengthTest.java
```diff
@@ -185,4 +195,46 @@ void gzipDecodedBodyReportsUnknownLength() throws Exception {
...
+    final InputStream body =
+        new ByteArrayInputStream(new byte[] {1}) {
+          @Override
+          public void close() throws IOException {
+            closed.countDown();
+            super.close();
+          }
+        };
+    final Request.Options options =
+        new Request.Options(1, TimeUnit.SECONDS, 1, TimeUnit.MILLISECONDS, true);
```
velo, line 211: A 1ms timeout plus a 5s latch await is asking for trouble on CI — this asserts on scheduler latency, not on behaviour. `Http2ClientTest` is already one of our flakier suites and I'd rather not add to it.

If the timeout wrapper takes its scheduler (or a clock) as a seam rather than reaching for a static field, this becomes a deterministic test with no sleeping at all.

java11/src/test/java/feign/http2client/test/Http2ClientTest.java
```diff
@@ -379,4 +420,30 @@ public void parsesRequestAndResponse() throws IOException, InterruptedException
...
     return Feign.builder().client(new Http2Client());
   }
+
+  private static final class TwoArgumentOverrideClient extends Http2Client {
+
+    private boolean invoked;
+
+    @Override
+    protected Response toFeignResponse(
+        Request request, java.net.http.HttpResponse<InputStream> response) {
```
velo, line 430: Please use an import rather than the fully-qualified `java.net.http.HttpResponse` here and in `ThreeArgumentOverrideClient` — house style throughout the repo.

That said, both of these test doubles only exist to verify that a protected overload dispatches to its sibling. If the wrapping moves onto the Feign `Response` (see `Http2Client.java`), the overload and both of these classes are unnecessary.

velo (changes requested):
Good fix for a real gap — the JDK's HttpRequest.timeout() doesn't reliably cover ofInputStream() body reads, and the TimeoutHttpResponse wrapper is justified (not just indirection): the added tests correctly confirm the pre-existing protected 2-arg toFeignResponse(Request, HttpResponse<InputStream>) override point still gets invoked polymorphically once the 3-arg/timeout path is in place.

One blocker: BODY_READ_TIMEOUT_EXECUTOR is a `private static final ScheduledExecutorService` (Executors.newSingleThreadScheduledExecutor, daemon thread, no shutdown). That's the same shape as AsyncFeign's old LazyInitializedExecutorService, which this repo already removed via #3394 (issue #3178) because a static singleton executor pins its defining ClassLoader forever in redeploy scenarios. It's actually a bit worse here: AsyncFeign used newCachedThreadPool, whose idle threads self-terminate after the default 60s keep-alive with no extra work; newSingleThreadScheduledExecutor's core thread does not idle-timeout by default, so this would leak even if just moved to an instance field. Could you scope this per-Http2Client instance (matching AsyncFeign.defaultExecutorService()'s pattern), built as `new ScheduledThreadPoolExecutor(1, threadFactory)` with `allowCoreThreadTimeOut(true)` so it can actually self-terminate?

Two smaller, non-blocking suggestions:
- TimeoutInputStream hand-rolls pass-through overrides for available()/mark()/reset()/markSupported() that java.io.FilterInputStream already provides — extending FilterInputStream instead would drop ~20 lines of boilerplate and only need read()/read(byte[],int,int)/skip()/close() overridden.
- Worth a one-line comment on translateException/findTimeoutException explaining why two independent timeout mechanisms (the JDK-level HttpRequest timeout and this new stream-level one) coexist and need reconciling — that's non-obvious rationale a future reader would otherwise have to reconstruct from test names.

Happy to merge once the executor lifecycle is fixed.

PR #3463, opened 2026-07-01
Avoid converting empty GET bodies to POST

### LINKED ISSUE #2872: Unexpected behavior: GET converted to POST without body when @RequestParam is missing in OpenFeign
## Description
When using OpenFeign, I originally intended to send a query parameter with a GET request.

For example:
```java
@FeignClient(name = "test-client")
public interface TestClient {

    @GetMapping("/test")
    DataRs getData(String query); // `query` has a non-null value
}
```

My goal was to send the request like:
```java
GET /test?query=value
```

## What actually happened
- Because the `query` parameter was not annotated with `@RequestParam`, OpenFeign treated it as a request body.
- However, since GET requests should not have a body, the internal HTTP client (`HttpURLConnection`) automatically changed the method to POST at runtime:
```java
// https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L1326
if (method.equals("GET")) {
    method = "POST"; // Backward compatibility
}
```
Image

- Even though the HTTP method was changed to POST, the body was empty (Content-Length: 0).

- So:
  - The request became POST,
  - But no request body was actually sent,
  - And there was no error or warning during Feign request generation.

## Important note
- When I explicitly add @RequestParam, everything works correctly:
```java
@FeignClient(name = "test-client")
public interface TestClient {

    @GetMapping("/test")
    DataRs getData(@RequestParam String query); // ✅ Works as intended: GET with query string
}
```
- So I fully understand that @RequestParam is needed when sending query parameters.

## Suggestion
Given this behavior, I would like to ask:

Wouldn't it be more appropriate for OpenFeign to throw a warning or an error early — for example, during startup or client creation — if it detects a GET method that has parameters but is missing annotations like @RequestParam or @SpringQueryMap?

In many cases, when a developer defines a GET method with parameters, the expected behavior is to send those as query parameters, not as a body.
If the annotation is missing, and OpenFeign ends up silently treating the request as a POST with an empty body, it can easily cause confusion, unexpected behavior, or even runtime issues under strict API gateways or HTTP method validation rules.

**Additionally, when I checked the client-side logs, it still showed the request being sent as a GET,
but when I confirmed with the receiving server, the request was actually received as a POST.
This discrepancy made it very difficult to quickly identify and debug the issue.**
```java
@Slf4j
@Configuration
public class FeignLoggingInterceptorConfig {

    @Bean
    public RequestInterceptor requestLoggingInterceptor() {
        return template -> log.info("[Feign Request] method={}, url={}, body={}",
                                    template.method(), template.url(), template.body() != null ? new String(template.body()) : "null");
    }
}
```

I fully understand that @RequestParam is necessary when intending to send query parameters.
However, it seems like providing an early warning in cases where a GET method's parameters would be treated incorrectly could greatly help developers catch such issues before runtime.

If there is a particular reason why OpenFeign currently does not enforce this behavior — for example, to preserve flexibility or for backward compatibility — I would be happy to understand that as well.

hutiefang76:
Fixes #2872.

When the default client sends a GET request with a non-null but empty body, HttpURLConnection treats setDoOutput(true) as a signal that output will be written. That can silently turn the request into POST before it reaches the server.

This keeps the existing behavior for non-empty bodies and for methods that normally send bodies, but skips writing an empty body for GET requests so the method stays GET.

Validation:
- Added a regression test that sends a direct GET request with `new byte[0]` as the body and verifies MockWebServer receives GET.
- Confirmed the new test failed before the fix with `expected: "GET" but was: "POST"`.
- `JAVA_HOME=$(/usr/libexec/java_home -v 26) ./mvnw -pl core test`
- `git diff --check`

### DIFF
--- a/core/src/main/java/feign/DefaultClient.java
+++ b/core/src/main/java/feign/DefaultClient.java
@@ -189,7 +189,7 @@ public class DefaultClient implements Client {
 
     byte[] body = request.body();
 
-    if (body != null) {
+    if (body != null && (body.length > 0 || request.httpMethod() != Request.HttpMethod.GET)) {
       /*
        * Ignore disableRequestBuffering flag if the empty body was set, to ensure that internal
        * retry logic applies to such requests.
--- a/core/src/test/java/feign/client/DefaultClientTest.java
+++ b/core/src/test/java/feign/client/DefaultClientTest.java
@@ -24,6 +24,8 @@ import feign.Client.Proxied;
 import feign.DefaultClient;
 import feign.Feign;
 import feign.Feign.Builder;
+import feign.Request;
+import feign.Request.HttpMethod;
 import feign.RetryableException;
 import feign.assertj.MockWebServerAssertions;
 import java.io.IOException;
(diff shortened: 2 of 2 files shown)


velo (approved):

PR #3464, opened 2026-07-02
Support deflate request encoding in HC5

### LINKED ISSUE #2460: java.lang.IllegalArgumentException: Deflate Content-Encoding is not supported by feign-hc5
When we upgraded to spring-boot-dependencies-3.3.0, we also upgrade the spring-cloud according to https://spring.io/blog/2024/05/31/spring-cloud-2023-0-2-oss-and-2022-0-7-enterprise-have-been-released, Then my service  met error 
"Caused by: java.lang.IllegalArgumentException: Deflate Content-Encoding is not supported by feign-hc5"

Here is some finding of troubleshooting:

As we set spring.cloud.openfeign.compression.request.enabled=true, so the FeignContentGzipEncodingInterceptor(spring-cloud-openfeign-core-4.1.2) is enabled too
~~~java
public class FeignContentGzipEncodingInterceptor extends BaseRequestInterceptor {
    @Override
    public void apply(RequestTemplate template) {
        if (requiresCompression(template)) {
		addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, HttpEncoding.GZIP_ENCODING,
					HttpEncoding.DEFLATE_ENCODING);
	}
    }
}
~~~
If the compression condition matched, it will add fixed header(Content-Encoding: gzip,deflate), but after the PR [https://github.com/OpenFeign/feign/pull/2247](https://github.com/OpenFeign/feign/pull/2247), it will throw IllegalArgumentException once 'deflate' exists in header 'Content-Encoding'
~~~java
if (headerName.equalsIgnoreCase(Util.CONTENT_ENCODING)) {
        isGzip = headerEntry.getValue().stream().anyMatch(Util.ENCODING_GZIP::equalsIgnoreCase);
        boolean isDeflate =
            headerEntry.getValue().stream().anyMatch(Util.ENCODING_DEFLATE::equalsIgnoreCase);
        if (isDeflate) {
          // DeflateCompressingEntity not available in hc5 yet
          throw new IllegalArgumentException(
              "Deflate Content-Encoding is not supported by feign-hc5");
        }
      }
~~~

I think it is incompatible change, @fbus or term, Please take a look on it

kmdy7991:
Spring Cloud OpenFeign request compression can add `Content-Encoding: gzip,deflate`. `feign-hc5` currently rejects any request that includes `deflate`, so the request fails before it is sent.

This removes the unconditional rejection and adds request deflate support in HC5. The sync client uses Apache HttpClient 5's `DeflateCompressingEntity`; the async client compresses the simple request body bytes directly. When both `gzip` and `deflate` are present, `gzip` remains the effective encoding, matching Feign's existing default client precedence.

The same error was reproduced in Spring Cloud OpenFeign versions 4.1.2, 4.1.5, and 5.0.2 prior to the modification.
After applying the `feign-hc5` changes, we confirmed that the same application is working correctly.

Tests:

* `./mvnw -Pdev -pl hc5 -Dtest=DeflateHttp5ClientTest,GzipHttp5ClientTest,AsyncApacheHttp5ClientTest test`
* `./mvnw -Pdev -pl hc5 -am clean test`

Fixes #2460

### DIFF
--- a/hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
+++ b/hc5/src/main/java/feign/hc5/ApacheHttp5Client.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import org.apache.hc.client5.http.classic.HttpClient;
 import org.apache.hc.client5.http.config.Configurable;
 import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.entity.DeflateCompressingEntity;
 import org.apache.hc.client5.http.entity.GzipCompressingEntity;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
@@ -126,6 +127,7 @@ public final class ApacheHttp5Client implements Client {
     // request headers
     boolean hasAcceptHeader = false;
     boolean isGzip = false;
+    boolean isDeflate = false;
     for (final Map.Entry<String, Collection<String>> headerEntry : request.headers().entrySet()) {
       final String headerName = headerEntry.getKey();
       if (headerName.equalsIgnoreCase(ACCEPT_HEADER_NAME)) {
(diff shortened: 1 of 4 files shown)


velo (approved):

Issue #2752, opened 2025-01-31
Failing unit tests in Feign core


trumpetinc:
I'm running Feign Core unit tests on Windows platform, Java SE 21 (OpenJDK) and finding test failures.

These two appear to be actual functional problems:

1. **feign.DefaultContractInheritanceTest.overrideParameterizedApiSupported()**
2. **feign.DefaultContractInheritanceTest.multipleInheritanceDoneCorrectly2()**

Both failures look like this:

java.lang.IllegalStateException: Method OverrideParameterizedApi#get(String) not annotated with HTTP method type (ex. GET, POST)
Warnings:
- Class OverrideParameterizedApi has no annotations, it may affect contract Default
	at feign.Util.checkState(Util.java:121)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:117)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:71)
	at feign.DeclarativeContract.parseAndValidateMetadata(DeclarativeContract.java:40)
	at feign.DefaultContractInheritanceTest.overrideParameterizedApiSupported(DefaultContractInheritanceTest.java:88)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

Less important, but all 3 tests in this test unit are failing:

3. **feign.MultipleLoggerTest**

Failure is: org.junit.platform.commons.JUnitException: Failed to close extension context

I believe that these 3 failures are because JUL is grabbing file locks, which prevents JUnit from cleaning up it's temp folders under Windows.  One solution would be to add a close() method to JavaLogger...

velo:
Thanks for the Windows-specific test report. The DefaultContractInheritanceTest failures on Java 21/Windows and the JUL file-locking behavior around MultipleLoggerTest both look worth fixing, so I marked this as a bug.

The path I'd welcome here is a focused PR with tests or build adjustments that make these pass on Windows without weakening the existing contract behavior. I'll review that when it's up; if you'd rather sponsor/commission the fix, I'm also available through GitHub Sponsors: https://github.com/sponsors/velo.

trumpetinc:
@velo I created a PR for fix to the JUL file locking issue.  https://github.com/OpenFeign/feign/pull/3493

trumpetinc:
[...]

PR #3468, opened 2026-07-02
Close JavaLogger file handlers

hutiefang76:
## Summary

This addresses the `MultipleLoggerTest` / JUL `FileHandler` part of #2752.

`Logger.JavaLogger.appendToFile(...)` adds `FileHandler` instances to the underlying JUL logger, but there was no supported way to close them. On Windows this can keep the temp log files locked until the JVM exits, which prevents JUnit from cleaning up `@TempDir`.

The change makes `JavaLogger` `Closeable`, tracks only the handlers added by that `JavaLogger` instance, and closes/removes those handlers from `close()`. The `MultipleLoggerTest` cases now close their loggers after checking handler setup.

## Verification

- RED: `./mvnw -pl core -Dtest=MultipleLoggerTest#javaLoggerClosesFileHandlers test -DskipITs -Dlicense.skip=true -Dcheckstyle.skip=true -Drat.skip=true` failed before the production change because `JavaLogger.close()` did not exist.
- GREEN: `./mvnw -pl core -Dtest=MultipleLoggerTest test -DskipITs -Dlicense.skip=true -Dcheckstyle.skip=true -Drat.skip=true` passed, 4 tests.
- GREEN: `./mvnw -pl core -Dtest=DefaultContractInheritanceTest,MultipleLoggerTest test -DskipITs -Dlicense.skip=true -Dcheckstyle.skip=true -Drat.skip=true` passed, 13 tests.
- GREEN: `./mvnw -pl core test -DskipITs -Dlicense.skip=true` passed, 635 tests, 0 failures/errors, 3 skipped.
- GREEN: `./mvnw -pl core git-code-format:validate-code-format -Dlicense.skip=true -DskipTests` passed.
- GREEN: `git diff --check` passed.

I did not change the `DefaultContractInheritanceTest` path in this PR; that looks like a separate reflection-order/generic bridge-method issue and is better handled independently.

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -21,10 +21,14 @@ import static feign.Util.ensureClosed;
 import static feign.Util.valuesOrEmpty;
 import static java.util.Objects.nonNull;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.logging.FileHandler;
+import java.util.logging.Handler;
 import java.util.logging.LogRecord;
 import java.util.logging.SimpleFormatter;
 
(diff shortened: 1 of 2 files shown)


trumpetinc:
FYI - I have an easier fix for this issue: https://github.com/OpenFeign/feign/pull/3493

The reality is that closing log files while the application is running is not a normal thing, so I think the easy approach to just ignore the file deletion during test is appropriate.

velo (approved):

PR #3493, opened 2026-07-27
Windows prevents folder deletion if log file is still locked. There is

trumpetinc:
Simple fix for https://github.com/OpenFeign/feign/issues/2752


trumpetinc:
@velo FYI - here's a fix for #2572

velo (approved):

PR #3503, opened 2026-07-31
Stop Http2ClientTest from calling live nghttp2.org

### LINKED ISSUE #3483: 🐛 bug(feign-14): `Http2ClientTest` depends on external service `nghttp2.org/httpbin` and fails the build when it is down
## Description

The test class `java11/src/test/java/feign/http2client/test/Http2ClientTest.java` contains four tests that make real
HTTP calls to `https://nghttp2.org/httpbin/`:

- `patch()`
- `noResponseBodyForPatch()`
- `getWithRequestBody()`
- `deleteWithRequestBody()`

These tests are guarded by `assumeTrue(HTTPBIN_REACHABLE, ...)`, but the reachability check only performs a DNS lookup (
`InetAddress.getByName("nghttp2.org")`). When the DNS resolves but the HTTP service itself is down or unresponsive (
e.g., the `https://nghttp2.org/httpbin/patch` endpoint returns 504 Gateway Timeout, or the server is timing out), the
`assumeTrue` passes but the subsequent HTTP requests either fail or **hang indefinitely** — causing the **entire project
build to fail or freeze**.

Image

## Steps to Reproduce

1. Run `./mvnw -Pdev clean license:format install` (or `./mvnw -Pdev -pl java11 clean license:format install`)
2. Observe that `https://nghttp2.org/httpbin/patch` is unresponsive or returning errors
3. Build fails with connection/timeout/HTTP error exceptions from the four tests above, **or hangs indefinitely** (e.g.,
   when the endpoint returns 504 Gateway Timeout, the test freezes rather than failing fast)

## Expected Behavior

Tests that depend on external services should either:

- Be skipped reliably when the service is unavailable (e.g., check HTTP connectivity, not just DNS), or
- Use `MockWebServer` (already available as a test dependency) instead of a live external service

## Actual Behavior

The DNS-based reachability check passes, but the HTTP requests fail (or hang indefinitely, e.g., on a 504 Gateway
Timeout), causing the entire build to fail or freeze.

## Notes

I may look into this later, but not in the near future.

## Affected Code

https://github.com/OpenFeign/feign/blob/14.x/java11/src/test/java/feign/http2client/test/Http2ClientTest.java

Lines 43–54 (reachability check), and lines 98–112, 171–188 (the four affected tests).

https://github.com/OpenFeign/feign/blob/b216d52064348d77e489b0ccc4407456fa5e58f2/java11/src/test/java/feign/http2client/test/Http2ClientTest.java#L43-L54

https://github.com/OpenFeign/feign/blob/b216d52064348d77e489b0ccc4407456fa5e58f2/java11/src/test/java/feign/http2client/test/Http2ClientTest.java#L96-L112

https://github.com/OpenFeign/feign/blob/b216d52064348d77e489b0ccc4407456fa5e58f2/java11/src/test/java/feign/http2client/test/Http2ClientTest.java#L170-L188

arimu1:
## What
`Http2ClientTest` on `14.x` had four tests that hit live `https://nghttp2.org/httpbin/` after a DNS-only reachability check. When DNS resolves but the HTTP service is down or slow, those tests fail or hang the whole build.

## How
- `patch` / `noResponseBodyForPatch` / `deleteWithRequestBody` now use the existing MockWebServer fixture.
- `getWithRequestBody` uses a small local `ServerSocket` server (MockWebServer rejects GET requests that carry a body).
- Removed the `HTTPBIN_REACHABLE` DNS gate and the `JSONAssert` dependency from these tests.
- Approach matches what already landed on `master`.

## Testing
```
./mvnw -Dtoolchain.skip=true -pl java11 -am test -Dtest=Http2ClientTest -Dsurefire.failIfNoSpecifiedTests=false
```
Result: **Tests run: 33, Failures: 0, Errors: 0, Skipped: 0**

Fixes #3483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Version 14.0
 
+* `Http2ClientTest` no longer depends on live `nghttp2.org/httpbin` calls; patch/delete use MockWebServer and GET-with-body uses a local socket server so CI does not hang or fail when the external host is down. (https://github.com/OpenFeign/feign/issues/3483)
+
 * `DefaultEncoder` now supports streaming request bodies for `File`, `Path`, `InputStream`, and `Request.Body` types,
   avoiding in-memory buffering. New `Request.PathBody` and `Request.InputStreamBody` implementations are provided for
   these cases. (https://github.com/OpenFeign/feign/pull/3396)
--- a/java11/src/test/java/feign/http2client/test/Http2ClientTest.java
+++ b/java11/src/test/java/feign/http2client/test/Http2ClientTest.java
@@ -19,7 +19,6 @@ import static feign.Util.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.entry;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import feign.*;
 import feign.assertj.MockWebServerAssertions;
@@ -27,31 +26,23 @@ import feign.client.AbstractClientTest;
 import feign.http2client.Http2Client;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.net.InetAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
 import java.net.http.HttpTimeoutException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import mockwebserver3.MockResponse;
 import mockwebserver3.RecordedRequest;
 import org.junit.jupiter.api.Test;
-import org.skyscreamer.jsonassert.JSONAssert;
 
 /** Tests client-specific behavior, such as ensuring Content-Length is sent when specified. */
 public class Http2ClientTest extends AbstractClientTest {
-
-  private static final boolean HTTPBIN_REACHABLE;
-
-  static {
-    boolean reachable = false;
-    try {
-      java.net.InetAddress.getByName("nghttp2.org");
-      reachable = true;
-    } catch (java.net.UnknownHostException _) {
-      // ignore
-    }
-    HTTPBIN_REACHABLE = reachable;
-  }
 
   public interface TestInterface {
     @RequestLine("PATCH /patch")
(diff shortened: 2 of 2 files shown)


kdelay (commented):
Ran this on the PR head (3408015c) with JDK 26.0.1 on macOS arm64:

```
./mvnw -Dtoolchain.skip=true -pl java11 -am test -Dtest=Http2ClientTest -Dsurefire.failIfNoSpecifiedTests=false
Tests run: 33, Failures: 0, Errors: 0, Skipped: 0
```

The result you reported reproduces, and the three MockWebServer conversions (`patch`, `noResponseBodyForPatch`, `deleteWithRequestBody`) read correctly to me. One point on `getWithRequestBody` I think is worth settling before this lands.

**The local socket server turns a future regression into a ~108s failure with the wrong message.**

The stated goal is that CI no longer hangs or fails on an external host. The three MockWebServer tests get that. `getWithRequestBody` does not, because writing the response is on the success path only.

The read loop exits on `contains("some request body")` or EOF. If the request body ever stops arriving, which is exactly the regression this test exists to catch, the loop blocks until `setSoTimeout(5000)` fires, the thread dies with `UncheckedIOException`, and no response is ever written.

I measured it by changing only the sentinel string to one that never appears and leaving the rest of your branch untouched:

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Time elapsed: 107.7 s
[ERROR] Http2ClientTest.getWithRequestBody:215 » Retryable HTTP connect timed out executing GET http://localhost:52262/anything
```

The real cause appears only in `system-err` of the surefire report, detached from the failure:

```
Exception in thread "Thread-1" java.io.UncheckedIOException: java.net.SocketTimeoutException: Read timed out
```

So a missing request body is reported as a connect timeout, 108 seconds later. On the timing: the first connection is accepted and then dropped when the server thread dies, and subsequent retry attempts land in a backlog of 1 that nobody accepts again. I did not instrument the individual retries, so treat that mechanism as my reading of the numbers rather than something I measured directly.

Two adjustments, both already idiomatic in this file:

1. `timeoutTest` just above uses `.retryer(Retryer.NEVER_RETRY)` with an explicit `Request.Options`. The same here bounds a failure to one short attempt instead of the retried default.
2. Run the server body on an `ExecutorService` or `CompletableFuture` and resolve it with a timeout before asserting, so a server-side `IOException` surfaces as the test failure rather than a stray stderr trace. That also covers `serverThread.join(5000)`, whose return value is currently ignored: if the join times out the test still proceeds and asserts on a possibly-null `receivedRequest`, which surfaces as "Expecting actual not to be null" instead of the underlying I/O error.

For the record, I only exercised this branch; I did not compare against the master-side approach you mention.

velo (approved):

velo:
would be a nice change targeting main

PR #3516, opened 2026-08-12
Skip synthetic/bridge methods in BaseContract parsing

AzazelSensei:
## Summary

`Contract.BaseContract.parseAndValidateMetadata` walks `Class.getMethods()` but did not skip compiler-generated methods. Covariant overrides produce synthetic bridge methods without HTTP annotations, which then fail with `Method … not annotated with HTTP method type`.

This skips `isSynthetic()` / `isBridge()` methods in that loop so inheritance cases like `overrideParameterizedApiSupported` and `multipleInheritanceDoneCorrectly2` parse cleanly.

## Related

Fixes the remaining `DefaultContractInheritanceTest` failures from #2752 (JUL/`MultipleLoggerTest` side is handled separately).

## Test plan

- [x] `mvn test -pl core -Dtest=DefaultContractInheritanceTest` (9 tests, 0 failures)
- [ ] Confirm CI green on this PR

### DIFF
--- a/core/src/main/java/feign/Contract.java
+++ b/core/src/main/java/feign/Contract.java
@@ -60,7 +60,9 @@ public interface Contract {
         if (method.getDeclaringClass() == Object.class
             || (method.getModifiers() & Modifier.STATIC) != 0
             || Util.isDefault(method)
-            || method.isAnnotationPresent(FeignIgnore.class)) {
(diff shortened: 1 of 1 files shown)


AzazelSensei:
Superseded by https://github.com/OpenFeign/feign/pull/3517.

PR #3534, opened 2026-08-21
Apply review feedback to the multi-encoder and multi-decoder surface

velo:
@trumpetinc's review of #3527, applied to both sides of the codec surface. Everything here is still
`@Experimental`, so this is the moment to move it.

### Taken

**A multi-codec is itself predicated.** `MultiEncoder` implements `PredicatedEncoder` and
`MultiDecoder` implements `PredicatedDecoder`, accepting whatever any of their delegates accepts,
so one nests inside another. That is the composition you asked for — a library can ship a set as a
single unit:

```java
Feign.builder().encoders(StreamingFeign.encoders(), new JacksonEncoder());
```

The winning delegate's `canEncode` is evaluated twice, once by the outer `anyMatch` and once by the
inner loop. Predicates are already documented as cheap and free of side effects, so that is
acceptable, but it is worth saying out loud.

Nested sets are unfolded in the failure message, so contributing a bundle does not turn it into one
opaque line:

```
Encoders tried, in order:
  - JacksonEncoder
  - MultiEncoder:
    - StreamingStringEncoder when the return type is a Stream
    - StreamingBytesEncoder when Content-Type is application/octet-stream
```

**`DefaultEncoder` and `DefaultDecoder` declare themselves.** You were right and I was wrong: the
predicate is not a yes-man, it is exactly what `encode`/`decode` handles. This makes raw
passthrough composable, which is genuinely useful — a `String` body reaching Jackson comes out
quoted, so putting the default first fixes that:

```java
Feign.builder().encoders(new DefaultEncoder(), new JacksonEncoder());
```

One correction to the predicate you proposed: `DefaultEncoder.encode` also accepts a null body,
which it sends as no body at all. Dropping that case would break

```java
@RequestLine("POST /orders") void create(Order order);
client.create(null);
```

so the predicate is `bodyType == String.class || bodyType == byte[].class || object == null`.
`DefaultDecoder` gets the same treatment, including 404/204 and a null body, which it handles for
any type.

**An empty builder fails at wiring time**, `IllegalStateException` from `build()`, rather than
producing a codec that can only ever throw on the first request. `encoders()`/`decoders()` with no
arguments fail the same way.

**`Builder.narrow(predicate, codec)`**, so the replace/AND pair sits side by side where the
difference is visible, and the javadoc on `of` and `narrowing` now says which is which in one line.
More on that below.

**The `formRequests()` javadoc wording**, and a note on the `FormEncoder(Encoder)` constructor that
chaining belongs in `encoders(...)` and the delegate is expected to be deprecated once this surface
stops being experimental.

### Pushed back on

**`narrowing` is not redundant with `add(predicate, encoder)`.** `PredicatedEncoder.of` *replaces*
whatever the encoder declares — deliberately, so you can point a `GsonEncoder` at a vendor content
type it would otherwise refuse. `narrowing` is the AND. Both are needed. That said, the fact that it
read as redundant is the finding: the names were not carrying the difference. Hence the builder
`narrow` and the sharper javadoc rather than a rename, since the statics are already published.

**The failure message reports `Content-Type` and `Accept`, not every header.** `EncodeException`
messages land in logs and the header map routinely holds `Authorization`. The detail is not lost —
every predicate carries a description and is printed alongside the codec it guards, so a predicate
keyed on an unusual header says so itself:

```java
.add(EncoderPredicate.describedAs("X-Acme-Format is protobuf", ...), protoEncoder)
```
```
Unable to encode com.acme.Order (Content-Type: text/plain, Accept: application/json) for POST /orders.
Encoders tried, in order:
  - JacksonEncoder
  - ProtobufEncoder when X-Acme-Format is protobuf
Register an encoder that accepts it, or add a catch-all (EncoderPredicate.any()) last.
```

The hint wording is softened too, since `.any()` is an implementation detail for anyone who came in
through `encoders(...)`.

On the decode side the second header is the *request's* `Accept`, which is what the caller asked
for.

### Not done

**`DefaultEncoder` as a `MultiEncoder`** — you flagged it as an observation rather than a proposal,
and I agree it is one. It trades a ten-line class for a list allocation and makes the failure
message worse.

**Centralizing content-type parsing** on the parser in #3494 — agreed in principle, but that PR is
still open. The `Util` helpers are `@Experimental`, so whichever lands second swaps the internals;
no need to couple them now.

`mvn clean install` passes.

### DIFF
--- a/core/src/main/java/feign/codec/DefaultDecoder.java
+++ b/core/src/main/java/feign/codec/DefaultDecoder.java
@@ -20,11 +20,11 @@ import feign.Util;
 import java.io.IOException;
 import java.lang.reflect.Type;
 
-public class DefaultDecoder extends StringDecoder implements PredicatedDecoder {
+public class DefaultDecoder extends StringDecoder {
 
   /**
-   * Accepts exactly what {@link #decode} handles: a {@code String} or {@code byte[]} return type,
-   * and any type at all when there is no body to read.
+   * Accepts exactly what {@link #decode} handles: everything {@link StringDecoder} accepts, plus a
+   * {@code byte[]} return type.
    *
    * @param response {@inheritDoc}
    * @param type {@inheritDoc}
@@ -32,11 +32,7 @@ public class DefaultDecoder extends StringDecoder implements PredicatedDecoder {
    */
   @Override
   public boolean canDecode(Response response, Type type) {
-    return response.status() == 404
-        || response.status() == 204
-        || response.body() == null
-        || byte[].class.equals(type)
-        || String.class.equals(type);
+    return byte[].class.equals(type) || super.canDecode(response, type);
   }
 
   @Override
--- a/core/src/main/java/feign/codec/DefaultDecoder.java
+++ b/core/src/main/java/feign/codec/DefaultDecoder.java
@@ -20,7 +20,24 @@ import feign.Util;
 import java.io.IOException;
 import java.lang.reflect.Type;
 
-public class DefaultDecoder extends StringDecoder {
+public class DefaultDecoder extends StringDecoder implements PredicatedDecoder {
+
+  /**
+   * Accepts exactly what {@link #decode} handles: a {@code String} or {@code byte[]} return type,
+   * and any type at all when there is no body to read.
+   *
+   * @param response {@inheritDoc}
+   * @param type {@inheritDoc}
+   * @return {@inheritDoc}
+   */
+  @Override
+  public boolean canDecode(Response response, Type type) {
+    return response.status() == 404
+        || response.status() == 204
+        || response.body() == null
+        || byte[].class.equals(type)
+        || String.class.equals(type);
+  }
 
   @Override
   public Object decode(Response response, Type type) throws IOException {
--- a/core/src/main/java/feign/codec/MultiDecoder.java
+++ b/core/src/main/java/feign/codec/MultiDecoder.java
@@ -17,6 +17,7 @@ package feign.codec;
 
 import feign.Experimental;
 import feign.FeignException;
+import feign.Request;
 import feign.Response;
 import feign.Util;
 import java.io.IOException;
@@ -52,11 +53,19 @@ import java.util.stream.Collectors;
  * naming what was tried. Add a decoder guarded by {@link DecoderPredicate#any()} last to act as a
  * default, as above.
  *
+ * <p>A multi-decoder is itself a {@link PredicatedDecoder}, accepting whatever any of its decoders
+ * accepts, so one can be added to another. That is how a library ships a set of decoders as a
+ * single unit:
+ *
+ * <pre>
+ * Feign.builder().decoders(StreamingFeign.decoders(), new JacksonDecoder());
+ * </pre>
+ *
  * @see PredicatedDecoder
  * @see DecoderPredicate
  */
 @Experimental
-public class MultiDecoder implements Decoder {
+public class MultiDecoder implements PredicatedDecoder {
 
   private final List<PredicatedDecoder> decoders;
 
@@ -67,6 +76,18 @@ public class MultiDecoder implements Decoder {
   /** Starts building a multi-decoder. */
   public static Builder builder() {
     return new Builder();
+  }
+
+  /**
+   * Whether any of the decoders accepts the response.
+   *
+   * @param response {@inheritDoc}
+   * @param type {@inheritDoc}
+   * @return {@inheritDoc}
+   */
+  @Override
+  public boolean canDecode(Response response, Type type) {
+    return decoders.stream().anyMatch(decoder -> decoder.canDecode(response, type));
   }
 
   /**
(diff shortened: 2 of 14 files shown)


core/src/main/java/feign/codec/DefaultDecoder.java
```diff
@@ -20,7 +20,24 @@
 import java.io.IOException;
 import java.lang.reflect.Type;
 
-public class DefaultDecoder extends StringDecoder {
+public class DefaultDecoder extends StringDecoder implements PredicatedDecoder {
```
trumpetinc: For completeness, should we make StringDecoder a PredicatedDecoder?

(PS - it is unfortunate that DefaultDecoder extends StringDecoder... inheritance vs composition mistake)

core/src/main/java/feign/codec/DefaultDecoder.java
```diff
@@ -20,7 +20,24 @@
...
+   * @param type {@inheritDoc}
+   * @return {@inheritDoc}
+   */
+  @Override
+  public boolean canDecode(Response response, Type type) {
+    return response.status() == 404
+        || response.status() == 204
+        || response.body() == null
+        || byte[].class.equals(type)
+        || String.class.equals(type);
```
trumpetinc: If we make StringDecoder a PredictedDecoder, this should check super.canDecode() instead of hard coding the String.class.equals(type) check.

core/src/main/java/feign/codec/MultiDecoder.java
```diff
@@ -52,11 +53,19 @@
  * naming what was tried. Add a decoder guarded by {@link DecoderPredicate#any()} last to act as a
  * default, as above.
  *
+ * <p>A multi-decoder is itself a {@link PredicatedDecoder}, accepting whatever any of its decoders
+ * accepts, so one can be added to another. That is how a library ships a set of decoders as a
+ * single unit:
+ *
+ * <pre>
+ * Feign.builder().decoders(StreamingFeign.decoders(), new JacksonDecoder());
```
trumpetinc: I want to call out that StreamingFeign.decoders() doesn't currently exist (streaming requires breaking changes, so is not part of head at this time).

It may be advisable to say "Given a hypothetical library called StreamingFeign that has a decoders() method that returns a MultiDecoder of it's various decoders, this is how the library's decoders would be added:"

Or maybe cleaner to just not talk about this at all until we have a proper example?

core/src/main/java/feign/codec/MultiDecoder.java
```diff
@@ -95,31 +116,61 @@ private String unableToDecode(Response response, Type type) {
...
-            .filter(header -> Util.CONTENT_TYPE.equalsIgnoreCase(header.getKey()))
+  /**
+   * Lists the decoders one per line, unfolding nested multi-decoders so that a set contributed as a
+   * single unit still shows what it contains.
+   */
+  private void appendTo(StringBuilder message, String indent) {
+    for (PredicatedDecoder decoder : decoders) {
+      if (decoder instanceof MultiDecoder) {
+        message.append(indent).append("- MultiDecoder:");
+        ((MultiDecoder) decoder).appendTo(message, indent + "  ");
```
trumpetinc, line 140: Good call on the indentation - that will make it very clean to diagnose issues.

core/src/main/java/feign/codec/MultiDecoder.java
```diff
@@ -157,8 +211,33 @@ public Builder add(DecoderPredicate predicate, Decoder decoder) {
       return add(PredicatedDecoder.of(predicate, decoder));
     }
 
-    /** Builds the multi-decoder. */
+    /**
+     * Adds a decoder that already declares itself, narrowed by the given predicate: both it and the
```
trumpetinc: "that already declares itself" is a bit confusing.  I think this could be replaced with:

"Adds a Decoder, narrowed by the given predicate.  If the Decoder is itself a PredicatedDecoder, the provided predicate is applied in addition to the PredicatedDecoder's existing predicate."

core/src/main/java/feign/codec/MultiDecoder.java
```diff
@@ -157,8 +211,33 @@ public Builder add(DecoderPredicate predicate, Decoder decoder) {
...
+     * decoder's own {@code canDecode} have to accept the response.
+     *
+     * <pre>
+     * MultiDecoder.builder()
+     *     .narrow(DecoderPredicate.status(200), new JacksonDecoder())
+     *     .add(new JacksonDecoder())
```
trumpetinc: Example is a little confusing.  Should we specify a different decoder type for status=200 ?  Maybe change the .add() line to new DefaultDecoder() ?

core/src/main/java/feign/codec/MultiEncoder.java
```diff
@@ -50,11 +50,19 @@
  * naming what was tried. Add an encoder guarded by {@link EncoderPredicate#any()} last to act as a
  * default, as above.
  *
+ * <p>A multi-encoder is itself a {@link PredicatedEncoder}, accepting whatever any of its encoders
+ * accepts, so one can be added to another. That is how a library ships a set of encoders as a
+ * single unit:
+ *
+ * <pre>
+ * Feign.builder().encoders(StreamingFeign.encoders(), new JacksonEncoder());
```
trumpetinc: same comment as above about StreamingFeign.encoders() not existing.

core/src/main/java/feign/codec/MultiEncoder.java
```diff
@@ -155,8 +207,33 @@ public Builder add(EncoderPredicate predicate, Encoder encoder) {
...
 
-    /** Builds the multi-encoder. */
+    /**
+     * Adds an encoder that already declares itself, narrowed by the given predicate: both it and
+     * the encoder's own {@code canEncode} have to accept the request.
+     *
+     * <pre>
+     * MultiEncoder.builder()
+     *     .narrow(EncoderPredicate.contentType("application/vnd.acme+json"), new GsonEncoder())
+     *     .add(new GsonEncoder())
```
trumpetinc: similar to comment above in MultiDecoder.narrow() - shold the .add() line call out a different encoder type so the example makes sense?

trumpetinc (commented):
@velo review is done.  The functional code looks solid to me, most comments are small things about the documentation, one question about StringDecoder.

Super excited to see this getting added to Feign.

core/src/main/java/feign/codec/MultiEncoder.java
```diff
@@ -208,13 +209,14 @@ public Builder add(EncoderPredicate predicate, Encoder encoder) {
...
-     * the encoder's own {@code canEncode} have to accept the request.
+     * Adds an encoder, narrowed by the given predicate. If the encoder is itself a {@link
+     * PredicatedEncoder}, the predicate applies in addition to the encoder's own {@code canEncode}
+     * rather than instead of it: both have to accept the request.
      *
      * <pre>
      * MultiEncoder.builder()
      *     .narrow(EncoderPredicate.contentType("application/vnd.acme+json"), new GsonEncoder())
-     *     .add(new GsonEncoder())
+     *     .add(EncoderPredicate.any(), new DefaultEncoder())
```
trumpetinc, line 219: For what it's worth, I don't like using EncoderPredicate.any() with DefaultEncoder.  The reality is that .any() isn't really a good predicate for any Encoder or Decoder...  I suggest:

```
.add(new DefaultEncoder())
```

PR #3535, opened 2026-08-23
Encode Gson request bodies using the runtime class

### LINKED ISSUE #2485: incomplete Object information when serialize object with Gson
```java

    @PostMapping("/message/send?access_token={access_token}")
    MessageResponseDTO send(@RequestParam("access_token") String accessToken, @RequestBody AbsMessage message);
````
Such as the code, I need to call it,the interface has different performance depends on the Object passed.
However AbsMessage is a abstract class,when I call it need to pass an object which instance of AbsMessage.
But the reuest body doesn't conatin any field from the runtime sub class whic I passed. All fields are from the abstract class.
I find the cause at last.

```java
  GsonEncoder#encode(Object object, Type bodyType, RequestTemplate template)
  @Override
    template.body(gson.toJson(object, bodyType));
  }
```
It serializes object with the type which declared in the function,not  runtime  type of the passed object.
If gson.toJson(object),It appears normal.

tanvir-ux:
Fixes #2485.

**Problem**
When a Feign method declares an abstract (or superclass) body type, `GsonEncoder` serializes only the fields on that declared type. Fields on the concrete subclass never appear in the JSON.

**Root cause**
`GsonEncoder.encode` called `gson.toJson(object, bodyType)`. Gson treats the second argument as the schema, so a `SimpleTextMessage` passed as `Message` is written as `{"toUser":"..."}`.

**Fix**
If `bodyType` is a raw `Class`, serialize with `object.getClass()`. Parameterized types (`TypeToken` / generics) still use the declared type so type arguments are not erased. Jackson and the other encoders are unchanged.

**Tests**
`GsonCodecTest.encodesSubclassFieldsWhenBodyTypeIsAbstract` encodes a subclass as the abstract type and asserts the JSON includes the subclass fields. Existing TypeToken encoding tests still pass.

`./mvnw test -pl gson -am` (gson: 11 tests, 0 failures).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### Version 13.15
+
+* `GsonEncoder` serializes request bodies using the runtime type when the declared body type is a
+  raw `Class`, so subclass fields on an abstract `@RequestBody` type are included. Parameterized
+  types (TypeToken / generics) still use the declared type (#2485).
+
 ### Version 13.14
 
 * Add `@Experimental` `MultiEncoder`, `PredicatedEncoder` and `EncoderPredicate`, letting a single
--- a/gson/src/main/java/feign/gson/GsonEncoder.java
+++ b/gson/src/main/java/feign/gson/GsonEncoder.java
@@ -43,7 +43,19 @@ public class GsonEncoder implements Encoder, PredicatedEncoder, JsonEncoder {
 
   @Override
   public void encode(Object object, Type bodyType, RequestTemplate template) {
-    template.body(gson.toJson(object, bodyType));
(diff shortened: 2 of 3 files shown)


velo (approved):
Clean, correctly-scoped fix. Verified locally: mvn -pl gson -am test passes (all 12 gson tests green), the CHANGELOG entry matches the current SNAPSHOT version, and the fix lives in the one file that owns the problem rather than leaking into the shared codec layer. typeToEncode() correctly preserves ParameterizedType/TypeToken handling so the existing generics path isn't regressed. Two purely cosmetic nits (a stray comment-ratio note and one inconsistent text-block indent in the test) aren't worth holding this up for.

PR #3544, opened 2026-08-27
Document that Logger reflects Feign's Request, not the wire request

### LINKED ISSUE #2682: Logger logs unsent request headers even though Default in Client has restricted headers
Since the default class `Default` in `Client` uses `sun.net.www.protocol.http.HttpURLConnection`, which has  the set of restricted headers, even if I add a header using `RequestTemplate#header(String, String)`, the actual HTTP request does not contain the header.

However `feign.Logger` logs not the actual request but the `Request` made by `SynchronousMethodHandler#executeAndDecode()`, which contains the restricted header I have added.

This inconsistency made me hard to discover the problem because there's no such documentation tells the logged request may not be the same as the actual HTTP request.

I hope to add some documentation in javadoc or, if possible, make `Logger` logs the actual HTTP request.

ChrisJr404:
Adds a note to the Logger class javadoc explaining that request logging reflects the Request that Feign builds, which can differ from what the underlying Client actually sends. The default HttpURLConnection based Client drops headers the JDK treats as restricted, so a header can show up in the log without being on the wire. Closes #2682.

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -32,7 +32,18 @@ import java.util.logging.Handler;
 import java.util.logging.LogRecord;
 import java.util.logging.SimpleFormatter;
 
-/** Simple logging abstraction for debug messages. Adapted from {@code retrofit.RestAdapter.Log}. */
+/**
+ * Simple logging abstraction for debug messages. Adapted from {@code retrofit.RestAdapter.Log}.
+ *
+ * <p>Request logging reflects the {@link Request} that Feign builds from the method invocation,
+ * captured before it is handed to the {@link Client}. Depending on the {@link Client}
+ * implementation, the bytes actually sent over the wire may differ from what is logged here. In
+ * particular the default {@link Client.Default} is backed by {@link java.net.HttpURLConnection},
+ * which silently drops headers the JDK treats as restricted (for example {@code Host}, {@code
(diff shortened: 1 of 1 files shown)


velo (changes requested):
Thanks for adding this — the clarification about Logger reflecting Feign's Request rather than the wire request is a genuinely useful thing to document, and I confirmed the core claim is accurate (Logger.logRequest is called in SynchronousMethodHandler before client.execute, so the request logged really can differ from what goes over the wire).

Two things before merging:

1. The doc points at `{@link Client.Default}`, but that class is `@Deprecated` (see its javadoc: "use `DefaultClient` instead"). `Feign.Builder` actually defaults to `new DefaultClient(...)`, and the restricted-header-dropping behavior you're describing lives in `DefaultClient` (e.g. the `Content-Length: 0` handling around line 228), not in the deprecated wrapper. Please repoint the `{@link}` to `DefaultClient`.

2. Consider moving the detailed JDK-specific rationale (the `HttpURLConnection` restricted-headers behavior, `sun.net.http.allowRestrictedHeaders`, specific header names) onto `DefaultClient`'s own class javadoc, with a short pointer from `Logger`'s doc ("see `{@link DefaultClient}` for a concrete example"). That keeps the detail next to the code that actually causes it, where someone touching `DefaultClient`'s header handling would be looking.

Happy to merge once the class reference is fixed.

PR #3546, opened 2026-08-28
Add opt-in target host and port observation tags

### LINKED ISSUE #2630: feign.micrometer.DefaultFeignObservationConvention not adding hostname and port to Observations
The method feign.micrometer.DefaultFeignObservationConvention#getLowCardinalityKeyValues adds method, status, clientName and uri to observations as attributes.  The uri is only the path and not the full url.  As such, none of our observations contain the host and port (aka the base url) of any downstream calls.  When loading these observations into our APM system, the only way to identify these calls is the clientName attribute which is the class name of the feign client object.  There is no good way to correlate spans that call the same service if those class names aren't identical across applications (and they typically are not the same)

It would be beneficial to add the downstream hostname and port that is included in the FeignTarget object in the observations so that we can see what host is being called and correlate that across spans for different applications.

adding something something like this would help a lot:
`
URL url = URI.create(requestTemplate.feignTarget().url()).toURL();
lowCardinalityKeyValues.and("http.host", url.getHost()).and("http.port", String.valueOf(url.getPort));
`

henriquejsza:
## Summary

- add an opt-in `TargetHostAndPortFeignObservationConvention`
- emit `net.peer.host` and an explicitly configured `net.peer.port`
- preserve the existing default low-cardinality key set
- safely omit target tags for URL-less or invalid targets
- document activation, port semantics, and cardinality considerations

`supportsContext` is declared explicitly with the existing semantics so JApiCmp can validate the new public convention.

## Testing

- `./mvnw -Pdev -Dtoolchain.skip=true -pl micrometer -am verify`

Fixes #2630

### DIFF
--- a/micrometer/README.md
+++ b/micrometer/README.md
@@ -54,6 +54,49 @@ pass it to the constructor:
 new MicrometerObservationCapability(observationRegistry, new MyFeignObservationConvention());
 ```
 
+#### Emitting the target host and port
+
+By default the observation does **not** carry the target host or port. If you
+want them, opt in with the bundled
+`TargetHostAndPortFeignObservationConvention`:
+
+```java
+GitHub github = Feign.builder()
+                     .addCapability(
+                         new MicrometerObservationCapability(
+                             observationRegistry,
+                             TargetHostAndPortFeignObservationConvention.INSTANCE))
+                     .target(GitHub.class, "https://api.github.com");
+```
+
+This adds the following low-cardinality key values on top of the defaults:
+
+| Key | Value |
+| --- | ----- |
+| `net.peer.host` | Host parsed from the configured `Target` url (e.g. `api.github.com`). |
+| `net.peer.port` | Port parsed from the configured `Target` url — **only when the url declares one explicitly**. |
+
+Port semantics:
+
+* The host and port are read from the configured `Target` url
+  (`requestTemplate().feignTarget().url()`), not the final resolved request url.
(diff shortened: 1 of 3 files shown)


velo (changes requested):
Nice addition — subclassing DefaultFeignObservationConvention and wiring in through MicrometerObservationCapability's existing customConvention extension point is exactly the right seam for this. A couple of things before merging:

1. `supportsContext` is overridden with a body identical to the inherited default (`context instanceof FeignContext`). The PR description says it's there so japicmp can validate the new convention, but that doesn't apply — this is a brand-new class with no prior release baseline, and the sibling `DefaultFeignObservationConvention` it extends doesn't override this either. Please drop the override; inheritance already does the right thing.

2. `targetUri()` catches `URISyntaxException | UnsupportedOperationException` around `feignTarget().url()`, but `RequestTemplate.feignTarget()` can return `null` for any RequestTemplate not routed through `RequestTemplateFactoryResolver` (e.g. a hand-built RequestTemplate from a custom Client or test double). That path would NPE uncaught, which defeats the class's stated guarantee that "a missing or invalid url never breaks the observation." Please null-check `feignTarget()` explicitly (or widen the catch) so that guarantee actually holds.

Smaller, non-blocking:
- No test exercises the `URISyntaxException` branch in `targetUri()` — worth a case for parity with the other catch arm.
- README typo: "urless" → "url-less".

Happy to merge once 1 and 2 are addressed.

PR #3548, opened 2026-09-01
Allow retryable exceptions without a request

### LINKED ISSUE #1296: RetryableException requires Request
The constructors for RetryableException were updated in #1039 to require the original Request object. In some cases the Request may contain sensitive information and never be used from the RetryableException, so it would be nice to have an option to create one without the Request to limit surface area. 

Additionally, FeignException already exposes a `hasRequest()` method, so clients would be able to check whether RetryableException has a Request before attempting to access it.

I'd be more than happy to open a PR if this sounds like a reasonable use case. Thanks!

taljeon:
## Summary

- add `RetryableException` overloads that do not require or retain a `Request`
- support both cause and no-cause construction using the current millisecond `Long retryAfter` value
- verify request absence and preserved retry metadata with focused tests

## Testing

- `mvn -pl core -Pdev -Dtoolchain.skip=true -Dtest=RetryableExceptionTest test` (5 passed)
- `mvn -pl core -Pdev -Dtoolchain.skip=true verify` (731 passed, 4 skipped)
- `mvn clean install -pl core -Pdev -Dtoolchain.skip=true` (731 passed, 4 skipped)
- current-head CircleCI setup and full PR build after merging `master` (JDK 25)
- `git diff --check`
- publication leak scan (no new findings in changed lines or this PR body)

## AI assistance

OpenAI Codex assisted with code navigation, implementation, and validation. I reviewed and verified the final changes.

Fixes #1296

### DIFF
--- a/core/src/main/java/feign/RetryableException.java
+++ b/core/src/main/java/feign/RetryableException.java
@@ -71,6 +71,44 @@ public class RetryableException extends FeignException {
   }
 
   /**
+   * Represents a retryable exception without retaining the original request.
+   *
+   * <p>Use this constructor when the request is unavailable or should not be retained by the
+   * exception.
+   *
+   * @param status the HTTP status code
+   * @param message the exception message
+   * @param httpMethod the HTTP method (GET, POST, etc.)
+   * @param retryAfter the retry delay in milliseconds
+   */
+  public RetryableException(int status, String message, HttpMethod httpMethod, Long retryAfter) {
+    super(status, message);
+    this.httpMethod = httpMethod;
+    this.retryAfter = retryAfter;
+    this.methodKey = null;
+  }
+
+  /**
+   * Represents a retryable exception without retaining the original request.
+   *
+   * <p>Use this constructor when the request is unavailable or should not be retained by the
+   * exception.
+   *
+   * @param status the HTTP status code
+   * @param message the exception message
+   * @param httpMethod the HTTP method (GET, POST, etc.)
+   * @param cause the underlying cause of the exception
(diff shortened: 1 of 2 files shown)


velo:
Code review: this is a small, correctly-scoped addition — two new constructors on RetryableException matching what #1296 asked for (building a retryable exception without a Request), reusing the existing hasRequest()/request() accessors with no new nullability surface. No structural issues. The overload growth (12 constructors now, disambiguated by parameter type) is a pre-existing pattern in this file, not something this PR should be asked to fix on its own — worth a follow-up migrating the class to named static factories at some point, not blocking here.

Only holding off on approving because ci/circleci: setup-environment is currently failing on this PR and I don't have visibility into why from here (the API isn't returning step-level logs). Once CI is green this looks mergeable to me — could you take a look at the failure or push an empty commit to retrigger if it looks like a flake?

taljeon:
Thanks for flagging this. I traced the old setup failure to the repository's mvnd 1.0.2 download URL. Current `master` includes #3557, which upgrades the wrapper to 1.0.6 specifically for that broken URL, so I merged current `master` into this branch and retriggered CI.

The setup job and full PR build are now green on `16dd36c3`, and the PR diff remains limited to the two `RetryableException` source/test files. Would you mind taking another look?

velo (approved):
Code looks good, and the CI failure that was blocking this is resolved (mvnd wrapper bump came in from master). Additive-only constructors, no behavior change to existing paths. Thanks for tracking down the CI cause, @taljeon.

PR #3553, opened 2026-09-02
Refactor Logger level checks and encapsulate Response.Builder fields

saikat709:
Refactors logging level comparisons and encapsulates `Response.Builder` fields for better API hygiene.

### Summary of Changes

* **`Logger.Level#atLeast`**: Adds `atLeast(Level other)` helper to `Logger.Level` and replaces `ordinal()` comparisons across `Logger.java`.
* **Logger helper extraction**: Splits `logAndRebufferResponse` into `logResponseHeaders` and `rebufferBody` private helpers.
* **`Response.Builder` encapsulation**: Makes `Builder` fields private (`status`, `reason`, `headers`, `body`, `request`).

### Tests Added

* `LoggerMethodsTest`: Adds tests for `Logger.Level.atLeast()` covering equal, more verbose, and less verbose level comparisons.

$ ./mvnw test -pl core -Dtest=LoggerMethodsTest,LoggerTest,LoggerRebufferTest -Dtoolchain.skip=true
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/core/src/main/java/feign/Logger.java
+++ b/core/src/main/java/feign/Logger.java
@@ -68,7 +68,7 @@ public abstract class Logger {
   protected void logRequest(String configKey, Level logLevel, Request request) {
     String protocolVersion = resolveProtocolVersion(request.protocolVersion());
     log(configKey, "---> %s %s %s", request.httpMethod().name(), request.url(), protocolVersion);
-    if (logLevel.ordinal() >= Level.HEADERS.ordinal()) {
+    if (logLevel.atLeast(Level.HEADERS)) {
 
       for (String field : request.headers().keySet()) {
         if (shouldLogRequestHeader(field)) {
@@ -81,7 +81,7 @@ public abstract class Logger {
       int bodyLength = 0;
       if (request.body() != null) {
         bodyLength = request.length();
-        if (logLevel.ordinal() >= Level.FULL.ordinal()) {
+        if (logLevel.atLeast(Level.FULL)) {
           String bodyText =
               request.charset() != null ? new String(request.body(), request.charset()) : null;
           log(configKey, ""); // CRLF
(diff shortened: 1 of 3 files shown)


velo (changes requested):
The Level.atLeast() extraction is a good, well-named improvement over the scattered ordinal() comparisons, and shortening logAndRebufferResponse by pulling out logResponseHeaders/rebufferBody is a reasonable move. A few things need fixing before merge:

1. The six new Response.Builder getters aren't needed. Builder is a static nested class of Response, so the enclosing class already has access to its private fields directly — I verified this compiles fine (mvn -pl core compile) with the getters removed and the constructor reading builder.status/builder.request/etc. directly. The getters add a second, wider access surface (package-visible, no benefit) and collide in name with the existing fluent setters (status() getter vs status(int) setter). Please drop them and keep direct field access from the constructor — the private modifier change alone already achieves the stated goal.

2. logResponseHeaders(String, Level, Response) takes a logLevel parameter that's never used — the caller already gates the call behind logLevel.atLeast(Level.HEADERS). Please drop the unused parameter.

3. One spot wasn't converted to the new abstraction: `logLevel.compareTo(Level.NONE) > 0` should be `logLevel.atLeast(Level.BASIC)`, consistent with the rest of the refactor.

4. LoggerMethodsTest.java currently fails the project's own formatting check (git-code-format-maven-plugin / verify-formatting) — there's a stray trailing-whitespace-only blank line before one of the new @Test methods. This will fail CI as-is; please run the formatter.

Happy to merge once these are addressed.

saikat709:
Thanks for the review on this PR. I have addressed all your suggestions about the edits. And pushed the updates. Additionally, I have synced the branch with the latest main branch to keep this branch up to date.

velo (approved):
All four points addressed: Builder getters dropped with fields made private, unused `logLevel` param gone, the `compareTo(Level.NONE)` spot converted to `atLeast(Level.BASIC)`, and formatting is clean. `Level.atLeast` is purely additive and the Builder fields were never public API. Thanks!

Issue #2944, opened 2025-07-10
Add ability to not throw exception when http status code > 399


ihsansfd:
Currently our team convention is to use response status instead of throwing exception. Say our model is like this:

```
public class BaseResponse {
   private String message;
   private int httpStatusCode;
   private Boolean isFailed;
   private int errorCode;
}
```

So when our API endpoint failed, we return BaseResponse with failed status instead. Now this is not possible to be read by the client using openfeign since feign will throw exception directly if >399 response status retrieved, so our returned response is useless. We can make our api to return 200 status code instead, but that would violate RESTFUL.

Understand that openfeign is designed to be strictly RESTFUL, but I don't think forcing your user to throw exception is the right thing, because different teams have different convention used on how they handle the failure, and in our case we use our response entity that would describe the failure status.

I appreciate if this feature can be considered, and I will be happy to contribute if needed. Thank you.

matevskial:
I wonder how is this feature not implemented in feign in 21-st century year 2025?

Why is there no easy way of configuring openfeign to NOT THROW  EXCEPTION if status code > 399?

I get it, there is a way of writing a custom error decoder, and there is a way of manually parsing the response using my own choice of object mappers or other mappers like gson, but that is not what we want.

Why is openfeign so bad? Nevermind, don't answer this question, my previous two questions are better for coming up with productive answer.

velo:
Thanks for the report. This project is community-maintained, and the maintainer will not implement fixes or features directly. If you want this to move forward, please open a pull request with automated tests that cover the change.

EduardoRadieske:
Interesting point. I ran into a very similar situation recently.

In our case, the API correctly returned an HTTP error status (e.g. 400/404) along with a structured JSON body containing business error details. However, since Feign throws a `FeignException` before the response reaches the `Decoder`, the only way I found to access that payload was by catching the exception and reading the body manually:

```java
try {
    BaseResponse response = client.call();
} catch (FeignException e) {
    String body = e.contentUTF8();
    // deserialize body into BaseResponse
}
```

This works, but it feels like a workaround rather than the intended approach, especially when the API intentionally returns a well-defined error model.

Did your team ever find a cleaner solution for this? Or has OpenFeign introduced any mechanism since this issue was opened that allows decoding error responses without relying on exceptions?

PR #3559, opened 2026-09-05
Add ability to make error as a value.

ihsansfd:
Solving issue #2944.

Unfortunately for status code 204 and 404, all decoders are implemented to return empty body. I don't know what's the best way to handle this, so I leave it as is (make them keep returning empty body to avoid backward incompatibility). Any guideline to handle this better is much appreciated.

This is my first PR in this repo, do tell me if there is anything need to be fixed. Thanks 🙏

### DIFF
--- a/core/src/main/java/feign/AsyncFeign.java
+++ b/core/src/main/java/feign/AsyncFeign.java
@@ -245,6 +245,7 @@ public final class AsyncFeign<C> {
                       dismiss404,
                       closeAfterDecode,
                       decodeVoid,
+                      decodeErrorResponses,
                       responseInterceptorChain()),
                   AsyncResponseHandler.class,
                   capabilities);
--- a/core/src/main/java/feign/AsyncResponseHandler.java
+++ b/core/src/main/java/feign/AsyncResponseHandler.java
@@ -38,6 +38,28 @@ class AsyncResponseHandler {
       boolean closeAfterDecode,
       boolean decodeVoid,
       ResponseInterceptor.Chain executionChain) {
+    this(
+        logLevel,
+        logger,
+        decoder,
+        errorDecoder,
+        dismiss404,
+        closeAfterDecode,
+        decodeVoid,
+        false,
+        executionChain);
+  }
+
+  AsyncResponseHandler(
+      Level logLevel,
+      Logger logger,
+      Decoder decoder,
+      ErrorDecoder errorDecoder,
+      boolean dismiss404,
+      boolean closeAfterDecode,
+      boolean decodeVoid,
+      boolean decodeErrorResponses,
+      ResponseInterceptor.Chain executionChain) {
     this.responseHandler =
         new ResponseHandler(
             logLevel,
@@ -47,6 +69,7 @@ class AsyncResponseHandler {
             dismiss404,
             closeAfterDecode,
             decodeVoid,
+            decodeErrorResponses,
             executionChain);
   }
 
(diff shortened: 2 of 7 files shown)


velo:
Thanks for this — the design is sound and you clearly thought about `Retryer`, `TypedResponse` and decoder applicability. It's correctly targeted at master: everything is additive and binary-compatible, nothing here needs 14.x. A round of fixes and I'll take it.

## Blocking

**1. `void` methods silently swallow errors.**

The new branch at `InvocationContext.java:105-118` runs *before* the void guard and never re-checks it:

```java
if (!shouldDecodeResponseBody) {
    ...
    return decodeErrorResponseBody(response, error);   // decoder called with void.class
}

if (isVoidType(returnType) && !decodeVoid) {           // :120 — success path only
```

With `Feign.builder().decoder(new JacksonDecoder()).decodeErrorResponses()`, a `void` method that receives a 500 with a JSON body returns normally — no exception at all. `JacksonDecoder.canDecode` is just `Util.isJsonContentType(response)`, and `decode(..., void.class)` returns `null` without throwing. So every failed `void` POST/DELETE reads as a success. It also breaks the `decodeVoid=false` contract, which is meant to guarantee the decoder never sees a void return type.

Fold `isVoidType(returnType) && !decodeVoid` into `shouldDecodeErrorResponseBody`, and add a test for it.

**2. The flag implicitly stops throwing on 404, 204 and bodiless errors.**

With the stock builder (no `.decoder(...)`):

| response | today | with the flag |
| --- | --- | --- |
| 404 + body, `String` return | throws `NotFound` | returns `null` |
| 401 + body, `String` return | throws `Unauthorized` | returns the body |
| bodiless 401, any return type | throws `Unauthorized` | returns `null` |

`StringDecoder.canDecode` returns true for status 404, status 204 **or `response.body() == null`**, so the gate at `:149-150` passes and the exception disappears. The bodiless case is the one that worries me — an auth failure reading as an empty result is a bug report waiting to happen. Either exclude the 404/204/null-body cases from the new path, or document and test the change explicitly. The javadoc at `BaseBuilder.java:283-287` notes the 404 envelope "arrives empty" but not that the exception is gone.

**3. No README.** New public functionality needs documenting — `README.md:1244` (`### Error Handling`) is the place.

**4. Test hygiene.** `DecodeErrorResponsesTest.java:37` never closes its `MockWebServer` — there's no `@AfterEach`, so it leaks a server, its thread pool and a port per test class. See `FeignTest.java:1591` for the pattern. And `:191` uses a fully-qualified `new feign.optionals.OptionalDecoder(...)`; please use an import.

## Strongly recommended

- **Mark `decodeErrorResponses()` `@Experimental`.** Every other recent `BaseBuilder` addition is, and `PredicatedDecoder` — which this feature's semantics hang on — is too. It keeps the behavior out of japicmp's contract while 1 and 2 get settled.
- **Cap `bufferBody`** (`InvocationContext.java:178-187`). It has no size limit, unlike `disconnectResponseBodyIfNeeded` eleven lines below, which stops at `MAX_RESPONSE_BUFFER_SIZE`. With the flag on you buffer the entire body on every 4xx/5xx before the `ErrorDecoder` even runs, and the decoded value then holds a second copy.
- **The `AsyncFeign` path has no tests** — `AsyncFeign.java:248` and the new `AsyncResponseHandler` constructor are untested.

## Other coverage gaps

Beyond the void case: the `!(decoder instanceof PredicatedDecoder)` branch at `:149` is never exercised (every decoder in the test file is a `PredicatedDecoder`), the `ErrorDecoder`-returns-null branch at `:169-171` is dead, and nothing tests the default decoder — which is exactly where point 2 bites.

One smaller thing: `decodeErrorResponseBody` catches only `FeignException` (`:168`). That holds today because `decode()` normalizes everything, but if anything else escapes, the original `error` is lost with no `addSuppressed`.

